PR #1, opened 2020-01-07
Updating pt-BR translation

rating89us:


mezzazon:
Thank you. I'm sorry that delayed review of the PR. We'll include it into next release.

PR #33, opened 2020-02-24
Cwa 69 openalias

OleksandrSobol:


lib/src/screens/send/send_page.dart
```diff
@@ -48,10 +49,42 @@ class SendFormState extends State<SendForm> {
...
+
   bool _effectsInstalled = false;
 
   final _formKey = GlobalKey<FormState>();
 
+  @override
+  void initState() {
+    _focusNode.addListener(() async {
+      if (!_focusNode.hasFocus &&_addressController.text.isNotEmpty) {
+        await fetchXmrAddressAndRecipientName(formatDomainName(_addressController.text)).then((map) async {
```
mezzazon: Use `await` or promises (`.then()`) don't mix it.

lib/src/domain/common/openalias.dart
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  return formattedName;
+}
+
+Future<Map<String, String>> fetchXmrAddressAndRecipientName(String name) async {
+  final map = {"recipient_address" : name, "recipient_name" : name};
+
+  try {
+    await DnsUtils.lookupRecord(name, RRecordType.TXT, dnssec: true).then((txtRecord) {
```
mezzazon: Use await or promises (.then()) don't mix it.

lib/src/domain/common/openalias.dart
```diff
@@ -0,0 +1,50 @@
...
+      if (txtRecord != null) {
+        String record;
+
+        for (int i = 0; i < txtRecord.length; i++) {
```
mezzazon: Use forEach (`for(final element in collection)`) or `.fold(initValue, hanlderFunc)` instead of `for` loop with with manual indecies counting.

lib/src/domain/common/openalias.dart
```diff
@@ -0,0 +1,50 @@
...
+  String formattedName = name;
+
+  if (name.contains(".")) {
+    formattedName = name.replaceAll("@", ".");
+  }
+
+  return formattedName;
+}
+
+Future<Map<String, String>> fetchXmrAddressAndRecipientName(String name) async {
```
mezzazon: Better to create separated `class` for this info (like `OpenaliasRecord` or smth like that) with properties `name` and `address`. And rename the function for more short one (but keep it informative).

lib/src/screens/send/send_page.dart
mezzazon: Why not move this logic into send store ? And there check is entered text is domain or xmr address and then if it is domain fetch needed info there ?

lib/src/domain/common/openalias.dart
mezzazon: For example `final record = await DnsUtils.lookupRecord(...);` and then use record "synchronized".

lib/src/screens/send/send_page.dart
```diff
@@ -57,34 +56,38 @@ class SendFormState extends State<SendForm> {
...
-          }
-        });
+        getOpenaliasRecord(context);
       }
     });
 
     super.initState();
   }
 
+  void getOpenaliasRecord(BuildContext context) async {
```
mezzazon: Return `Future<void>` instead of just `void` if is used `async`

lib/src/domain/common/openalias_record.dart
```diff
@@ -0,0 +1,61 @@
...
+          }
+        }
+      }
+    } catch (e) {
+      print("${e.toString()}");
+    }
+  }
+
+}
+
```
mezzazon: Rework this class.

lib/src/domain/common/openalias_record.dart
```diff
@@ -0,0 +1,61 @@
...
+  OpenaliasRecord({this.name});
+
+  String name;
+  String address;
+
+  String get recordName => name;
+  String get recordAddress => address;
+
+  String formatDomainName() {
```
mezzazon: Make this method static.

lib/src/domain/common/openalias_record.dart
```diff
@@ -0,0 +1,61 @@
...
+    String formattedName = name;
+
+    if (name.contains("@")) {
+      formattedName = name.replaceAll("@", ".");
+    }
+
+    return formattedName;
+  }
+
+  Future<void> fetchAddressAndName(String name) async {
```
mezzazon: Make this method static and return `OpenaliasRecord` if will find a record.

lib/src/domain/common/openalias_record.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:basic_utils/basic_utils.dart';
+
+class OpenaliasRecord {
+
+  OpenaliasRecord({this.name});
+
+  String name;
+  String address;
```
mezzazon: Make this properties `final`.

lib/src/domain/common/openalias_record.dart
```diff
@@ -0,0 +1,61 @@
...
+
+  String get recordName => name;
+  String get recordAddress => address;
```
mezzazon: Remove it.

lib/src/domain/common/openalias_record.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:basic_utils/basic_utils.dart';
+
+class OpenaliasRecord {
```
mezzazon: Let make this class as dump data holder which will just store name and monero address. So for get it user will be needed to call static method which will return filled instance of `OpenaliasRecord` if will find the record.

lib/src/stores/send/send_store.dart
```diff
@@ -154,6 +158,16 @@ abstract class SendStoreBase with Store {
     }
   }
 
+  Future<bool> isOpenaliasRecord(String name) async {
```
mezzazon: We can check for not is the text is openalias record, firstly we can try to check is it domain url or not and then if it's domain url we can think that it's openalias. It can be more faster if you are checking is it openalias on change of each character in address text field.

PR #75, opened 2021-02-03
[cake-254] Added ability to track trade directly with a web browser for ChangeNow transactions

tannerdsilva:
Created new trade list item TrackTradeListItem to display and handle this information

### DIFF
--- a/lib/view_model/trade_details_view_model.dart
+++ b/lib/view_model/trade_details_view_model.dart
@@ -96,7 +96,7 @@ abstract class TradeDetailsViewModelBase with Store {
           title: "Track",
           value: "https://changenow.io/exchange/txs/${trade.id.toString()}",
           onTap: () {
-            launch("https://www.blockchain.com/btc/tx/${tx.id}");
+            launch("https://changenow.io/exchange/txs/${trade.id.toString()}");
           }));
     }
 
--- a/lib/view_model/trade_details_view_model.dart
+++ b/lib/view_model/trade_details_view_model.dart
@@ -91,12 +91,14 @@ abstract class TradeDetailsViewModelBase with Store {
           value: trade.provider.toString()));
     }
 
-    if (trade.provider.title == "ChangeNOW") {
+    if (trade.provider is ChangeNowExchangeProvider) {
+      final buildURL =
+          'https://changenow.io/exchange/txs/${trade.id.toString()}';
       items.add(TrackTradeListItem(
-          title: "Track",
-          value: "https://changenow.io/exchange/txs/${trade.id.toString()}",
+          title: 'Track',
+          value: buildURL,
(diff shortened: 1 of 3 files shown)


lib/view_model/trade_details_view_model.dart
```diff
@@ -90,6 +91,15 @@ abstract class TradeDetailsViewModelBase with Store {
           value: trade.provider.toString()));
     }
 
+    if (trade.provider.title == "ChangeNOW") {
+      items.add(TrackTradeListItem(
+          title: "Track",
+          value: "https://changenow.io/exchange/txs/${trade.id.toString()}",
+          onTap: () {
+            launch("https://changenow.io/exchange/txs/${trade.id.toString()}");
```
mezzazon, line 99: Use single quotes instead of double. Don't repeat the link twice, use `final url = https://changenow.io/exchange/txs/${trade.id.toString()};` for not repeat yourself.

lib/view_model/trade_details_view_model.dart
```diff
@@ -90,6 +91,15 @@ abstract class TradeDetailsViewModelBase with Store {
           value: trade.provider.toString()));
     }
 
+    if (trade.provider.title == "ChangeNOW") {
```
mezzazon, line 94: Unnamed constant. If `provider` in current context is `ExchangeProviderDescription` then compare with a static constant which you need instead of compare title strings.

mezzazon (changes requested):

mezzazon:
cc @tannerdsilva

lib/view_model/trade_details_view_model.dart
tannerdsilva: Like this? `if (trade.provider is ChangeNowExchangeProvider) {`

lib/view_model/trade_details_view_model.dart
```diff
@@ -90,6 +91,17 @@ abstract class TradeDetailsViewModelBase with Store {
           value: trade.provider.toString()));
     }
 
+    if (trade.provider is ChangeNowExchangeProvider) {
```
mezzazon, line 94: It will never works, this statement will be always false. `provider` is not sub class (or implements) `ExchangeProvider` or `ChangeNowExchangeProvider`.

lib/view_model/trade_details_view_model.dart
mezzazon: Nope. `provider` in current context is `ExchangeProviderDescription` not `ExchangeProvider` or `ChangeNowExchangeProvider`.

tannerdsilva: Implemented static comparison between `trade.provider` and `ExchangeProviderDescription.changeNow`

PR #76, opened 2021-02-03
[CAKE-128] Subaddresses are automatically given a name when created

tannerdsilva:


mezzazon:
merge conflict to new main branch, will reimplement a bit later

PR #95, opened 2021-03-19
CAKE-272 | implemented localization generation script to the app (gen…

alex-suomi:
…erate_localization.dart)


tool/generate_localization.dart
```diff
@@ -0,0 +1,199 @@
...
+    output += 'class \$$locale extends S {' + '\n';
+    output += '  const \$$locale();' + '\n';
+
+    if (locale != locales.first) {
+      output += textDirectionDeclaration;
+
+      inputContent = File(inputPath + 'strings_$locale.arb').readAsStringSync();
+      config = json.decode(inputContent) as Map<String, dynamic>;
+
+      output += localizedStrings(config: config, hasOverride: true);
```
mezzazon, line 142: Why here we have special case if we can try to generate `hasOverride` inside of the loop body ?

mezzazon (changes requested):

mezzazon:
@alex-suomi cc

mezzazon (approved):

PR #150, opened 2021-06-12
Translate update

Zy143L:
update


mezzazon:
Thanks for your attention

Issue #121, opened 2021-04-30
Typo - German translation from seed to "Samen" is not very accurate.


KKA11010:
Running on android mobile v4.1.6
Quite funny.
Unfortunately, i have zero experience with any translation file logic or with the languages that are required to contribute.

If the Cake Wallet website is also open source, i would love to contribute there, since i write JS/TS/CSS and have many hours hobby experience in web developement and the tools there.

BTW, the website logo is not updated yet... ;-P

I wish you all a great time!

    Thank you for your contributions to the open source world.

hope-with-me:
Sorry for the delayed reply. You can send a pull request with fixes here 
https://github.com/cake-tech/cake_wallet/blob/main/res/values/strings_de.arb

PR #151, opened 2021-06-13
Improve German translation

dvaupel:
Correct errors and make it more natural sounding.

Vocabulary is also more adequate, as requested by #121.


mezzazon:
Thanks for your attention

PR #152, opened 2021-06-16
Cake 192 implement unstoppable domain

alex-suomi:


android/app/src/main/java/com/cakewallet/cake_wallet/MainActivity.java
```diff
@@ -10,14 +10,20 @@
...
+import com.unstoppabledomains.resolution.Resolution;
+
 import java.security.SecureRandom;
 
 public class MainActivity extends FlutterFragmentActivity {
     final String UTILS_CHANNEL = "com.cake_wallet/native_utils";
-    
+    final String UNSTOPPABLE_DOMAIN_CHANNEL = "com.cakewallet.cake_wallet/unstoppable-domain";
+    final int UNSTOPPABLE_DOMAIN_MIN_VERSION_SDK = 24;
+
```
mezzazon, line 26: Better to keep single one channel for our native calls (same for iOS)

ios/Runner/AppDelegate.swift
```diff
@@ -1,8 +1,13 @@
 import UIKit
 import Flutter
+import UnstoppableDomainsResolution
 
 @UIApplicationMain
 @objc class AppDelegate: FlutterAppDelegate {
+    lazy var resolution : Resolution? =  {
+               return  try? Resolution()
+            }()
```
mezzazon, line 9: Remove unneeded spaces

ios/Runner/AppDelegate.swift
```diff
@@ -75,6 +80,40 @@ import Flutter
             }
         })
         
+        let unstoppableDomainChannel = FlutterMethodChannel(name: "com.cakewallet.cake_wallet/unstoppable-domain", binaryMessenger: controller.binaryMessenger)
+        unstoppableDomainChannel.setMethodCallHandler({ [weak self]
```
mezzazon, line 84: Better to keep single one channel for our native calls (same for android)

ios/Runner/AppDelegate.swift
```diff
@@ -75,6 +80,40 @@ import Flutter
...
+                                      let domain = args["domain"],
+                                      let ticker = args["ticker"] else {
+                                    result(nil)
+                                    return
+                                }
+                                
+                                guard let resolution = self?.resolution else {
+                                    result(nil)
+                                    return
+                                }
```
mezzazon, line 98: Combine all cases for `guard` condition in one `guard` instruction (move let resolution = self?.resolution into `guard` condition upper)

ios/Runner/AppDelegate.swift
```diff
@@ -75,6 +80,40 @@ import Flutter
...
+                                        
+                                resolution.addr(domain: domain, ticker: ticker) { addrResult in
+                                  var address : String = ""
+                                    
+                                  switch addrResult {
+                                      case .success(let returnValue):
+                                        address = returnValue
+                                      case .failure(let error):
+                                        print("Expected Address, but got \(error)")
```
mezzazon, line 107: Here you have different error catch with android side (` handler.post(() -> result.error("INVALID DOMAIN", e.getMessage(), null));` in android code). So need to send error in same way as on android side.

lib/src/screens/send/parse_address_from_domain.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'package:cake_wallet/generated/i18n.dart';
+
+const topLevelDomain = 'crypto';
+
+Future<String> parseAddressFromDomain(
+    BuildContext context, String domain, String ticker) async {
+  try {
+    final name = domain.split('.').last;
+
+    if (name.contains(topLevelDomain)) {
```
mezzazon, line 15: Handle case if `domain` variable will be like ".crypto" or just "crypto". So better to make something like:
```
 final domainParts = domain.split('.');
 final name = domainParts.last;
 
 if (domainParts.length > 1 && name.contains(topLevelDomain)) {
 ...
```

lib/src/screens/send/parse_address_from_domain.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'package:cake_wallet/entities/unstoppable_domain_address.dart';
+import 'package:cake_wallet/src/widgets/alert_with_one_action.dart';
+import 'package:cake_wallet/utils/show_pop_up.dart';
+import 'package:flutter/material.dart';
+import 'package:cake_wallet/generated/i18n.dart';
+
+const topLevelDomain = 'crypto';
+
+Future<String> parseAddressFromDomain(
+    BuildContext context, String domain, String ticker) async {
```
mezzazon, line 11: This function is overloaded it's parsing domain and presenting popup, so need to separate those functionality into each own function. Move "logic" part from "screens" directory.

lib/src/screens/send/send_template_page.dart
```diff
@@ -145,7 +145,7 @@ class SendTemplatePage extends BasePage {
                                     .primaryTextTheme
                                     .headline
                                     .decorationColor),
-                            validator: sendViewModel.addressValidator,
+                            //validator: sendViewModel.addressValidator,
```
mezzazon, line 148: Better to make new validator which will use address validator and domain validator (create if not exists) and keep validation for address fields.

lib/src/screens/exchange/exchange_template_page.dart
```diff
@@ -150,9 +150,8 @@ class ExchangeTemplatePage extends BasePage {
                               .color,
                           currencyValueValidator: AmountValidator(
                               type: exchangeViewModel.wallet.type),
-                          addressTextFieldValidator: AddressValidator(
-                              type:
-                              exchangeViewModel.depositCurrency),
+                          //addressTextFieldValidator: AddressValidator(
```
mezzazon, line 153: Better to make new validator which will use address validator and domain validator (create if not exists) and keep validation for address fields.

lib/src/screens/exchange/exchange_template_page.dart
```diff
@@ -190,8 +189,8 @@ class ExchangeTemplatePage extends BasePage {
                                 .decorationColor,
                             currencyValueValidator: AmountValidator(
                                 type: exchangeViewModel.wallet.type),
-                            addressTextFieldValidator: AddressValidator(
-                                type: exchangeViewModel.receiveCurrency),
+                            //addressTextFieldValidator: AddressValidator(
```
mezzazon, line 192: Better to make new validator which will use address validator and domain validator (create if not exists) and keep validation for address fields.

lib/src/screens/send/send_page.dart
```diff
@@ -173,6 +173,9 @@ class SendPage extends BasePage {
                                         .headline
                                         .decorationColor),
                                 validator: sendViewModel.addressValidator,
```
mezzazon, line 175: Better to make new validator which will use address validator and domain validator (create if not exists) and keep validation for address fields.

lib/src/screens/exchange/exchange_page.dart
```diff
@@ -223,6 +223,15 @@ class ExchangePage extends BasePage {
                                     type: exchangeViewModel.wallet.type),
                                 addressTextFieldValidator: AddressValidator(
```
mezzazon, line 224: Better to make new validator which will use address validator and domain validator (create if not exists) and keep validation for address fields.

mezzazon (changes requested):

mezzazon:
cc @alex-suomi

PR #157, opened 2021-07-01
CAKE-343 | applied subaddresses for contact book; added addresses fie…

alex-suomi:
…ld to wallet_info.dart; added updateAddressesInfo() to wallet_base.dart; applied updateAddressesInfo() to wallet services and view models


lib/monero/monero_wallet.dart
```diff
@@ -445,4 +445,26 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
...
+  Future<void> updateAddressesInfo() async {
+    final Map<String, String> _addresses = {};
+    final _subaddressList = MoneroSubaddressList();
+
+    accountList.accounts.forEach((account) {
+      _subaddressList.update(accountIndex: account.id);
+      _subaddressList.subaddresses.forEach((subaddress) {
+        _addresses.addAll({subaddress.address:subaddress.label});
+      });
+    });
```
mezzazon, line 459: Why need to create new `subaddressList` ? You can use subaddressList property in MoneroWallet class.

lib/view_model/contact_list/contact_list_view_model.dart
```diff
@@ -16,11 +16,23 @@ class ContactListViewModel = ContactListViewModelBase
...
+            walletContacts.add(WalletContact(
+                address,
+                info.name + ' ($label)',
+                walletTypeToCryptoCurrency(info.type)));
+          });
+        } else {
+          walletContacts.add(WalletContact(info.address, info.name,
+              walletTypeToCryptoCurrency(info.type)));
+        }
+      }
```
mezzazon, line 33: Avoid nested conditions. Use "optional" operator for call forEach method on info.addresses (`info.addresses?.forEach(...)`).

lib/monero/monero_wallet.dart
```diff
@@ -445,4 +445,26 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
...
+
+  @override
+  Future<void> updateAddressesInfo() async {
+    final Map<String, String> _addresses = {};
+    final _subaddressList = MoneroSubaddressList();
+
+    accountList.accounts.forEach((account) {
+      _subaddressList.update(accountIndex: account.id);
+      _subaddressList.subaddresses.forEach((subaddress) {
+        _addresses.addAll({subaddress.address:subaddress.label});
```
mezzazon, line 457: Instead of create new `Map` on each iteration better set value via literal `_addresses[subaddress.address] = subaddress.label`.

lib/view_model/contact_list/contact_list_view_model.dart
mezzazon: Need to rework that conditional, because that condition does not catch case when `address` is empty or null  but `addresses` are not empty.

mezzazon: Need to set `info.addresses` to empty map as default value, then check does it not empty. Maybe will be better to remove `info.address` property at all and use `info.addresses.keys.first` or `info.addresses.values.first` (or last element if you wanna to show "last used") as address by default.

mezzazon (changes requested):

mezzazon:
cc @alex-suomi

mezzazon:
Work with wallet addresses should be reworked more deeper. Need to make WalletAddresses source of getting addresses in all points where addresses are getting. Need to remove addresses from Wallet classes and move logic for creation, fetching and editing into WalletAddresses classes.

@alex-suomi cc

PR #161, opened 2021-07-22
Cake 345 batch sending

alex-suomi:


cw_monero/lib/transaction_history.dart
```diff
@@ -102,6 +106,59 @@ PendingTransactionDescription createTransactionSync(
...
+      int accountIndex = 0}) {
+  final List<Pointer<Utf8>> addressesPointers = addresses.map(Utf8.toUtf8).toList();
+  final Pointer<Pointer<Utf8>> addressesPointerPointer = allocate(count: size);
+
+  final List<Pointer<Utf8>> amountsPointers = amounts.map(Utf8.toUtf8).toList();
+  final Pointer<Pointer<Utf8>> amountsPointerPointer = allocate(count: size);
+
+  for (int i = 0; i < size; i++) {
+    addressesPointerPointer[ i ] = addressesPointers[ i ];
+    amountsPointerPointer[ i ] = amountsPointers[ i ];
```
mezzazon, line 124: remove unnecessary spaces around indecies for lists.

lib/bitcoin/electrum_wallet.dart
```diff
@@ -174,28 +172,75 @@ abstract class ElectrumWalletBase extends WalletBase<ElectrumBalance,
...
+      stringDoubleToBitcoinAmount(item.cryptoAmount.replaceAll(',', '.')) <= 0)
+          .toList();
+
+      if (nullAmountItems?.isNotEmpty ?? false) {
+        throw BitcoinTransactionWrongBalanceException(currency);
+      }
+
+      credentialsAmount = sendItemList.fold(0, (previousValue, element) =>
+      previousValue + stringDoubleToBitcoinAmount(
+          element.cryptoAmount.replaceAll(',', '.')));
```
mezzazon, line 204: better formatting for that piece of text

lib/bitcoin/electrum_wallet.dart
```diff
@@ -174,28 +172,75 @@ abstract class ElectrumWalletBase extends WalletBase<ElectrumBalance,
...
+      }
+
+      amount = allAmount - credentialsAmount < minAmount
+          ? allAmount
+          : credentialsAmount;
+
+      fee = amount == allAmount
+          ? allAmountFee
+          : calculateEstimatedFee(transactionCredentials.priority, amount,
+          outputsCount: sendItemList.length + 1);
```
mezzazon, line 217: for multi destination case we doesn't support to send all in that implementation, so we need to avoid "allAmount" or "sendAll" for multi destination implementation

lib/bitcoin/electrum_wallet.dart
```diff
@@ -174,28 +172,75 @@ abstract class ElectrumWalletBase extends WalletBase<ElectrumBalance,
...
-        ? allAmount
-        : credentialsAmount;
-    final fee = transactionCredentials.amount == null || amount == allAmount
-        ? allAmountFee
-        : calculateEstimatedFee(transactionCredentials.priority, amount);
+    var credentialsAmount = 0;
+    var amount = 0;
+    var fee = 0;
+
+    if (sendItemList.length > 1) {
```
mezzazon, line 187: better to hold result of that condition into "hasMultiDestination" variable, because then you could to use it into next logic blocks

lib/bitcoin/electrum_wallet.dart
mezzazon: use another name for previousValue argument (maybe acc aka accamulator) and maybe value instead of element

lib/bitcoin/electrum_wallet.dart
mezzazon, line 217: i think better if `allAmount - credentialsAmount < minAmount` then throw exception, because we have no support for that in multi destination case

mezzazon: same for fee variable (it should not have branch for case that amount is equal to sendAll amount)

lib/bitcoin/electrum_wallet.dart
```diff
@@ -244,11 +289,18 @@ abstract class ElectrumWalletBase extends WalletBase<ElectrumBalance,
       }
     });
 
-    txb.addOutput(
-        addressToOutputScript(transactionCredentials.address, networkType),
-        amount);
+    sendItemList.forEach((item) {
+      final _amount = item.sendAll
```
mezzazon, line 293: avoid "_" prefix for non private variables/methods

lib/bitcoin/bitcoin_transaction_credentials.dart
```diff
@@ -1,9 +1,9 @@
...
+import 'package:cake_wallet/view_model/send/send_item.dart';
 
 class BitcoinTransactionCredentials {
-  BitcoinTransactionCredentials(this.address, this.amount, this.priority);
+  BitcoinTransactionCredentials(this.sendItemList, this.priority);
 
-  final String address;
-  final String amount;
+  final List<SendItem> sendItemList;
   BitcoinTransactionPriority priority;
```
mezzazon, line 8: use outputs instead of sendItemList for naming

lib/view_model/send/send_item.dart
```diff
@@ -0,0 +1,197 @@
...
+import 'package:cake_wallet/store/settings_store.dart';
+import 'package:cake_wallet/generated/i18n.dart';
+
+part 'send_item.g.dart';
+
+const String cryptoNumberPattern = '0.0';
+
+class SendItem = SendItemBase with _$SendItem;
+
+abstract class SendItemBase with Store {
```
mezzazon, line 24: "SendItem" is not best name, let use just "Output" for now ("outputs" for list of Output objects)

lib/view_model/send/send_item.dart
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  Key key;
+
+  @observable
+  String fiatAmount;
+
+  @observable
+  String cryptoAmount;
+
```
mezzazon, line 39: if we have access to a wallet in this class, why we cannot to calculate result amount (crypto amount or "sendAll" amount) here ?

cw_monero/lib/transaction_history.dart
```diff
@@ -102,6 +106,59 @@ PendingTransactionDescription createTransactionSync(
       pointerAddress: pendingTransactionRawPointer.address);
 }
 
+PendingTransactionDescription createTransactionMultDestSync(
+    {List<String> addresses,
+      String paymentId,
+      List<String> amounts,
+      int size,
```
mezzazon, line 113: we need single structure for address + amount and remove size argument, because in case if we would to have list of MoneroOutput(maybe ?) class here we doesn't need to rely on unconsistent ""size" argument.

mezzazon (changes requested):

mezzazon:
cc @alex-suomi

lib/bitcoin/electrum_wallet.dart
```diff
@@ -174,28 +173,68 @@ abstract class ElectrumWalletBase extends WalletBase<ElectrumBalance,
...
+      if (sendAllItems?.isNotEmpty ?? false) {
+        throw BitcoinTransactionWrongBalanceException(currency);
+      }
+
+      final nullAmountItems = outputs.where((item) =>
+          item.formattedCryptoAmount <= 0).toList();
+
+      if (nullAmountItems?.isNotEmpty ?? false) {
+        throw BitcoinTransactionWrongBalanceException(currency);
+      }
```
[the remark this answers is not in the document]
mezzazon: Combine into single condition.

lib/monero/monero_wallet.dart
```diff
@@ -149,31 +151,78 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
...
+          item.formattedCryptoAmount <= 0).toList();
+
+      if (nullAmountItems?.isNotEmpty ?? false) {
+        throw MoneroTransactionCreationException('Wrong balance. Not enough XMR on your balance.');
+      }
+
+      var credentialsAmount = 0;
+
+      credentialsAmount = outputs.fold(0, (acc, value) =>
+          acc + value.formattedCryptoAmount);
```
mezzazon, line 182: Why this is not final ?

cw_monero/lib/transaction_history.dart
```diff
@@ -145,3 +216,15 @@ Future<PendingTransactionDescription> createTransaction(
       'priorityRaw': priorityRaw,
       'accountIndex': accountIndex
     });
+
+Future<PendingTransactionDescription> createTransactionMultDest(
+    {List<MoneroOutput> outputs,
+      String paymentId,
```
mezzazon, line 222: PaymentId is kind of deprecated, so better to make keep it optional with empty string by default.

mezzazon (changes requested):

mezzazon:
cc @alex-suomi

PR #217, opened 2022-01-09
Update app_config.sh

plowsof:
No .dart_tool/package_config.json file found, please run "pub get".

This seems to fix this


mkyq:
Thank you!

PR #223, opened 2022-01-11
Cw 3 update UI sending template

godilite:
update sending template
Remove (text) Template, add + icon if n >= 1 templates exist

**Screen shots**
IMG_3940

IMG_3941


mkyq:
@godilite please update you branch to mainstream

mkyq:
not sure that you added translation for all localization files

mkyq:
Good that you fixed your issue with wallet manager, but for next time please send only code related for the task. For fixes like that better to create new branches and send PR from them. Also maybe better to disable auto code formatter, at least for files which you currently do not edit, because it's hard to find out what you exactly changed for the task.

mkyq (changes requested):
See comments to the PR

godilite:
> not sure that you added translation for all localization files

updated

PR #244, opened 2022-01-21
Cw 8 keep screen on while wallet synchronization

godilite:
Add native implementations for IOS and Android to keeping screen awake while synchronizing

### DIFF
--- a/android/app/src/main/java/com/monero/app/MainActivity.java
+++ b/android/app/src/main/java/com/monero/app/MainActivity.java
@@ -42,11 +42,11 @@ public class MainActivity extends FlutterFragmentActivity {
             switch (call.method) {
                 case "enableWakeScreen":
                     getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
-                    handler.post(() -> result.success("screen wake turned ON"));
+                    handler.post(() -> result.success(true));
                     break;
                 case "disableWakeScreen":
                     getWindow().clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
-                    handler.post(() -> result.success("screen wake turned ON"));
+                    handler.post(() -> result.success(true));
                     break;
                 case "sec_random":
                     int count = call.argument("count");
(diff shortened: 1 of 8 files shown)


android/app/src/main/java/com/monero/app/MainActivity.java
```diff
@@ -39,6 +40,14 @@ private void handle(@NonNull MethodCall call, @NonNull MethodChannel.Result resu
 
         try {
             switch (call.method) {
+                case "enableWakeScreen":
+                    getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
+                    handler.post(() -> result.success("screen wake turned ON"));
+                    break;
+                case "disableWakeScreen":
+                    getWindow().clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
+                    handler.post(() -> result.success("screen wake turned ON"));
```
mkyq, line 49: If you need to pass something to result.success better pass boolean value.

ios/CakeWallet/wakeLock.swift
```diff
@@ -0,0 +1,20 @@
...
+func enableWakeScreen() -> String {
+    UIApplication.shared.isIdleTimerDisabled = true
+
+    return "screen wake turned ON"
+}
+
+func disableWakeScreen() -> String{
+    UIApplication.shared.isIdleTimerDisabled = false
+    return "screen wake turned OFF"
+}
```
mkyq, line 20: Same here. If you need to pass some argument into result function, better pass nil(null) in case if you expect void as result for the function, or pass boolean value based on state what changed your function.

lib/reactions/on_wallet_sync_status_change.dart
```diff
@@ -1,21 +1,31 @@
...
 import 'package:cw_core/sync_status.dart';
+import 'package:flutter/services.dart';
 
 ReactionDisposer _onWalletSyncStatusChangeReaction;
 
 void startWalletSyncStatusChangeReaction(
     WalletBase<Balance, TransactionHistoryBase<TransactionInfo>,
             TransactionInfo>
         wallet) {
+  final WakeLock _wakeLock = getIt.get<WakeLock>();
```
mkyq, line 17: Don't define type of `final` variable, because it's necessary as dart will get type of your value, as that's `final` variable, then you will not change the value

mkyq (changes requested):

mkyq (approved):

PR #249, opened 2022-01-25
filter templates by current wallet type

godilite:
filter templates by current wallet type

### DIFF
--- a/lib/src/screens/send/send_page.dart
+++ b/lib/src/screens/send/send_page.dart
@@ -197,7 +197,7 @@ class SendPage extends BasePage {
                             itemCount: itemCount,
                             itemBuilder: (context, index) {
                               final template = templates[index];
-
+                              print(template.cryptoCurrency);
                               return TemplateTile(
                                 key: UniqueKey(),
                                 to: template.name,
--- a/lib/src/screens/send/send_page.dart
+++ b/lib/src/screens/send/send_page.dart
@@ -197,7 +197,6 @@ class SendPage extends BasePage {
                             itemCount: itemCount,
                             itemBuilder: (context, index) {
                               final template = templates[index];
-                              print(template.cryptoCurrency);
                               return TemplateTile(
                                 key: UniqueKey(),
                                 to: template.name,
--- a/lib/view_model/send/send_view_model.dart
+++ b/lib/view_model/send/send_view_model.dart
@@ -130,10 +130,9 @@ abstract class SendViewModelBase with Store {
   bool get isReadyForSend => _wallet.syncStatus is SyncedSyncStatus;
 
   @computed
-  ObservableList<Template> get templates => sendTemplateViewModel.templates
+  List<Template> get templates => sendTemplateViewModel.templates
       .where((template) => template.cryptoCurrency == _wallet.currency.title)
-      .toList()
-      .asObservable();
+      .toList();
 
   @computed
   bool get isElectrumWallet =>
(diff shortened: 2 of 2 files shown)


lib/src/screens/send/send_page.dart
```diff
@@ -197,7 +197,7 @@ class SendPage extends BasePage {
                             itemCount: itemCount,
                             itemBuilder: (context, index) {
                               final template = templates[index];
-
+                              print(template.cryptoCurrency);
```
mkyq, line 200: remove `print`

lib/view_model/send/send_view_model.dart
```diff
@@ -127,15 +130,18 @@ abstract class SendViewModelBase with Store {
   bool get isReadyForSend => _wallet.syncStatus is SyncedSyncStatus;
 
   @computed
-  ObservableList<Template> get templates => sendTemplateViewModel.templates;
+  ObservableList<Template> get templates => sendTemplateViewModel.templates
+      .where((template) => template.cryptoCurrency == _wallet.currency.title)
+      .toList()
+      .asObservable();
```
mkyq, line 136: Not sure that here is needed `ObservableList` as you use `computed` then it will regenerate the result of getter body in any case. Keep `ObservableList` only if you manually edit content of exactly special list. Better to use just `List` here.

lib/view_model/send/send_view_model.dart
```diff
@@ -200,19 +206,16 @@ abstract class SendViewModelBase with Store {
...
-        return bitcoin.createBitcoinTransactionCredentials(
-            outputs, priority);
+        return bitcoin.createBitcoinTransactionCredentials(outputs, priority);
       case WalletType.monero:
         final priority = _settingsStore.priority[_wallet.type];
 
         return monero.createMoneroTransactionCreationCredentials(
-            outputs: outputs,
-            priority: priority);
+            outputs: outputs, priority: priority);
```
mkyq, line 218: I think better to disable auto formatter in your editor, because it's hard to see too much changes in PRs which are not related to the task.

mkyq (changes requested):

lib/view_model/send/send_view_model.dart
godilite: sorry about that, done.

lib/view_model/send/send_view_model.dart
godilite: okay, done

lib/view_model/send/send_view_model.dart
```diff
@@ -127,15 +130,17 @@ abstract class SendViewModelBase with Store {
   bool get isReadyForSend => _wallet.syncStatus is SyncedSyncStatus;
 
   @computed
-  ObservableList<Template> get templates => sendTemplateViewModel.templates;
+  List<Template> get templates => sendTemplateViewModel.templates
+      .where((template) => template.cryptoCurrency == _wallet.currency.title)
```
mkyq, line 134: For compare strings like that better to normalize them, for example be sure that they are in needed register. Please force them to lower case when comparing string like those.

mkyq (changes requested):

mkyq:
cc @godilite

lib/view_model/send/send_view_model.dart
godilite: Alright

mkyq (approved):

PR #252, opened 2022-01-26
soft keyboard dismisses when generate-name icon is pressed

RafiaChy:
@mkyq , I have added the dismissable-keyboard property to generate-name button. The keyboard was not getting dismissed completely because the button resided inside the TextFormField and so onPress method of the button also triggered the focus on TextFormField and thus invoked the onTap method on TextFormField. However, I have tackled it using Stack to separate the onTap and onPress triggering each other.

### DIFF
--- a/lib/src/screens/new_wallet/new_wallet_page.dart
+++ b/lib/src/screens/new_wallet/new_wallet_page.dart
@@ -105,18 +105,50 @@ class _WalletNameFormState extends State<WalletNameForm> {
               padding: EdgeInsets.only(top: 24),
               child: Form(
                 key: _formKey,
-                child: TextFormField(
-                  onChanged: (value) => _walletNewVM.name = value,
-                  controller: _controller,
-                  textAlign: TextAlign.center,
-                  style: TextStyle(
-                      fontSize: 20.0,
-                      fontWeight: FontWeight.w600,
-                      color: Theme.of(context).primaryTextTheme.title.color),
-                  decoration: InputDecoration(
-                    suffixIcon: IconButton(
+                child: Stack(
+                  alignment: Alignment.centerRight,
+                  children: [
+                    TextFormField(
+                      onChanged: (value) => _walletNewVM.name = value,
+                      controller: _controller,
+                      textAlign: TextAlign.center,
+                      style: TextStyle(
+                          fontSize: 20.0,
+                          fontWeight: FontWeight.w600,
+                          color:
+                              Theme.of(context).primaryTextTheme.title.color),
+                      decoration: InputDecoration(
+                        hintStyle: TextStyle(
+                            fontSize: 18.0,
+                            fontWeight: FontWeight.w500,
+                            color: Theme.of(context)
+                                .accentTextTheme
+                                .display3
+                                .color),
+                        hintText: S.of(context).wallet_name,
+                        focusedBorder: UnderlineInputBorder(
+                            borderSide: BorderSide(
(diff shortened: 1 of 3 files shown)


lib/src/screens/restore/wallet_restore_from_keys_form.dart
```diff
@@ -37,6 +37,13 @@ class WalletRestoreFromKeysFromState extends State<WalletRestoreFromKeysFrom> {
   final TextEditingController viewKeyController;
   final TextEditingController spendKeyController;
   final TextEditingController nameTextEditingController;
+  bool buttonPressed;
+
+  @override
+  void initState() {
+    super.initState();
+    buttonPressed = false;
+  }
```
mkyq, line 47: What is doing that ?

lib/src/screens/restore/wallet_restore_from_keys_form.dart
```diff
@@ -54,36 +61,44 @@ class WalletRestoreFromKeysFromState extends State<WalletRestoreFromKeysFrom> {
...
+              children: [
+                BaseTextFormField(
+                  controller: nameTextEditingController,
+                  hintText: S.of(context).wallet_name,
+                  validator: WalletNameValidator(),
+                ),
+                IconButton(
+                  onPressed: () async {
                     final rName = await generateName();
+                    FocusManager.instance.primaryFocus?.unfocus();
```
mkyq, line 75: is necessary that here ?

lib/src/screens/new_wallet/new_wallet_page.dart
```diff
@@ -105,18 +105,50 @@ class _WalletNameFormState extends State<WalletNameForm> {
...
+                                  .decorationColor,
+                              width: 1.0),
+                        ),
+                      ),
+                      validator: WalletNameValidator(),
+                    ),
+                    IconButton(
                       onPressed: () async {
                         final rName = await generateName();
+                        FocusManager.instance.primaryFocus?.unfocus();
```
mkyq, line 150: is necessary that here ?

mkyq (changes requested):

mkyq:
cc @RafiaChy

lib/src/screens/new_wallet/new_wallet_page.dart
RafiaChy: @mkyq , I have tested this again by commenting this statement ` FocusManager.instance.primaryFocus?.unfocus();` and this is essential. Because without this the keyboard won't dismiss if the user decides to switch from typing to pressing the icon button.

lib/src/screens/restore/wallet_restore_from_keys_form.dart
RafiaChy: @mkyq , I have tested this again by commenting this statement ` FocusManager.instance.primaryFocus?.unfocus();` and this is essential. Without this the keyboard is not dismissible.

lib/src/screens/restore/wallet_restore_from_keys_form.dart
RafiaChy: @mkyq , this is my mistake. I was trying to use variable to control onTap. Forgot to delete this one. I am deleting these lines.

mkyq: @RafiaChy ok

mkyq (approved):

PR #256, opened 2022-01-27
show exchange screen template keyboard done button

godilite:
show exchange screen template keyboard done button


mkyq:
Cause of formatting I cannot to review and check all changes

PR #258, opened 2022-01-31
rework templates on exchange screen

godilite:
Templates on the exchange screen follow the old format. upgraded to match the new format used for the Send screen to save space, making it easier for someone to initiate an exchange.


lib/src/screens/exchange/exchange_page.dart
```diff
@@ -501,6 +381,111 @@ class ExchangePage extends BasePage {
...
+                              ),
+                      ),
+                    ),
+                  ),
+                ),
+                ListView.builder(
+                  scrollDirection: Axis.horizontal,
+                  shrinkWrap: true,
+                  physics: NeverScrollableScrollPhysics(),
+                  itemCount: itemCount,
```
mkyq, line 446: if you use `itemCount` only in this place then please replace by `templates.length` and remove `itemCount` variable

mkyq (changes requested):

mkyq (approved):

PR #279, opened 2022-02-21
Redesigned bottom bar of main screen according Figma

RafiaChy:


lib/src/screens/dashboard/dashboard_page.dart
```diff
@@ -132,34 +136,52 @@ class DashboardPage extends BasePage {
...
-                ActionButton(
-                  image: buyImage,
-                  title: S.of(context).buy,
-                  onClick: () async => await _onClickBuyButton(context),
+
+         
+        ClipRect(
+          child:Container(
+          height: 75,
+          width: 343,
```
mkyq, line 144: Can we use flex instead of fixed width, or set margin/padding offset our of the container ?

mkyq: Why we need here fixed width ?

mkyq (changes requested):

mkyq:
cc @RafiaChy

lib/src/screens/dashboard/dashboard_page.dart
RafiaChy: Yeah, you are right Sir. we can rely on margins. I have now used margins only.

mkyq (approved):

PR #281, opened 2022-02-23
fixed subaddress label

Serhii-Borodenko:
displays previously used addresses like first and last 8 characters of subaddress

### DIFF
--- a/lib/src/screens/receive/widgets/address_cell.dart
+++ b/lib/src/screens/receive/widgets/address_cell.dart
@@ -39,7 +39,8 @@ class AddressCell extends StatelessWidget {
   final Function(String) onTap;
   final Function() onEdit;
 
-  String get label => name ?? address;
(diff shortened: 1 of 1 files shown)


mkyq (approved):

lib/src/screens/receive/widgets/address_cell.dart
```diff
@@ -39,7 +39,8 @@ class AddressCell extends StatelessWidget {
   final Function(String) onTap;
   final Function() onEdit;
 
-  String get label => name ?? address;
+  String get label => name.isEmpty?address.substring(0,8)+'...'+
+      address.substring(address.length-8,address.length):name;
```
mkyq, line 43: Please make sure that address is more than 8 chars before cut the string. Make that is less than or equal 8 chars then return full string.

mkyq (changes requested):

mkyq (approved):

PR #284, opened 2022-02-26
Migrations

godilite:
- add migration script
- add new migrations to autogenerated  switch cases
- Run command to generate new migrations `flutter packages pub run tool/create_new_migration.dart`


mkyq:
Please move back current migrations back to default_migration, but please keep script for generation of new migration files and adding them migrations list.

godilite:
@mkyq

PR #290, opened 2022-03-07
Added French localization

hbs:
French mnemonics could be selected but French was not supported as a language. This PR solves this!


mkyq:
@hbs thanks

PR #309, opened 2022-03-29
cw-60: fill amount on scan qr on exchange page

clear-xmr:


lib/utils/payment_request.dart
```diff
@@ -0,0 +1,18 @@
+class PaymentRequest {
+  String address;
+  String amount;
```
mkyq, line 3: Please use `final` for those properties.

mkyq (changes requested):

mkyq (approved):

PR #315, opened 2022-03-31
android: build: add Dockerfile for build environment

clear-xmr:


OmarHatem28:
thanks for opening the PR, but unfortunately, will close it as it's old and we are currently using monero.dart which have different build steps, we will try and update our current docker configs tho

Issue #340, opened 2022-04-15
secure screen


ghost:
Currently, you can record and take screenshots of the app, which isnt very secure. Most banking apps aswell as apps like signal and keepassdx prevent this by having the FLAG_SECURE for all activites. Im not sure if this is possible with flutter, but I think it should be added considering this app contains financial data

SamsungGalaxyPlayer:
I don't think this is a feature we want to add, or at least it should be disabled by default. Will check with the team.

Mohit-Joshi-dev:
@SamsungGalaxyPlayer sure. Thanks for the reply.

BobIsMyManager:
> I don't think this is a feature we want to add, or at least it should be disabled by default. Will check with the team.

What exactly would be the issue if it was enabled by default?

SamsungGalaxyPlayer:
@BobIsMyManager a lot of people like to take screenshots for support and backups. We get a lot of both casual and expert users. Casual users might just take a screenshot of their seed and/or keys.

We also have many people who take screenshots of their receive screen.

BobIsMyManager:
Maybe in the toast that appears when trying to take a screenshot, it can say where to disable it

PR #341, opened 2022-04-16
Blocking screenshots and screenrecording.

Mohit-Joshi-dev:
Fixed:  [https://github.com/cake-tech/cake_wallet/issues/340](https://github.com/cake-tech/cake_wallet/issues/340)


SamsungGalaxyPlayer:
I don't think this is a feature we want to add, or at least it should be disabled by default. Will check with the team.

SamsungGalaxyPlayer:
Blocking screenshots can be an optional feature, but it needs to be disabled by default.

PR #343, opened 2022-04-19
Cw 55 allow saving templateswith fiat amount

Serhii-Borodenko:
Added ability to save templates with fiat amount


lib/src/screens/send/send_page.dart
```diff
@@ -49,6 +53,13 @@ class SendPage extends BasePage {
   @override
   AppBarStyle get appBarStyle => AppBarStyle.transparent;
 
+  @override
+  void onClose(BuildContext context) {
+    settingsViewModel.setFiatCurrency(fiatFromSettings);
+    Navigator.of(context).pop();
+    return print(fiatFromSettings.title);
+  }
+
```
mkyq, line 62: What that code should do ?

mkyq (changes requested):

mkyq:
cc @Serhii-Borodenko

lib/src/screens/send/send_page.dart
Serhii-Borodenko: Removed print statement

Serhii-Borodenko: @mkyq

mkyq (approved):

PR #346, opened 2022-04-21
Select sub address or standard address instead of both.

Krakaw:
### N.B. I don't actually have flutter installed so I didn't test this.

Just some small fixes here, the api dev url was missing the subdomain `a`.

Instead of showing both the subaddress and the standard address, choose just the sub address and if it's not available use the standard address.

### DIFF
--- a/lib/core/yat_service.dart
+++ b/lib/core/yat_service.dart
@@ -25,23 +25,25 @@ class YatService {
   Future<List<YatRecord>> fetchYatAddress(String emojiId, String ticker) async {
     final formattedTicker = ticker.toUpperCase();
     final formattedEmojiId = emojiId.replaceAll(' ', '');
+    final tag = tags[formattedTicker];
     final uri = Uri.parse(lookupEmojiUrl(formattedEmojiId)).replace(
         queryParameters: <String, dynamic>{
-          "tags": tags[formattedTicker]
+          "tags": tag
         });
-
     final yatRecords = <YatRecord>[];
 
     try {
       final response = await get(uri);
       final resBody = json.decode(response.body) as Map<String, dynamic>;
-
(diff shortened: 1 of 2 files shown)


lib/core/yat_service.dart
```diff
@@ -35,9 +37,12 @@ class YatService {
...
       final results = resBody["result"] as Map<dynamic, dynamic>;
-      results.forEach((dynamic key, dynamic value) {
-        yatRecords.add(YatRecord.fromJson(value as Map<String, dynamic>));
-      });
+   
+      // Favour a subaddress over a standard address.
+      final yatRecord = results[MONERO_SUB_ADDRESS] ?? results[MONERO_STD_ADDRESS];
+      if (yatRecord != null) {
+        yatRecords.add(YatRecord.fromJson(value as Map<String, dynamic>))
+      }
```
mkyq, line 45: Here is unclean from where come `value`.

mkyq: Maybe you meant `yatRecord` instead of `value` ?

mkyq: And as i see that `fetchYatAddress` will works only for `xmr` or i miss something ?

mkyq (changes requested):

lib/core/yat_service.dart
Krakaw: > And as i see that `fetchYatAddress` will works only for `xmr` or i miss something ?

It will fetch whatever tags are passed to it, based on if the user is looking up XMR / BTC / LTC. If nothing is passed in it will return all attached wallet addresses.

However if the user has added a standard and a sub address, then you will get back 2 results.

mkyq (approved):

PR #359, opened 2022-05-10
Add additional coins to exchange

SamsungGalaxyPlayer:
Adds the following to the exchange:

APE
AVAXC
BTT
BTTBSC
DOGE
FIRO
HBAR
NANO (re-adds)
SC
SOL
USDC
USDCSOL
USDTTRC20
UST
XVG
ZADDR (shielded Zcash)
ZEC (transparent Zcash)
ZEN

Please test. Should include applicable tags. I also improved the address validator slightly for some assets, including XMR.

There should be no conflicts for these assets between Sideshift and ChangeNOW. Please test that the following work:

USDCSOL (only listed on Sideshift)
FIRO (only listed on ChangeNOW)

Please also especially test ZEC and ZADDR to make sure the tags are correct. ZEC is supported on both, but ZADDR is only supported on Sideshift.


SamsungGalaxyPlayer:
Please remove UST

PR #370, opened 2022-06-01
Added ability sending to FIO address

Serhii-Borodenko:
Added support for sending funds to FIO addresses for the selected asset.

### DIFF
--- a/lib/entities/parse_address_from_domain.dart
+++ b/lib/entities/parse_address_from_domain.dart
@@ -4,6 +4,7 @@ import 'package:cake_wallet/entities/parsed_address.dart';
 import 'package:cake_wallet/entities/unstoppable_domain_address.dart';
 import 'package:cake_wallet/entities/emoji_string_extension.dart';
 import 'package:flutter/foundation.dart';
+import 'package:cake_wallet/entities/fio_address_provider.dart';
 
 class AddressResolver {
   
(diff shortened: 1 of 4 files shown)


lib/entities/parse_address_from_domain.dart
```diff
@@ -26,6 +27,14 @@ class AddressResolver {
 
   Future<ParsedAddress> resolve(String text, String ticker) async {
     try {
+      if(text.contains('@')){
```
SamsungGalaxyPlayer, line 30: Please tighten up this up to not check if the string includes any periods `.` since OpenAlias also supports emails, eg for XMR justin@ehrenhofer.org. It should check `justin@ehrenhofer` for an FIO address, not `justin@ehrenhofer.org`.

lib/entities/parsed_address.dart
```diff
@@ -2,7 +2,7 @@ import 'package:cake_wallet/entities/openalias_record.dart';
 import 'package:cake_wallet/entities/yat_record.dart';
 import 'package:flutter/material.dart';
 
-enum ParseFrom { unstoppableDomains, openAlias, yatRecord, notParsed }
+enum ParseFrom { unstoppableDomains, openAlias, yatRecord,fio, notParsed }
```
SamsungGalaxyPlayer, line 5: missing space?

lib/src/screens/send/widgets/extract_address_from_parsed.dart
```diff
@@ -23,6 +23,11 @@ Future<String> extractAddressFromParsed(
       content = S.of(context).openalias_alert_content(parsedAddress.name);
       address = parsedAddress.addresses.first;
       break;
+    case ParseFrom.fio:
+      title = S.of(context).address_detected;
+      content = S.of(context).openalias_alert_content(parsedAddress.name);
```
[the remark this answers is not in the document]
Serhii-Borodenko: openalias_alert_content is the "You will be sending funds" message. I can rename this variable to unify, I would not like to duplicate it, if there should be a different text, please let me know.

lib/entities/parse_address_from_domain.dart
Serhii-Borodenko: done

mkyq (approved):

PR #384, opened 2022-06-19
Fixed French translations

hbs:
Fixed some French translations which were changed by @paulverbeke.

While the terms "graine" for "seed" and "portefeuille" for "wallet" are lexicographically correct, their use in the context of crypto currencies is far from widespread.

Therefore I have modified the French messages to replace "graine" with "phrase secrète" with "(seed)" added after the first occurrence in the message.

The term "portefeuille" when used in a financial context means "portfolio", therefore I have added "(wallet)" after the first occurrence of "portefeuille" in messages to avoid confusing the users.

Also fixed some minor typoes.


paulverbeke:
That's a great compromise thank you ! I wasn't sure if "seed" and "wallet" could pass the "girlfriend test" 😅. Didn't even noticed that there was an FAQ section 👌

mkyq (approved):

PR #390, opened 2022-06-22
QR code fullscreen

OmarHatem28:


lib/src/screens/receive/fullscreen_qr_page.dart
```diff
@@ -0,0 +1,86 @@
+import 'package:cake_wallet/themes/theme_base.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:cake_wallet/src/screens/base_page.dart';
+
+import 'widgets/qr_image.dart';
```
mkyq, line 6: Please use package import path instead of local path

lib/router.dart
```diff
@@ -402,6 +396,16 @@ Route<dynamic> createRoute(RouteSettings settings) {
               getIt.get<UnspentCoinsDetailsPage>(
                   param1: args));
 
+    case Routes.fullscreenQR:
+      final args = settings.arguments as Map;
```
mkyq, line 400: Add type for map key and values (I assume Map<String, dynamic>)

mkyq (changes requested):

mkyq:
cc @OmarHatem28

PR #392, opened 2022-06-22
Hide keys for BTC & LTC

OmarHatem28:
Remove WIF and Keys from BTC and LTC wallets in the Wallet seed/keys screen

### DIFF
--- a/lib/src/screens/wallet_keys/wallet_keys_page.dart
+++ b/lib/src/screens/wallet_keys/wallet_keys_page.dart
@@ -1,5 +1,4 @@
 import 'package:cake_wallet/utils/show_bar.dart';
-import 'package:cw_core/wallet_type.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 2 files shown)


lib/src/screens/wallet_keys/wallet_keys_page.dart
```diff
@@ -13,7 +13,10 @@ class WalletKeysPage extends BasePage {
   WalletKeysPage(this.walletKeysViewModel);
 
   @override
-  String get title => S.current.wallet_keys;
+  String get title =>
+      walletKeysViewModel.walletType == WalletType.bitcoin || walletKeysViewModel.walletType == WalletType.litecoin
+          ? S.current.wallet_seed
+          : S.current.wallet_keys;
```
mkyq, line 19: Better to move that part to view model and call it from title

mkyq (changes requested):

mkyq (approved):

PR #394, opened 2022-06-23
Enable differenct currencies

OmarHatem28:


OmarHatem28:
@mkyq I noticed that there is already a widget `Picker` that has some of the desired UI while looking into CW-104 and 106, anyway it would have needed some refactoring to allow the required change currency design.
However, I am refactoring `Picker` right now while working on the new design and the new `CurrencyAlertDialog` in this PR will be removed then.

lib/view_model/buy/buy_amount_view_model.dart
```diff
@@ -11,7 +11,8 @@ abstract class BuyAmountViewModelBase with Store {
   @observable
   String amount;
 
-  FiatCurrency get fiatCurrency => FiatCurrency.usd;
+  @observable
+  FiatCurrency fiatCurrency = FiatCurrency.usd;
```
mkyq, line 15: set value inside of constructor

lib/src/widgets/currency_alert_dialog.dart
```diff
@@ -0,0 +1,101 @@
...
+        child: Container(
+          color: Colors.transparent,
+          child: Container(
+            color: Color(0xff010615).withOpacity(0.8),
+            child: Center(
+              child: Column(
+                children: [
+                  Expanded(child: const SizedBox()),
+                  Text(
+                    "Change Currency",
```
mkyq, line 26: Add to localization file

lib/src/widgets/currency_alert_dialog.dart
```diff
@@ -0,0 +1,101 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      backgroundColor: Colors.transparent,
+      body: GestureDetector(
+        onTap: () => Navigator.of(context).pop(),
+        child: Container(
+          color: Colors.transparent,
+          child: Container(
+            color: Color(0xff010615).withOpacity(0.8),
```
mkyq, line 20: Please move all unnamed colors into theme

mkyq (changes requested):

PR #398, opened 2022-06-28
Rework settings pickers

OmarHatem28:
based on #397


lib/src/screens/settings/widgets/settings_choices_cell.dart
```diff
@@ -0,0 +1,71 @@
...
+              scrollDirection: Axis.horizontal,
+              child: Container(
+                decoration: BoxDecoration(
+                  borderRadius: BorderRadius.circular(30),
+                  color: Color(0xffF2F0FA),
+                ),
+                child: Row(
+                  mainAxisAlignment: MainAxisAlignment.center,
+                  children: choicesListItem.items.map((dynamic e) {
+                    final bool isSelected = choicesListItem.selectedItem == e;
```
mkyq, line 42: remove type definition for `final`

lib/src/screens/settings/widgets/settings_choices_cell.dart
```diff
@@ -0,0 +1,71 @@
...
+                      child: Container(
+                        padding: EdgeInsets.symmetric(horizontal: 32, vertical: 8),
+                        decoration: BoxDecoration(
+                          borderRadius: BorderRadius.circular(30),
+                          color: isSelected ? Theme.of(context).accentTextTheme.body2.color : null,
+                        ),
+                        child: Text(
+                          choicesListItem.displayItem?.call(e) ?? e.toString(),
+                          style: TextStyle(
+                            color: isSelected ? Colors.white : Theme.of(context).primaryTextTheme.caption.color,
```
mkyq, line 56: Better to use color from theme here too

mkyq (changes requested):

lib/src/screens/settings/widgets/settings_choices_cell.dart
OmarHatem28: why? it's more readable this way

mkyq: It's final and value will be not changed and definition type will not be changed in runtime too, as it's clean from right side expression that it's bool.

lib/src/screens/settings/widgets/settings_choices_cell.dart
OmarHatem28: it's white in all 3 themes.
as far as i understand the theme's structure in this project, moving it to themes would force us to declare a new styling element with the same color in the 3 themes.

mkyq (approved):

PR #469, opened 2022-08-11
sort Giftcards alphabetically

Serhii-Borodenko:
sort Giftcards alphabetically


Serhii-Borodenko:
Based on the service documentation, there is no method to get a sorted list of merchants

mkyq (approved):

PR #511, opened 2022-09-16
Truncate to double

godilite:


mkyq:
Please make `remaining` as string value and make it fixed in view model while calculation, for avoid case of different value in model and view, specially for be sure that you will use exactly right remaining value for another operation where is require that data

mkyq:
@godilite cc

mkyq (changes requested):
@godilite see comments above

mkyq (approved):

PR #525, opened 2022-09-30
[CW-175] refactoring of crypto currency list

Serhii-Borodenko:


ProkhorZ:
This looks much cleaner indeed, but mapFromString seems to map from integers now? Let me know if I misunderstand this, I'm just passing by.

Serhii-Borodenko:
> This looks much cleaner indeed, but mapFromString seems to map from integers now? Let me know if I misunderstand this, I'm just passing by.

Yes, there is a mistake here, thanks for letting me know.

Serhii-Borodenko:
> This looks much cleaner indeed, but mapFromString seems to map from integers now? Let me know if I misunderstand this, I'm just passing by.

Yes, there is a mistake here, thanks for letting me know.

OmarHatem28 (approved):

mkyq:
@Serhii-Borodenko @OmarHatem28 We have `all` list of currencies, yes for optimize searching of by raw we need to get map for faster search. So let generate that map based on `all` list, as every currency has raw property which you can use as key for that map, for example you can just use `all.fold<Map<int, CryptoCurrency>>(<int, CryptoCurrency>{}, (acc, item) => ...)` for generate `mapFromInt` and i would pick better name for that static variable.

Serhii-Borodenko:
fixed the file, does it make sense to change `mapFromString` as well, for example, add a parameter to the `CryptoCurrency` class

OmarHatem28:
yeah, I would go with adding another attribute to the`CryptoCurrency` class and using it instead of writing this Map on its own, also you can make those 2 Maps private to avoid any miss-access for them since they are only used to get the CryptoCurrency from the 2 `from` functions

OmarHatem28:
@mkyq I checked this option, but nope, it's not consistent among all currencies, because some of them would require the title + tag and some won't require the tag, so it's not very dependable and might change

mkyq:
@OmarHatem28 @Serhii-Borodenko then need as you wrote above, need to add another required property for CryptoCurrency, then fill it for all currencies and then generate `mapFromString` based on that value

Serhii-Borodenko:
It should be a property "name", I will rename an already existing property to the "full name" since it is used to search in the currency picker.

Serhii-Borodenko:
@OmarHatem28 Cant add a required parameter to Hive adapter in `CryptoCurrency` class, only optional parameter "name". Is it safe enough?

OmarHatem28:
@mkyq I don't see any use for Hive in `CryptoCurrency` class, is it there for a reason?

if not, I would say we remove Hive and its Adapter from this class and I would suggest we make `title`, `raw`, and `name` required

ProkhorZ:
@OmarHatem28 I don't see it in the code here, but [Hive](https://hive.io/) is a cryptocurrency. It can be exchanged for Monero on the swap site [BlockTrades](https://blocktrades.us/en/trade).

OmarHatem28:
Hi @ProkhorZ we are talking about another [Hive](https://pub.dev/packages/hive?fbclid=IwAR37hZk3zld-EomtQrtgBCXtuxy0QkmBEgcKc6aPpr9ml5C0al09iwvV3YQ) it's for storing data locally on mobile devices, but thanks for the note, and please feel free to open a ticket for adding Hive (the cryptocurrency) to Cake Wallet

cw_core/lib/crypto_currency.dart
```diff
@@ -70,339 +67,100 @@ class CryptoCurrency extends EnumerableItem<int> with Serializable<int> {
...
+  static const stx = CryptoCurrency(title: 'STX', iconPath: 'assets/images/stx_icon.png', raw: 61, name: 'stx');
 
+  static final Map<int, CryptoCurrency> _rawCurrencyMap =
+  all.fold<Map<int, CryptoCurrency>>(<int, CryptoCurrency>{}, (acc, item) {
+    acc.addAll({item.raw: item});
+    return acc;
+  });
 
+  static final Map<String?, CryptoCurrency> _nameCurrencyMap =
+  all.fold<Map<String?, CryptoCurrency>>(<String?, CryptoCurrency>{}, (acc, item) {
```
OmarHatem28, line 143: Keys here cannot be nullable since the name field is required so it would be better to remove nullability

cw_core/lib/crypto_currency.dart
```diff
@@ -70,339 +67,100 @@ class CryptoCurrency extends EnumerableItem<int> with Serializable<int> {
...
+  static const xbtc = CryptoCurrency(title: 'XBTC', tag: 'XHV', raw: 20, name: 'xbtc');
+  static const xcad = CryptoCurrency(title: 'XCAD', tag: 'XHV', raw: 21, name: 'xcad');
+  static const xchf = CryptoCurrency(title: 'XCHF', tag: 'XHV', raw: 22, name: 'xchf');
+  static const xcny = CryptoCurrency(title: 'XCNY', tag: 'XHV', raw: 23, name: 'xcny');
+  static const xeur = CryptoCurrency(title: 'XEUR', tag: 'XHV', raw: 24, name: 'xeur');
+  static const xgbp = CryptoCurrency(title: 'XGBP', tag: 'XHV', raw: 25, name: 'xgbp');
+  static const xjpy = CryptoCurrency(title: 'XJPY', tag: 'XHV', raw: 26, name: 'xjpy');
+  static const xnok = CryptoCurrency(title: 'XNOK', tag: 'XHV', raw: 27, name: 'xnok');
+  static const xnzd = CryptoCurrency(title: 'XNZD', tag: 'XHV', raw: 28, name: 'xnzd');
+  static const xusd = CryptoCurrency(title: 'XUSD', tag: 'XHV', raw: 29, name: 'xusd');
```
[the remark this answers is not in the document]
Serhii-Borodenko: I believe that these assets only match Haven сurrency and are not separate.

cw_core/lib/crypto_currency.dart
Serhii-Borodenko: @OmarHatem28 done

OmarHatem28 (approved):

PR #532, opened 2022-10-12
Cw 181 simple swap extra id fix

godilite:


SamsungGalaxyPlayer:
Is this meant to address issues with Stellar and Ripple?

lib/exchange/simpleswap/simpleswap_exchange_provider.dart
```diff
@@ -106,14 +106,15 @@ class SimpleSwapExchangeProvider extends ExchangeProvider {
     final id = responseJSON['id'] as String;
     final inputAddress = responseJSON['address_from'] as String;
     final settleAddress = responseJSON['user_refund_address'] as String;
-
+    final extraId = responseJSON['extra_id_from'] as String;
```
mkyq, line 109: cast extraId as optional String

mkyq (changes requested):

lib/exchange/simpleswap/simpleswap_exchange_provider.dart
```diff
@@ -188,13 +187,15 @@ class SimpleSwapExchangeProvider extends ExchangeProvider {
     final to = CryptoCurrency.fromString(toCurrency);
     final inputAddress = responseJSON['address_from'] as String;
     final expectedSendAmount = responseJSON['expected_amount'].toString();
+    final extraId = responseJSON['extra_id_from'] as String;
```
mkyq, line 190: Could be extraId optional here ? Just if there could be null it will throw exception on null, so if it's not required field better to cast to optional String

lib/exchange/simpleswap/simpleswap_exchange_provider.dart
```diff
@@ -61,7 +60,7 @@ class SimpleSwapExchangeProvider extends ExchangeProvider {
 
       if (response.body == null || response.body == "null") return 0.00;
       final data = json.decode(response.body) as String;
-    
+
```
mkyq, line 63: Please don't use auto formatter for unrelated changes.

mkyq (changes requested):

mkyq (approved):

mkyq (approved):

PR #553, opened 2022-10-21
Fix gift card purchase

godilite:


lib/anypay/any_pay_payment_instruction.dart
```diff
@@ -17,8 +17,8 @@ class AnyPayPaymentInstruction {
 		return AnyPayPaymentInstruction(
 			type: obj['type'] as String,
 			requiredFeeRate: obj['requiredFeeRate'] as int,
-			txKey: obj['tx_key'] as bool,
-			txHash: obj['tx_hash'] as bool,
+			txKey: obj['tx_key'] != null ? obj['tx_key'] as bool : false,
+			txHash: obj['tx_hash']  != null ? obj['tx_hash'] as bool : false,
```
mkyq, line 21: Better to cast to optional bool and then use logical or to false

mkyq (changes requested):

PR #568, opened 2022-10-26
Update flushbar

OmarHatem28:


lib/src/screens/new_wallet/new_wallet_type_page.dart
```diff
@@ -67,8 +58,6 @@ class WalletTypeFormState extends State<WalletTypeForm> {
 
   WalletType? selected;
   List<WalletType> types;
-  // FIX-ME: Replace Flushbar
-  // Flushbar<void>? _progressBar;
```
OmarHatem28, line 71: I took a look at the history of this class and found that the usage for the `_progressBar` was moved to another class, so I removed it since it's no longer needed here

lib/view_model/auth_view_model.dart
```diff
@@ -55,8 +56,10 @@ abstract class AuthViewModelBase with Store {
     final isSuccessfulAuthenticated = await _authService.authenticate(password);
 
     if (isSuccessfulAuthenticated) {
-      state = ExecutedSuccessfullyState();
-      _failureCounter = 0;
+      WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+        state = ExecutedSuccessfullyState();
+        _failureCounter = 0;
+      });
```
OmarHatem28, line 62: ensure that any build/navigation caused by the previous state is finished first

mkyq (approved):

PR #569, opened 2022-10-26
Xchaha20 poly1305 integration

mkyq:


lib/src/screens/dashboard/wallet_menu.dart
```diff
@@ -52,6 +52,22 @@ class WalletMenu {
...
+          height: 16,
+          width: 16,
+          color: Palette.darkBlue),
+        handler: () {
+          Navigator.of(context).pushNamed(
+            Routes.auth,
+            arguments: (bool isAuthenticatedSuccessfully, AuthPageState auth) {
+              if (isAuthenticatedSuccessfully) {
+                auth.close();
+                Navigator.of(auth.context).pushNamed(Routes.backup);
```
OmarHatem28, line 67: this will break with the new flushbar package
if this will be merged first, then we will modify it in CW 186
else use `auth.close(route:Routes.backup)` instead after CW-186 is merged

OmarHatem28 (approved):
Added one minor comment, everything else seems LEGIT

PR #579, opened 2022-10-30
show warning if exchanging with slow fee

Serhii-Borodenko:
added a pop-up when exchanging with a slow fee with the ability to change it


lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -611,4 +631,19 @@ abstract class ExchangeViewModelBase with Store {
         selectedProviders
             .where((provider) => providersForCurrentPair().contains(provider)));
   }
+
+  @action
+  TransactionPriority? setDefaultTransactionPriority() {
```
OmarHatem28, line 636: The return type could be void.
Also, I checked the ticket, and yes the priorities are as set to default as the ticket description, but the default transaction priority for haven and monero is different in the wallet code, which one should be the default?

Reference:
`getDefaultTransactionPriority()` in `lib/haven/cw_haven.dart` and in `lib/monero/cw_monero.dart`

mkyq: Better to create new task for change default transaction priorities for haven and monero wallets

OmarHatem28 (approved):

PR #595, opened 2022-11-04
CW-118 Allow disabling of fiat

Serhii-Borodenko:
For Disable fiat mode:
Stoped connecting to the Cake fiat proxy
Removed fiat balances from the Home screen
Removed fiat line from Send screen
Removed fiat line from Transaction screen
Remove fiat balances from Send/Exchange confirmation screens
Remove fiat settings from Settings screen


lib/src/screens/send/widgets/send_card.dart
```diff
@@ -438,8 +440,8 @@ class SendCardState extends State<SendCard>
                                           Padding(
                                               padding:
                                               EdgeInsets.only(top: 5),
-                                              child: Text(
-                                                  output
+                                              child: sendViewModel.balanceViewModel.disableFiat ?
+                                              Container () : Text(output
```
OmarHatem28, line 444: you can add the if condition here too, then it won't have to build any unused widgets (Padding, Container).

if removing the padding will affect the UI then you can replace the container with `const SizedBox()`

Serhii-Borodenko: fixed

OmarHatem28 (approved):

lib/src/screens/send/send_page.dart
```diff
@@ -386,16 +386,14 @@ class SendPage extends BasePage {
...
+                            ' ' + sendViewModel.fiat.title,
                     fee: S.of(context).send_fee,
                     feeValue: sendViewModel.pendingTransaction!.feeFormatted,
-                    feeFiatAmount:
-                        sendViewModel.pendingTransactionFeeFiatAmount +
-                            ' ' +
-                            sendViewModel.fiat.title,
+                    feeFiatAmount: sendViewModel.balanceViewModel.disableFiat
+                        ? '' : sendViewModel.pendingTransactionFeeFiatAmount +
+                            ' ' + sendViewModel.fiat.title,
```
mkyq, line 396: Better to create new getter like `pendingTransactionFiatAmountFormatted` which will return pendingTransactionFiatAmount + ' ' +  fiat.title and if fiat api is disabled then return empty line.
Same for `pendingTransactionFeeFiatAmount` for send view model.

lib/src/screens/exchange_trade/exchange_trade_page.dart
```diff
@@ -378,9 +381,11 @@ class ExchangeTradeState extends State<ExchangeTradeForm> {
                           });
                     },
                     actionLeftButton: () => Navigator.of(context).pop(),
-                    feeFiatAmount: widget.exchangeTradeViewModel.sendViewModel.pendingTransactionFeeFiatAmount
+                    feeFiatAmount: widget.disableFiat ? ''
+                        : widget.exchangeTradeViewModel.sendViewModel.pendingTransactionFeeFiatAmount
                         +  ' ' + widget.exchangeTradeViewModel.sendViewModel.fiat.title,
-                    fiatAmountValue: widget.exchangeTradeViewModel.sendViewModel
+                    fiatAmountValue: widget.disableFiat ? ''
+                        : widget.exchangeTradeViewModel.sendViewModel
```
mkyq, line 388: Same as comments for send view model and pendingTransactionFeeFiatAmount, need to add formatted getter which will contains logic for formatted pendingTransactionFeeFiatAmount and logic for case when fiat is disabled.

lib/reactions/fiat_rate_update.dart
```diff
@@ -27,7 +27,9 @@ Future<void> startFiatRateUpdate(AppStore appStore, SettingsStore settingsStore,
           if (appStore.wallet!.type == WalletType.haven) {
             await updateHavenRate(fiatConversionStore);
           } else {
-            fiatConversionStore.prices[appStore.wallet!.currency] = await FiatConversionService.fetchPrice(
+            fiatConversionStore.prices[appStore.wallet!.currency] =
+            settingsStore.shouldDisableFiat ? 0.0
+                : await FiatConversionService.fetchPrice(
```
mkyq, line 32: Let make it more clean with if else instead of ternary operator.

mkyq, line 32: Here no case for disable for fetch rate update for haven. I would to force stop any calls for updating the rate for haven or another wallet types above on call stack and reset price for all currencies to 0.

mkyq: So better check is fiat disabled above in timer callback function and if it's disabled then reset prices and return from the function and below keep fetching as it was before.

mkyq (changes requested):
`FiatConversionService` calls are in different places of the code base. Please check all calls of this. Transform static FiatConversionService calls into singleton, but with dependency on SettingsStore and check inside of calling `fetchPrice` is disabled fiat or not, if it's disabled just return 0.0 or if it's not disabled then run `_fetchPrice`.

mkyq:
cc @Serhii-Borodenko @OmarHatem28 
Also need to force update prices on changing is disabled fiat or not (if disabled and force update it should returns 0.0 as I wrote about FiatConversionService behavior with SettingsStore dependency and disabledFiat property).

lib/reactions/fiat_rate_update.dart
Serhii-Borodenko: @mkyq @OmarHatem28 Didn't find any use for `fiat rate_update.dart`, fiat rate is no longer updated by timer. Do we need to restore this feature?
Currently `FiatConversionService` is used in `startCurrentFiatChangeReaction` and `startCurrentWalletChangeReaction`.

lib/view_model/send/send_view_model.dart
```diff
@@ -142,6 +142,16 @@ abstract class SendViewModelBase with Store {
...
   String get balance => balanceViewModel.availableBalance;
 
+  @computed
+  String get pendingTransactionFiatAmountFormatted =>
+      balanceViewModel.disableFiat ? '' : pendingTransactionFiatAmount +
+          ' ' + fiat.title;
+
+  @computed
+  String get pendingTransactionFeeFiatAmountFormatted =>
+      balanceViewModel.disableFiat ? '' : pendingTransactionFeeFiatAmount +
```
OmarHatem28, line 152: we can use this getter here too when defined

lib/src/screens/send/widgets/send_card.dart
```diff
@@ -331,6 +331,7 @@ class SendCardState extends State<SendCard>
                             ],
                           ),
                         )),
+                    if (sendViewModel.balanceViewModel.disableFiat)
```
OmarHatem28, line 334: shouldn't this be negated? i.e (`if (!sendViewModel.balanceViewModel.disableFiat)`)

lib/src/screens/send/widgets/send_card.dart
```diff
@@ -437,8 +438,9 @@ class SendCardState extends State<SendCard>
                                           Padding(
                                               padding:
                                               EdgeInsets.only(top: 5),
-                                              child: Text(
-                                                  output
+                                              child: sendViewModel.balanceViewModel.disableFiat
```
OmarHatem28, line 441: also this can use the getter that will be defined in the `sendViewModel`

lib/view_model/exchange/exchange_trade_view_model.dart
```diff
@@ -72,6 +72,18 @@ abstract class ExchangeTradeViewModelBase with Store {
...
 
+  @computed
+  String get pendingTransactionFiatAmountValueFormatted =>
+      sendViewModel.balanceViewModel.disableFiat
+          ? '' : sendViewModel.pendingTransactionFiatAmount
+          + ' ' + sendViewModel.fiat.title;
+
+  @computed
+  String get pendingTransactionFeeFiatAmountFormatted =>
+      sendViewModel.balanceViewModel.disableFiat
```
OmarHatem28, line 83: we can create a computed getter in the `sendViewModel` instead so it can be easily tracked

OmarHatem28 (changes requested):

res/values/strings_en.arb
```diff
@@ -654,12 +654,13 @@
   "use_suggested": "Use Suggested",
   "do_not_share_warning_text" : "Do not share these with anyone else, including support.\n\nThey will steal your money!",
   "help": "help",
+  "disable_fiat": "Disable fiat",
```
OmarHatem28, line 657: replace this with `Fiat API`

OmarHatem28 (changes requested):

OmarHatem28 (approved):

PR #596, opened 2022-11-04
[CW-215] switching to haven resets currency to usd

godilite:


lib/reactions/on_current_wallet_change.dart
```diff
@@ -68,6 +68,12 @@ void startCurrentWalletChangeReaction(AppStore appStore,
       if (wallet.type == WalletType.haven) {
         settingsStore.fiatCurrency = FiatCurrency.usd;
         await updateHavenRate(fiatConversionStore);
+      }else{
+        final savedFiat = await getIt
+          .get<SharedPreferences>().getString(PreferencesKey.currentFiatCurrencyKey);
+          if(savedFiat != null){
+            settingsStore.fiatCurrency = FiatCurrency.deserialize(raw: savedFiat);
+          }
```
OmarHatem28, line 76: CW-215 issue was that after switching to Haven wallet, the fiat currency gets reset to USD, and it's because of line 69, shouldn't removing this line solve the issue?
Is there a reason why we are saving the currentHavenFiatCurrency specifically?

OmarHatem28, line 76: CW-215 issue was that after switching to Haven wallet, the fiat currency gets reset to USD, and it's because of line 69, shouldn't removing this line solve the issue?
Is there a reason why we are saving the currentHavenFiatCurrency specifically?

OmarHatem28: ok, and we are saving also the haven currency so that if the user changed haven from USD, he can still see his choice when he gets back to haven, right?

OmarHatem28: because in the current implementation it will only use the saved haven currency on loading wallet, but not switching back to haven

PR #598, opened 2022-11-04
Add default values for hive fields for Node, UnspentCoinsInfo, Wallet…

mkyq:
…Info, Order, Contact, TransactionDescription, Trade

### DIFF
--- a/cw_core/lib/wallet_info.dart
+++ b/cw_core/lib/wallet_info.dart
@@ -33,31 +33,31 @@ class WalletInfo extends HiveObject {
   static const typeId = 4;
   static const boxName = 'WalletInfo';
 
-  @HiveField(0)
+  @HiveField(0, defaultValue: '')
   String id;
 
-  @HiveField(1)
+  @HiveField(1, defaultValue: '')
   String name;
 
   @HiveField(2)
   WalletType type;
 
-  @HiveField(3)
+  @HiveField(3, defaultValue: false)
   bool isRecovery;
 
-  @HiveField(4)
+  @HiveField(4, defaultValue: 0)
   int restoreHeight;
 
-  @HiveField(5)
+  @HiveField(5, defaultValue: 0)
   int timestamp;
 
-  @HiveField(6)
+  @HiveField(6, defaultValue: '')
   String dirPath;
 
-  @HiveField(7)
+  @HiveField(7, defaultValue: '')
   String path;
 
-  @HiveField(8)
+  @HiveField(8, defaultValue: '')
   String address;
 
   @HiveField(10)
--- a/lib/buy/order.dart
+++ b/lib/buy/order.dart
@@ -30,10 +30,10 @@ class Order extends HiveObject {
   static const boxName = 'Orders';
   static const boxKey = 'ordersBoxKey';
 
-  @HiveField(0)
+  @HiveField(0, defaultValue: '')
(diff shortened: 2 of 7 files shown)


cw_core/lib/wallet_info.dart
```diff
@@ -33,31 +33,31 @@ class WalletInfo extends HiveObject {
...
-  @HiveField(0)
+  @HiveField(0, defaultValue: '')
   String id;
 
-  @HiveField(1)
+  @HiveField(1, defaultValue: '')
   String name;
 
   @HiveField(2)
   WalletType type;
```
OmarHatem28, line 43: is this guaranteed it won't be null?

lib/buy/order.dart
```diff
@@ -42,24 +42,24 @@ class Order extends HiveObject {
   @HiveField(3)
   String? to;
 
-  @HiveField(4)
+  @HiveField(4, defaultValue: '')
   late String stateRaw;
```
OmarHatem28, line 46: with the current implementation of the constructor, this variable may not be initialized if the state is null, resulting in a `late variable not initialized` exception

lib/buy/order.dart
```diff
@@ -42,24 +42,24 @@ class Order extends HiveObject {
...
+  @HiveField(7, defaultValue: '')
   String receiveAddress;
 
-  @HiveField(8)
+  @HiveField(8, defaultValue: '')
   String walletId;
 
-  @HiveField(9)
+  @HiveField(9, defaultValue: 0)
   late int providerRaw;
```
OmarHatem28, line 63: with the current implementation of the constructor, this variable may not be initialized if the provider is null, resulting in a `late variable not initialized` exception

lib/entities/contact.dart
```diff
@@ -17,13 +17,13 @@ class Contact extends HiveObject with Keyable {
...
+  @HiveField(0, defaultValue: '')
   String name;
 
-  @HiveField(1)
+  @HiveField(1, defaultValue: '')
   String address;
 
-  @HiveField(2)
+  @HiveField(2, defaultValue: 0)
   late int raw;
```
OmarHatem28, line 27: with the current implementation of the constructor, this variable may not be initialized if the type is null, resulting in a `late variable not initialized` exception

lib/exchange/trade.dart
```diff
@@ -40,26 +40,26 @@ class Trade extends HiveObject {
   static const boxName = 'Trades';
   static const boxKey = 'tradesBoxKey';
 
-  @HiveField(0)
+  @HiveField(0, defaultValue: '')
   String id;
 
-  @HiveField(1)
+  @HiveField(1, defaultValue: 0)
   late int providerRaw;
```
OmarHatem28, line 47: same

lib/exchange/trade.dart
```diff
@@ -40,26 +40,26 @@ class Trade extends HiveObject {
...
 
-  @HiveField(3)
+  @HiveField(3, defaultValue: 0)
   late int toRaw;
 
   CryptoCurrency get to => CryptoCurrency.deserialize(raw: toRaw);
 
-  @HiveField(4)
+  @HiveField(4, defaultValue: '')
   late String stateRaw;
```
OmarHatem28, line 63: same

PR #601, opened 2022-11-05
[CW-159] filter contacts and wallets by type

Serhii-Borodenko:
filtered contacts and wallets in the address book by the type of the selected asset
on the Send and Exchange screens


lib/src/screens/contact/contact_list_page.dart
```diff
@@ -60,11 +62,13 @@ class ContactListPage extends BasePage {
...
   Widget body(BuildContext context) {
+    final contacts =
+    contactListViewModel.getContacts(isEditable, selectedCurrency);
+    final walletContacts =
+    contactListViewModel.getWallets(isEditable, selectedCurrency);
     return Container(
         padding: EdgeInsets.only(top: 20.0, bottom: 20.0),
-        child: Observer(
-          builder: (_) {
-            return CollapsibleSectionList(
```
OmarHatem28, line 67: There is a bug caused by removing this `Observer` try deleting a contact and it will still be there until you refresh the address book screen.
`Observer` should still be there and you can define `contacts` and `walletContacts` inside the observer so they are refreshed too or if you have another better implementation

lib/router.dart
```diff
@@ -298,11 +299,14 @@ Route<dynamic> createRoute(RouteSettings settings) {
 
     case Routes.addressBook:
       return MaterialPageRoute<void>(
-          builder: (_) => getIt.get<ContactListPage>(param1: true));
+          builder: (_) =>
+              getIt.get<ContactListPage>(param1: true, param2: null));
```
OmarHatem28, line 303: **minor**
you can ignore passing it at all, it would be still parsed as null too

OmarHatem28 (changes requested):

lib/src/screens/contact/contact_list_page.dart
Serhii-Borodenko: yes `Observer` should be there, PR updated manually possibly removed by mistake.

lib/router.dart
Serhii-Borodenko: fixed it

OmarHatem28 (approved):

lib/view_model/contact_list/contact_list_view_model.dart
```diff
@@ -44,4 +45,20 @@ abstract class ContactListViewModelBase with Store {
...
+    return contacts;
+  }
+
+  List<WalletContact> getWallets(bool isEditable, CryptoCurrency? cur) {
+    if (!isEditable) {
+      return walletContacts.where((element) => element.type == cur).toList();
+    }
+
+    return walletContacts;
+  }
```
mkyq, line 63: Could be here case when it's `isEditable` but we need to get only special currency filtered contacts ? Or maybe make sense to just check for is `null` for `cur` here ?

mkyq, line 63: Also could we make computable trigger to get this automatically and don't call those methods from the widget build and then use those computable getters inside of observers ?

mkyq (changes requested):

lib/view_model/contact_list/contact_list_view_model.dart
OmarHatem28: computed will only work with getters, so no params can be passed to it, so it's either a function as it is now and contacts will trigger the observer to rebuild thus getting the new contacts
else we can add these 2 params as attributes in the view model and make `getContacts` a computed getter

mkyq: Yep, i meant that make currency and isEditable as part of the view model and then make computed values for contacts and wallets based on that.

Serhii-Borodenko: I also removed  `isEditable`  for contact_list_page.dart

lib/view_model/contact_list/contact_list_view_model.dart
```diff
@@ -44,4 +45,20 @@ abstract class ContactListViewModelBase with Store {
...
+
+    return contacts;
+  }
+
+  List<WalletContact> getWallets(CryptoCurrency? cur) {
+    if (cur != null) {
+      return walletContacts.where((element) => element.type == cur).toList();
+    }
+
+    return walletContacts;
```
OmarHatem28, line 62: let's add the CryptoCurrency to the view model itself and pass it to the view model instead of the page, and we can then make `getContacts` and `getWallets` as getters and annotate them with `@computed` and they would return all the contacts if the currency is null
i.e
`List<ContactRecord> get getContacts => contacts.where((element) => currency == null || element.type == currency).toList();`

OmarHatem28 (changes requested):

lib/view_model/contact_list/contact_list_view_model.dart
Serhii-Borodenko: @OmarHatem28 I did not find easy way to pass `Currency` to the view model because it is passed to the `ContactListPage `from the `router` depending on where it is called.

OmarHatem28: why? we can pass it from the `ContactListPage` factory to the `SendViewModel` like this
`getIt.registerFactoryParam<ContactListPage, CryptoCurrency?, void>((CryptoCurrency? cur, _)
      => ContactListPage(getIt.get<ContactListViewModel>(param1: cur)));`

mkyq (approved):

PR #603, opened 2022-11-07
rework confirm sending screen for cake pay

Serhii-Borodenko:
reworked confirm sending screen for cake pay


lib/src/widgets/base_alert_dialog.dart
```diff
@@ -11,6 +11,10 @@ class BaseAlertDialog extends StatelessWidget {
   VoidCallback get actionLeft => () {};
   VoidCallback get actionRight => () {};
   bool get barrierDismissible => true;
+  Color? get leftActionButtonTextColor => null;
+  Color? get rightActionButtonTextColor => null;
```
OmarHatem28, line 15: not used

Serhii-Borodenko: used [here](https://github.com/cake-tech/cake_wallet/pull/603/files#diff-4513851fdf25d22afd919464aeb9dac12860a3d38e29edd084873a1656616376R321-R323) to change default parameters

OmarHatem28: I mean its value is not used in the widget (line 93 should be `rightActionButtonTextColor ?? etc..`)

Serhii-Borodenko: Yes, indeed, fixed it. Also fixed the text color, button layout and visibility of the sending address.

OmarHatem28 (approved):

PR #623, opened 2022-11-13
CW-255 SideShift confirm sending screen bug

Serhii-Borodenko:
fix fetching trade state for SideShift, After clicking the line on the transaction history screen for the SideShift provider,
trade detail screen loads and then disappears.


lib/exchange/sideshift/sideshift_exchange_provider.dart
```diff
@@ -250,8 +250,10 @@ class SideShiftExchangeProvider extends ExchangeProvider {
     final deposits = responseJSON['deposits'] as List?;
     TradeState? state;
 
-    if (deposits != null && deposits.isNotEmpty) {
-      final status = deposits[0]['status'] as String;
+    if (deposits != null && deposits.isEmpty) {
+      state = TradeState.deserialize(raw: 'created');
+    } else {
+      final status = deposits?[0]['status'] as String;
```
OmarHatem28, line 256: this should also handle nullability (i.e `as String?`) or check if `deposits` is not null first.

OmarHatem28 (changes requested):

PR #625, opened 2022-11-14
[CW-122] rework filter on the transactions list screen

Serhii-Borodenko:
rework filter on the transactions list screen


lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -60,24 +60,38 @@ abstract class DashboardViewModelBase with Store {
...
       S.current.trades: [
         FilterItem(
-            value: () => tradeFilterStore.displayChangeNow,
+            value: tradeFilterStore.displayAllTrades,
+            caption: 'S.current.all_trades',//Fixme
+            onChanged: () => tradeFilterStore
+                .toggleDisplayExchange(ExchangeProviderDescription.all)),
+        FilterItem(
+            value: tradeFilterStore.displayChangeNow,
             caption: 'Change.NOW',
```
SamsungGalaxyPlayer, line 86: This should be changed to `ChangeNOW`

Serhii-Borodenko: done

lib/store/dashboard/transaction_filter_store.dart
```diff
@@ -1,21 +1,20 @@
...
+  TransactionFilterStoreBase();
 
-  @observable
-  bool displayIncoming;
-
-  @observable
-  bool displayOutgoing;
+  Observable<bool> displayAll = Observable(true);
+  Observable<bool> displayIncoming = Observable(true);
+  Observable<bool> displayOutgoing = Observable(true);
```
OmarHatem28, line 17: same, annotation would be better, instead of having to access it with `.value`

lib/store/dashboard/transaction_filter_store.dart
```diff
@@ -24,10 +23,40 @@ abstract class TransactionFilterStoreBase with Store {
...
+  void toggleIAll() {
+    displayAll.value = (!displayAll.value);
+    if (displayAll.value) {
+      displayOutgoing.value = true;
+      displayIncoming.value = true;
+    }
+    if (!displayAll.value) {
+      displayOutgoing.value = false;
+      displayIncoming.value = false;
+    }
```
OmarHatem28, line 35: assign with `displayAll` instead of if

lib/store/dashboard/transaction_filter_store.dart
```diff
@@ -24,10 +23,40 @@ abstract class TransactionFilterStoreBase with Store {
...
 
   @action
-  void toggleOutgoing() => displayOutgoing = !displayOutgoing;
+  void toggleIncoming() {
+    displayIncoming.value = (!displayIncoming.value);
+    if (displayIncoming.value && displayOutgoing.value) {
+      displayAll.value = true;
+    }
+    if (!displayIncoming.value || !displayOutgoing.value) {
+      displayAll.value = false;
```
OmarHatem28, line 45: computed `displayAll`

lib/store/dashboard/transaction_filter_store.dart
OmarHatem28: also, I would suggest making those private with getters only, so they won't get mistakenly modified from outside

lib/store/dashboard/transaction_filter_store.dart
```diff
@@ -24,10 +23,40 @@ abstract class TransactionFilterStoreBase with Store {
   DateTime? endDate;
 
   @action
-  void toggleIncoming() => displayIncoming = !displayIncoming;
+  void toggleIAll() {
```
OmarHatem28, line 26: Typo in name

OmarHatem28 (changes requested):

res/values/strings_zh.arb
```diff
@@ -652,6 +652,8 @@
   "use_suggested": "使用建议",
   "do_not_share_warning_text" : "请勿与其他任何人分享这些信息，包括支持人员。\n\n您的资金可能而且将会被盗！",
   "help": "帮助",
+  "all_transactions": "所有交易",
+  "all_trades": "所有的变化",
```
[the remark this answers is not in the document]
Serhii-Borodenko: I'm not sure, but there is no difference in translation

lib/src/widgets/standard_list_row.dart
```diff
@@ -1,8 +1,8 @@
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 
-class StandartListRow extends StatelessWidget {
-  StandartListRow(
+class AnotherStandardListRow extends StatelessWidget {
+  AnotherStandardListRow(
```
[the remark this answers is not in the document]
Serhii-Borodenko: The same name is also in `standard_list.dart`. We had two classes StandardListRaw (typo) and StandardListRaw, which is a bit confusing. Maybe they need to be merged

lib/store/dashboard/transaction_filter_store.dart
Serhii-Borodenko: @OmarHatem28 I think we should still use if, because computed `displayAll` can't reverse itself, other comments fixed.

lib/src/screens/dashboard/widgets/filter_widget.dart
```diff
@@ -1,155 +1,143 @@
...
+                                          //   child: Text(
+                                          //     item.caption,
+                                          //     style: TextStyle(
+                                          //         color: Theme.of(context).primaryTextTheme.title.color,
+                                          //         fontSize: 18,
+                                          //         fontFamily: 'Lato',
+                                          //         fontWeight: FontWeight.w500,
+                                          //         decoration: TextDecoration.none),
+                                          //   ),
+                                          // ),
```
OmarHatem28, line 122: if this is commented and not used anymore, better remove it

lib/src/widgets/standard_list_row.dart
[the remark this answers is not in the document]
OmarHatem28: ok, please check the usage for each one and see if they can be merged into one, if not, then they most probably have a different purpose, thus can be renamed to something more specific to their usage

lib/src/screens/wallet_keys/wallet_keys_page.dart
```diff
@@ -57,10 +58,7 @@ class WalletKeysPage extends BasePage {
                       height: 1,
                       padding: EdgeInsets.only(left: 24),
                       color: Theme.of(context).accentTextTheme!.headline6!.backgroundColor!,
-                      child: Container(
-                        height: 1,
-                        color: Theme.of(context).dividerColor,
-                      ),
+                      child: SectionDivider(),
```
OmarHatem28, line 61: **Minor:**

since this widget has a const constructor, it would be a bit better for the performance to use the same saved instance by adding `const` before the constructor anywhere its used in

Serhii-Borodenko:
@OmarHatem28 Both StandardListRow are used for almost the same purposes, in order to marge them, we need to rework several screens, I think it would be right to put this in a separate PR

OmarHatem28:
ok, thanks for looking into it, I asked Justin to verify the latest changes and will merge after that

OmarHatem28 (approved):

PR #643, opened 2022-11-21
Cw 223 settings reorganization

godilite:


OmarHatem28 (commented):
Everything looks good, but I would suggest that we make a separate view model for each of these new settings pages, they will be small at first but looking at the Figma design they will be populated step by step, so better make it scalable and easily maintained from now, what do you think?

lib/view_model/settings/settings_view_model.dart
```diff
@@ -264,55 +69,31 @@ abstract class SettingsViewModelBase with Store {
...
-  void setShouldSaveRecipientAddress(bool value) =>
-      _settingsStore.shouldSaveRecipientAddress = value;
-
-  @action
-  void setAllowBiometricalAuthentication(bool value) =>
-      _settingsStore.allowBiometricalAuthentication = value;
-
   @action
   void toggleTransactionsDisplay() =>
       actionlistDisplayMode.contains(ActionListDisplayMode.transactions)
```
OmarHatem28, line 99: I can't see any use for those, please double-check on them and if so, please remove them and delete this class since its not used anymore

lib/view_model/settings/settings_view_model.dart
```diff
@@ -264,55 +69,31 @@ abstract class SettingsViewModelBase with Store {
...
-    final priority = _settingsStore.priority[_walletType];
+    final priority = _settingsStore.priority[walletType];
 
     if (priority == null) {
-      throw Exception('Unexpected type ${_walletType.toString()}');
+      throw Exception('Unexpected type ${walletType.toString()}');
     }
 
     return priority;
   }
```
OmarHatem28, line 79: only the Exchange view model uses this getter and it has a dependency for `settingsStore` too, so we can move it to the ExchangeViewModel and remove its dependency on this class

OmarHatem28 (changes requested):
Also, please change the name of the `_settingsViewModel` in each page to its corresponding class name as it's confusing. (i.e. `_displaySettingsViewModel`, etc...

OmarHatem28 (approved):

res/values/strings_en.arb
```diff
@@ -652,5 +652,12 @@
   "use_suggested": "Use Suggested",
   "do_not_share_warning_text" : "Do not share these with anyone else, including support.\n\nThey will steal your money!",
   "help": "help",
+  "connection_sync": "Connection and sync",
+  "security_and_backup": "Security and backup",
+  "create_backup": "Create backup",
```
OmarHatem28, line 657: @godilite these does not have a translation in other languages, please add them

PR #656, opened 2022-11-29
Rename generated test app and apk

OmarHatem28:


.github/workflows/pr_test_build.yml
```diff
@@ -111,6 +114,16 @@ jobs:
...
           curl ${{ secrets.SLACK_WEB_HOOK }} -H "Content-Type: application/json" -d '{"apk_link": "'"$var"'","ticket": "'"$GITHUB_HEAD_REF"'"}'
 
+      - name: Rename apk file
+        run: |
+          cd /opt/android/cake_wallet/build/app/outputs/apk/release
+          mkdir test-apk
+          mv app-release.apk test-apk/$GITHUB_HEAD_REF.apk
+
+      - name: Upload Artifact
+        uses: kittaakos/upload-artifact-as-is@v0
```
SamsungGalaxyPlayer, line 127: Why is this used instead of `actions/upload-artifact`?

OmarHatem28: I tried both, but this will upload the artifact with the given file name `actions/upload-artifact` had an issue with making the artifact name with a variable name and required more steps, but they are both the same, this one is even built based on `actions/upload-artifact`

PR #679, opened 2022-12-16
Fix Nullability issue in gift cards

OmarHatem28:
add missing default priority for btc and ltc


SamsungGalaxyPlayer:
@OmarHatem28 can you confirm that this means Anypay is returning null values for the Bitcoin and Litecoin fees? Or is something else going on here?

OmarHatem28:
it's not related to Anypay, it was an issue with our code where we don't send the transaction priority in btc and ltc

mkyq (approved):

PR #685, opened 2022-12-22
Cw 250 bitcoin errors where transactions do not appear

Serhii-Borodenko:


cw_core/lib/transaction_direction.dart
```diff
@@ -20,4 +20,14 @@ TransactionDirection parseTransactionDirectionFromNumber(String raw) {
...
+int parseTransactionDirectionFromType(TransactionDirection raw) {
+  switch (raw) {
+    case TransactionDirection.incoming:
+      return 0;
+    case TransactionDirection.outgoing:
+      return 1;
+    default:
+      throw Exception('Unexpected token: raw for TransactionDirection parseTransactionDirectionFromType');
+  }
+}
```
OmarHatem28, line 33: this can be accessed from the `index` attribute of the enum instead of defining a separate function, plus there won't be a case where it throws an exception as it will always have a value

cw_bitcoin/lib/electrum_transaction_history.dart
```diff
@@ -38,11 +39,26 @@ abstract class ElectrumTransactionHistoryBase
...
+          txsDirection.forEach((directionKey, directionValue) {
+            if (directionKey == key) {
+              value.direction = directionValue;
+            }
+          });
+          if (!txsDirection.containsKey(key)) {
+            value.direction = TransactionDirection.outgoing;
+          }
+          ;
+        });
```
OmarHatem28, line 59: no need for the nested `forEach` you can just set the direction of the transaction with the `txsDirection[key]` and if null set it with the outgoing

`

       transactions.forEach((key, value) {
            value.direction = txsDirection[key] ?? TransactionDirection.outgoing;
        });

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -617,7 +617,8 @@ abstract class ElectrumWalletBase extends WalletBase<ElectrumBalance,
       final transactions = await fetchTransactions();
       transactionHistory.addMany(transactions);
       walletAddresses.updateReceiveAddresses();
-      await transactionHistory.save();
+      final transactionDirection = transactions.map((key, value) => MapEntry(key, value.direction));
+      await transactionHistory.save(txsDirection:transactionDirection);
```
OmarHatem28, line 621: shouldn't this be handled by default from `addOne` and `addMany` functions? they take the whole transaction object / transactions list

OmarHatem28 (changes requested):

cw_bitcoin/lib/electrum_wallet.dart
Serhii-Borodenko: @OmarHatem28 the direction of transactions in an existing wallet has been hardcoded as incoming,this property is overwritten in json in `save()` with the correct value.Had to call _load() twice before updating transactions from the server and after, since transactions from the device are loaded with the wrong direction

PR #691, opened 2022-12-29
Fix haven currencies not getting serialized

OmarHatem28:


OmarHatem28:
@mkyq this issue was caused by refactoring the cryptocurrency deserialize methods, since `all` list did not include haven currencies, so I made another list for haven currencies and added them to the `_name` and `_raw` currency maps in order for them to be deserialized too, tho, I can't find any usage for these currencies in the current code, nor they were available in the `all` currencies list in previous versions

mkyq (approved):

PR #696, opened 2023-01-05
CW-252-Add-notice-when-connecting-to-.onion-node

Serhii-Borodenko:
Added an alert when connecting to an .onion node.


res/values/strings_de.arb
```diff
@@ -677,5 +677,6 @@
   "disabled": "Deaktiviert",
   "enabled": "Ermöglicht",
   "tor_only": "Nur Tor",
-  "unmatched_currencies": "Die Währung Ihres aktuellen Wallets stimmt nicht mit der des gescannten QR überein"
+  "unmatched_currencies": "Die Währung Ihres aktuellen Wallets stimmt nicht mit der des gescannten QR überein",
+  "orbot_running_alert": "Please make sure Orbot is running prior to connecting to this node."
```
OmarHatem28, line 681: Please correct this translation to german

OmarHatem28 (changes requested):
Please pull the latest main, and fix translation conflicts, also there are 2 newly added languages (thai and arabic), so please add the new text `orbot_running_alert` to them too

OmarHatem28 (approved):
but please, follow the same ordering for added strings in the localization files, so it can be easier to review later on

PR #697, opened 2023-01-05
Update build environment

OmarHatem28:
Github is deprecating Ubuntu 18.04, so I updated it to 20.04 and tested the build flow with and without cached data


SamsungGalaxyPlayer:
@mkyq please review when you can, thanks!

mkyq (approved):

PR #710, opened 2023-01-05
Fixed French translations

hbs:
Some translations were done using a poor quality automatic translation tool, leading to some fun mistakes :-)

Do not hesitate to contact me whenever there's a need to add French translation instead of me having to fix them after the releases.


OmarHatem28 (approved):
We wouldn't want to miss your PRs ^^ Thanks for your contribution.

PR #711, opened 2023-01-06
CW-269 replace webview plugin with inappwebview plugin

godilite:


lib/src/screens/buy/onramper_page.dart
```diff
@@ -70,17 +63,30 @@ class OnRamperPageBody extends StatefulWidget {
...
-        javascriptMode: JavascriptMode.unrestricted);
+    return InAppWebView(
+      initialOptions: InAppWebViewGroupOptions(
+        crossPlatform: InAppWebViewOptions(transparentBackground: true),
+      ),
+      initialUrlRequest: URLRequest(
+        url: Uri.tryParse(
+          widget.uri.toString(),
+        ),
+      ),
```
OmarHatem28, line 76: `widget.uri` is already a `URI` why convert to string and then parse to URI again

lib/src/screens/buy/onramper_page.dart
```diff
@@ -70,17 +63,30 @@ class OnRamperPageBody extends StatefulWidget {
...
+      androidOnPermissionRequest: (_, __, resources) async {
+        bool permissionNotGranted = await Permission.camera.status != PermissionStatus.granted;
+        if (permissionNotGranted) {
+          permissionNotGranted = await Permission.camera.request().isGranted;
```
OmarHatem28, line 80: if the user granted the permission so this will return `true` to the `permissionNOTgranted` variable, and then you will check on it and if it's true you will set the response action to `DENY` which should be `GRANT`

OmarHatem28 (changes requested):

OmarHatem28 (changes requested):
Please revert the changes in the android/ios files as they are unrelated, if there is a reason for changing them feel free to discuss it.
The only related files are 4
`android/app/src/main/AndroidManifestBase.xml`
`lib/src/screens/buy/buy_webview_page.dart`
`lib/src/screens/buy/onramper_page.dart`
`pubspec_base.yaml`

OmarHatem28 (approved):

PR #719, opened 2023-01-09
Create strings_id.arb

Babypoo-byte:


maddymodd (changes requested):
the indonesian flag was already in the project. The flag you added looks too tall. So the flag that was already in the repo should be used.

Babypoo-byte:
> the indonesian flag was already in the project. The flag you added looks too tall. So the flag that was already in the repo should be used.

ok i will fix it

Babypoo-byte:
> Thanks for your commit. the indonesian flag was already in the project. The flag you added looks too tall. So the flag that was already in the repo should be used.

Is it already correct ?

maddymodd:
> > Thanks for your commit. the indonesian flag was already in the project. The flag you added looks too tall. So the flag that was already in the repo should be used.
> 
> Is it already correct ?

The original file has to be put back in the project. your most recent commit removed the flag without putting the original one back

Babypoo-byte:
> > > Thanks for your commit. the indonesian flag was already in the project. The flag you added looks too tall. So the flag that was already in the repo should be used.
> > 
> > 
> > Is it already correct ?
> 
> The original file has to be put back in the project. your most recent commit removed the flag without putting the original one back

flag has been returned, then what about this file conflict ?

lib/entities/fiat_currency.dart
```diff
@@ -9,7 +9,7 @@ class FiatCurrency extends EnumerableItem<String> with Serializable<String> {
   static List<FiatCurrency> get all => _all.values.toList();
 
   static List<FiatCurrency> get currenciesAvailableToBuyWith =>
-      [aud, brl, cad, chf, czk, eur, dkk, gbp, hkd, ils, jpy, krw, mxn, myr, nok, nzd, pln, sek, sgd, thb, usd, zar];
+      [aud, brl, cad, chf, czk, eur, dkk, gbp, hkd, ils, jpy, krw, mxn, myr, nok, nzd, pln, sek, sgd, thb, usd, zar, idr];
```
OmarHatem28, line 12: please revert this change, only specific currencies are available to buy with, if you wanna add Indonesian currency, please add it in a separate PR with a separate ticket

OmarHatem28 (changes requested):
you deleted the Indonesian flag, please revert that too

lib/entities/fiat_currency.dart
OmarHatem28: also please fix conflicts with main

PR #722, opened 2023-01-10
Fix Additional characters on scanned QR addresses

OmarHatem28:
Trim the content scanned via QR to remove extra spaces/new lines

Screen Shot 2023-01-10 at 7 58 33 PM


SamsungGalaxyPlayer (approved):
Tested ok on android

PR #725, opened 2023-01-11
Add build scripts for macOS. Add macos for cw_monero plugin. Add maco…

mkyq:
Added build scripts for macOS; Added macOS for cw_monero plugin. Added macOS proj to the application.
For now available two configurations for build: arm64 or x86_64, each platform required same host, so because of that cannot to build universal binary directly, because some dependency failing to build because require native asm (did not find way how to avoid it yet). Universal binary will try add later if it will be necessary.
Scripts should be run like:
`cd scripts/macos`
`./cakewallet.sh`
`./build_all.sh`
`cd ../..`
`cd macos && pod install && cd ..` // reinstall pods
`
then generate mobx and another generating files as in regular flow for mobile platforms.


mkyq:
Haven is not added there yet, as there some strange build fail for haven which require iconv for boost (not for locale, for somewhat ?? there actually should be used platform provided lib for that). But it fails for some specially binary, so technically libs should works.

mkyq:
Install needed tools before build process
`brew install autoconf automake libtool pkg-config`

Configure cw_monero and the application macos project after `build_all.sh` script:
`./gen_arm64.sh` if you are on Apple Silicone or `./gen_x86_64.sh` if you are on Intel
`./cakewallet.sh`

OmarHatem28:
merged with [737](https://github.com/cake-tech/cake_wallet/pull/737)

PR #732, opened 2023-01-16
CW-247 exchange confirmation screen address labels

Serhii-Borodenko:
Added a new table item to the confirmation screen


lib/exchange/sideshift/sideshift_exchange_provider.dart
```diff
@@ -150,6 +150,7 @@ class SideShiftExchangeProvider extends ExchangeProvider {
       refundAddress: settleAddress,
```
OmarHatem28, line 150: why are we using the `settleAddress` as the `refundAddress`? we already have the refund address inside the request model.
I believe this is the reason why we are not using `settleAddress` as the `refundAddress` in `findTradeById`

OmarHatem28, line 150: why are we using the `settleAddress` as the `refundAddress`? we already have the refund address inside the request model.
I believe this is the reason why we are not using `settleAddress` as the `refundAddress` in `findTradeById`

lib/exchange/trade.dart
```diff
@@ -88,6 +89,9 @@ class Trade extends HiveObject {
   @HiveField(12)
   String? walletId;
 
+  @HiveField(13)
+  String? payoutAddress;
+
   static Trade fromMap(Map<String, Object?> map) {
```
OmarHatem28, line 95: `fromMap` and `toMap` are outdated and not used, we can remove them, right?

OmarHatem28 (approved):

PR #744, opened 2023-01-25
Qr scheme fix

OmarHatem28:
Issue Number (if Applicable): Fixes # N/A

# Description

Compare the scheme of the URI with the full name of the currency instead of the abbreviated one

# Pull Request - Checklist  


SamsungGalaxyPlayer (approved):
Works in testing

Issue #761, opened 2023-02-03
Cannot build the app by following instructions in howto-build-android.md


holecekp:
I have tried to build Cake Wallet according to the instructions in https://github.com/cake-tech/cake_wallet/blob/main/howto-build-android.md, but it failed.

I have created a virtual machine, downloaded Ubuntu 20.04, installed it, and followed the steps in the document.

The following command for installing dependencies failed: `sudo apt-get install -y curl unzip automake build-essential file pkg-config git python libtool libtinfo5 cmake openjdk-8-jre-headless clang`. The `python` has not been found. I have used `python3` and `python-is-python3` instead. This probably worked and I was able to continue.

However, the last command `flutter build apk --release` ended with an error and I am not able to create an APK. The following error message appeared:

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':cw_haven'.
> NDK at /home/pavel/Android/Sdk/ndk/25.1.8937393 did not have a source.properties file

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 21s
Running Gradle task 'assembleRelease'...                           22.5s
Gradle task assembleRelease failed with exit code 1
```

It looks like a version problem. The NDK version 25.1.8937393 is set in https://github.com/cake-tech/cake_wallet/blob/main/android/app/build.gradle, which is missing because the script https://github.com/cake-tech/cake_wallet/blob/main/scripts/android/install_ndk.sh probably installs a different version instead - 17.2.4988734. 

I have installed the required NDK version 25.1.8937393 in Android Studio and I tried to build Cake Wallet again (`flutter build apk --release` ) but it failed with a different error:

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':cw_haven:externalNativeBuildRelease'.
> Build command failed.
  Error while executing process /home/pavel/Android/Sdk/cmake/3.10.2.4988404/bin/ninja with arguments {-C /opt/android/cake_wallet/cw_haven/android/.cxx/cmake/release/armeabi-v7a cw_haven}
  ninja: Entering directory `/opt/android/cake_wallet/cw_haven/android/.cxx/cmake/release/armeabi-v7a'
[...]

PR #762, opened 2023-02-04
Add Czech language

holecekp:
# Description

Czech language has been added.

# Pull Request - Checklist  

I have not able to test the translation yet because of https://github.com/cake-tech/cake_wallet/issues/761. I will test it once the APK is created by pr_test_build.yml.


SamsungGalaxyPlayer:
Auto-built APK (downloads a ZIP): https://github.com/cake-tech/cake_wallet/suites/10779386839/artifacts/541708716

holecekp:
I have installed the APK from the link into an older phone but the tests did not went well. The app starts into English and a warning is displayed after installation that can be translated as: "This app contains a virus. Uninstall it immediately". Do you have any idea what could cause it?
cake1

I have ignored it thinking that the old Huawei phone probably just does not like sideloaded apps. I do not use this phone for other things than testing anyway so I continued. The bigger problem is that I have selected to create a new wallet, then I was prompted to choose a PIN, but any 4 or 6 digit PIN is refused with the following error so I am not able to continue:
cake2
If I select to recover a wallet instead of creating a new one, I am again required to select a pin, which ends up with the same error.

Justxd22:
@holecekp yes this happens for me, every workflow GitHub builds produce same error, I'm guessing it has something to do with the wrong keys used in the build or it's a test build  (unusable), I tested multiple devices Android 13, 11, 10 same error, 

you can ignore the virus thing it's safe, any new builds is triggered as virus as it was not scanned before

OmarHatem28:
@holecekp yeah as Justxd22 said, it is triggered as a virus as it was not scanned before and installed from an unknown source

regarding the generated apk error, the workflow build is configured to add some keys from the GitHub secrets and any **External** PR does not have access to these secrets thus it builds the apk with empty secrets thus leading to those errors.

regarding the PR, looks good and tests well, please sync your PR with the latest updates in main and fix the conflicts

OmarHatem28 (approved):

SamsungGalaxyPlayer:
Thank you for your contribution @holecekp!

PR #769, opened 2023-02-08
Fix buy feature on macos

OmarHatem28:

# Description

# Pull Request - Checklist  


lib/entities/main_actions.dart
```diff
@@ -42,10 +46,15 @@ class MainActions {
 
       switch (walletType) {
         case WalletType.bitcoin:
-          Navigator.of(context).pushNamed(Routes.onramperPage);
```
godilite, line 45: is this omission for bitcoin a mistake or required?

OmarHatem28: that is not an omission, if its a bitcoin, it will go to the next case since there is no `break`
you can read about `fall-through` [here](https://dart.dev/guides/language/language-tour#switch-and-case)

godilite (approved):

PR #773, opened 2023-02-09
add address lookup for pinned tweet

Serhii-Borodenko:


lib/twitter/twitter_api.dart
```diff
@@ -32,6 +32,16 @@ class TwitterApi {
       throw Exception(responseJSON['errors'][0]['detail']);
     }
 
-    return TwitterUser.fromJson(responseJSON['data'] as Map<String, dynamic>);
+    final user = responseJSON['data'] as Map<String, dynamic>;
+
+    try {
+      if (responseJSON['includes'] != null) {
+        user['pinnedTweet'] = responseJSON['includes']['tweets'][0]['text'];
```
OmarHatem28, line 39: Better move this to the `TwitterUser` object and take the list of tweets (new object) and search in all of them via a map or forEach

lib/entities/parse_address_from_domain.dart
```diff
@@ -44,10 +44,15 @@ class AddressResolver {
       if (text.startsWith('@') && !text.substring(1).contains('@')) {
         final formattedName = text.substring(1);
         final twitterUser = await TwitterApi.lookupUserByName(userName: formattedName);
-        final address = extractAddressByType(
+        final addressFromBio = extractAddressByType(
             raw: twitterUser.description ?? '', type: CryptoCurrency.fromString(ticker));
-        if (address != null) {
-          return ParsedAddress.fetchTwitterAddress(address: address, name: text);
+        final addressFromPinnedTweet = extractAddressByType(
+            raw: twitterUser.pinnedTweet ?? '', type: CryptoCurrency.fromString(ticker));
```
OmarHatem28, line 50: We can move this after the `if (addressFromBio != null)` as it won't have to look through the pinned tweets if it found an address in the Bio

OmarHatem28 (changes requested):

lib/twitter/twitter_user.dart
```diff
@@ -1,16 +1,73 @@
...
+  Includes.fromJson(Map<String, dynamic> json) {
+    tweets = List.from(json['tweets'] as Iterable<dynamic>)
+        .map((e) => Tweets.fromJson(e as Map<String, dynamic>))
+        .toList();
+  }
+}
+
+class Tweets {
+  Tweets({
+    required this.editHistoryTweetIds,
```
OmarHatem28, line 58: - Rename to Tweet (classes are always singular)
- we can remove the `editHistory` variable since we have no use for it

lib/twitter/twitter_user.dart
```diff
@@ -1,16 +1,73 @@
...
+
+  late final Data data;
+  late final Includes? includes;
+
+  TwitterUser.fromJson(Map<String, dynamic> json) {
+    data = Data.fromJson(json['data'] as Map<String, dynamic>);
+    includes = json['includes'] != null
+        ? Includes.fromJson(json['includes'] as Map<String, dynamic>)
+        : null;
+  }
```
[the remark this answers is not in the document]
OmarHatem28: and we can directly extract user's data without the class
i.e `username: json['data']['username'],` etc...

OmarHatem28 (approved):

PR #774, opened 2023-02-09
[CW-310] Add constraints to images on macos

godilite:

# Description

# Pull Request - Checklist  

### DIFF
--- a/lib/src/screens/new_wallet/new_wallet_page.dart
+++ b/lib/src/screens/new_wallet/new_wallet_page.dart
@@ -1,6 +1,7 @@
 import 'package:cake_wallet/entities/generate_name.dart';
 import 'package:cake_wallet/routes.dart';
 import 'package:cake_wallet/themes/theme_base.dart';
+import 'package:cake_wallet/utils/responsive_layout_util.dart';
 import 'package:cake_wallet/utils/show_pop_up.dart';
 import 'package:mobx/mobx.dart';
 import 'package:flutter_mobx/flutter_mobx.dart';
@@ -28,9 +29,6 @@ class NewWalletPage extends BasePage {
 
   @override
   String get title => S.current.new_wallet;
-
-  @override
-  Widget trailing(BuildContext context) => SizedBox.shrink();
 
   @override
   Widget body(BuildContext context) => WalletNameForm(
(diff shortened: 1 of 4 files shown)


OmarHatem28 (changes requested):
do you think keeping the aspect ratio, and wrapping it with a `ConstrainedBox` and giving it a `maxWidth` would be better than giving a size depending on the screen size, also I believe this `ConstrainedBox` should be on the whole body because the rest of the view is way stretched, check screenshots in the ticket for clearer details

lib/src/screens/new_wallet/new_wallet_type_page.dart
```diff
@@ -13,23 +13,22 @@ class NewWalletTypePage extends BasePage {
...
-  final walletTypeLightImage =
-      Image.asset('assets/images/wallet_type_light.png');
+  final walletTypeLightImage = Image.asset('assets/images/wallet_type_light.png');
 
   @override
   String get title => S.current.wallet_list_restore_wallet;
 
+  @override
+  Widget trailing(BuildContext context) => SizedBox.shrink();
+
```
OmarHatem28, line 23: Remove this please, I already handled the case of null trailing in the `nav_bar.dart`

lib/src/screens/pin_code/pin_code_widget.dart
```diff
@@ -165,109 +160,112 @@ class PinCodeState<T extends PinCodeWidget> extends State<T> {
...
+            )
+          ],
+          Spacer(flex: 1),
+          Flexible(
             flex: 24,
-            child: Container(
+            child: Center(
+              child: Container(
+                height: MediaQuery.of(context).size.height * 0.6,
+                width: MediaQuery.of(context).size.height * 0.6,
```
OmarHatem28, line 173: Constrained box would be better here too, cause this will affect the mobile view slightly

lib/src/screens/new_wallet/new_wallet_page.dart
```diff
@@ -90,141 +87,138 @@ class _WalletNameFormState extends State<WalletNameForm> {
...
-                            });
-                          },
-                          icon: Container(
-                            padding: const EdgeInsets.all(8),
-                            decoration: BoxDecoration(
-                              borderRadius: BorderRadius.circular(6.0),
-                              color: Theme.of(context).hintColor,
+      child: Center(
+        child: ConstrainedBox(
+          constraints: BoxConstraints(maxWidth: 600),
```
OmarHatem28, line 94: we can add this `600` to the responsive util as a const with a descriptive name

OmarHatem28 (changes requested):

OmarHatem28:
and please fix the conflicts here

PR #796, opened 2023-02-20
Fix Popup width constraint and add focus orders

godilite:

# Description

# Pull Request - Checklist  

### DIFF
--- a/lib/src/widgets/address_text_field.dart
+++ b/lib/src/widgets/address_text_field.dart
@@ -58,7 +58,6 @@ class AddressTextField extends StatelessWidget {
         TextFormField(
           onFieldSubmitted: (_) => FocusScope.of(context).unfocus(),
           enabled: isActive,
-          autofocus: true,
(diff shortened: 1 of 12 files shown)


lib/src/widgets/address_text_field.dart
```diff
@@ -58,12 +58,13 @@ class AddressTextField extends StatelessWidget {
         TextFormField(
           onFieldSubmitted: (_) => FocusScope.of(context).unfocus(),
           enabled: isActive,
+          autofocus: true,
```
OmarHatem28, line 61: I don't think this will be good to add

OmarHatem28 (changes requested):

lib/src/widgets/address_text_field.dart
OmarHatem28: same in `base_text_form_field`

OmarHatem28 (approved):

PR #818, opened 2023-03-04
CW-243-Better-error-message-for-too-many-decimals

Serhii-Borodenko:
Update amoung validation


lib/core/amount_validator.dart
```diff
@@ -1,32 +1,68 @@
...
+
+  late final SymbolsAmountValidator symbolsAmountValidator;
+
+  late final DecimalAmountValidator decimalAmountValidator;
+
+  String? call(String? value) => symbolsAmountValidator(value) ?? decimalAmountValidator(value);
+}
+
+class SymbolsAmountValidator extends TextValidator {
+  SymbolsAmountValidator({required CryptoCurrency currency, required bool isAutovalidate})
```
OmarHatem28, line 20: we can remove the currency from here since we only check for the symbols, and the `_pattern` function can directly return the 1 RegEx that you are currently using

lib/core/amount_validator.dart
```diff
@@ -1,32 +1,68 @@
...
-      case WalletType.monero:
-        return '^([0-9]+([.\,][0-9]{0,12})?|[.\,][0-9]{1,12})\$';
-      case WalletType.bitcoin:
-        return '^([0-9]+([.\,][0-9]{0,8})?|[.\,][0-9]{1,8})\$';
+  static String _pattern(CryptoCurrency currency) {
+    switch (currency) {
+      case CryptoCurrency.xmr:
+        return '^([0-9]+([.\,][0-9]{1,12})?|[.\,][0-9]+)\$';
+      case CryptoCurrency.btc:
+        return '^([0-9]+([.\,][0-9]{1,8})?|[.\,][0-9]+)\$';
```
OmarHatem28, line 54: please validate those RegEx and we can instead of adding all the currencies, just set the `default` case to the average amount validation, which I believe would be the same as the `btc` RegEx, but feel free to verify that too

OmarHatem28 (changes requested):

lib/core/amount_validator.dart
```diff
@@ -1,32 +1,69 @@
...
-      case WalletType.bitcoin:
-        return '^([0-9]+([.\,][0-9]{0,8})?|[.\,][0-9]{1,8})\$';
+  static String _pattern(CryptoCurrency currency) {
+    switch (currency) {
+      case CryptoCurrency.xmr:
+        return '^([0-9]+([.\,][0-9]{1,12})?|[.\,][0-9]{1,12})\$';
+      case CryptoCurrency.btc:
+      case CryptoCurrency.ltc:
+      case CryptoCurrency.xhv:
+        return '^([0-9]+([.\,][0-9]{1,8})?|[.\,][0-9]{1,8})\$';
```
OmarHatem28, line 55: use this as the default case instead and remove `ltc` and `xhv`, since we want the amount for all other currencies to match this pattern

lib/core/amount_validator.dart
OmarHatem28: still, the `curreny` parameter can be removed from this class and the `_pattern` can simply return that single RegEx we have
`'^([0-9]+([.\,][0-9]+)?|[.\,][0-9]+)\$'`

OmarHatem28 (changes requested):

OmarHatem28:
also please fix the conflicts with main

PR #823, opened 2023-03-08
Cw 314 trocador receive screen update

godilite:

# Description

# Pull Request - Checklist  

- [ ] Formate code


lib/view_model/anon_invoice_page_view_model.dart
```diff
@@ -0,0 +1,192 @@
...
+  @observable
+  double? maximum;
+
+  @action
+  void selectCurrency(Currency currency) {
+    selectedCurrency = currency;
+    maximum = minimum = null;
+    if (currency is CryptoCurrency) {
+      cryptoCurrency = currency;
+    }
```
OmarHatem28, line 89: else, the `cryptoCurrency` should be reset to the wallet currency

lib/view_model/anon_invoice_page_view_model.dart
```diff
@@ -0,0 +1,192 @@
...
+    maximum = limit.max != null ? limit.max! / 4 : null;
+    if (selectedCurrency is FiatCurrency) {
+      final fiatPrice = _fiatConversationStore.prices[cryptoCurrency];
+      maximum = maximum != null
+          ? double.tryParse(calculateFiatAmount(price: fiatPrice, cryptoAmount: maximum.toString()))
+          : null;
+      minimum = minimum != null
+          ? double.tryParse(calculateFiatAmount(price: fiatPrice, cryptoAmount: minimum.toString()))
+          : null;
+    }
```
OmarHatem28, line 165: this check should be first, before making the API call, cause it won't be necessary to make the API call if the condition is true

OmarHatem28: also, this will get the price for the fiat currency selected in the settings, not the one selected here

lib/view_model/anon_invoice_page_view_model.dart
```diff
@@ -0,0 +1,192 @@
...
+
+    final result = await anonPayApi.generateDonationLink(AnonPayRequest(
+      cryptoCurrency: cryptoCurrency,
+      address: address,
+      description: description,
+      email: receipientEmail,
+      name: receipientName,
+    ));
+
+    await sharedPreferences.setString(PreferencesKey.trocadorDonationLink, result.clearnetUrl);
```
OmarHatem28, line 148: I would suggest saving the whole result thus you don't have to call the API the next time the user goes here, causing the app to show the form for a couple of seconds then navigate to the QR page

OmarHatem28: then you can check for the saved value before even navigating to the donation form

res/values/strings_en.arb
```diff
@@ -685,6 +685,16 @@
...
+  "invoice_details": "Invoice details",
+  "donation_link_details": "Donation link details",
+  "anonpay_description": "Generate ${type}. The recipient can ${method} with any supported cryptocurrency, and you will receive funds in this wallet.",
+  "create_invoice": "Create invoice",
+  "create_donation_link": "Create donation link",
+  "optional_email_hint": "Optional payee notification email",
+  "optional_description": "Optional description",
+  "optional_name": "Optional recipient name",
+  "clearnet_link": "Clearnet link",
+  "onion_link": "Onion link",
```
OmarHatem28, line 697: missing other localizations

OmarHatem28 (changes requested):

lib/anonpay/anonpay_status_response.dart
```diff
@@ -0,0 +1,17 @@
+class AnonpayStatusResponse {
+  final String status;
+  final double? fiatAmount;
+  final String? fiatEquiv;
+  final double? amountTo;
```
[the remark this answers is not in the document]
OmarHatem28: as far as I can see it's the only way to know the amount, and it should return the amount for both crypto and fiat.
as we discussed shouldn't this be the amount of fiat if `fiatEquiv` is not null, else it would be the crypto amount

lib/view_model/anonpay_details_view_model.dart
```diff
@@ -0,0 +1,79 @@
...
+    _updateInvoiceDetail();
+    timer = Timer.periodic(Duration(seconds: 20), (_) async => _updateInvoiceDetail());
+  }
+
+  final AnonPayApi anonPayApi;
+  final SettingsStore settingsStore;
+  final AnonpayInvoiceInfo invoiceDetail;
+
+  @observable
+  ObservableList<StandartListItem> items;
```
OmarHatem28, line 40: **Minor:**
can be final

lib/store/dashboard/transaction_filter_store.dart
OmarHatem28, line 1: please reformat

OmarHatem28 (changes requested):

lib/src/screens/dashboard/widgets/present_fee_picker.dart
OmarHatem28, line 1: please rename the file as well

lib/src/screens/receive/anonpay_invoice_page.dart
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  final _nameController = TextEditingController();
+  final _emailController = TextEditingController();
+  final _descriptionController = TextEditingController();
+  final _amountController = TextEditingController();
+  FocusNode _nameFocusNode;
+  FocusNode _emailFocusNode;
+  FocusNode _descriptionFocusNode;
+  FocusNode _amountFocusNode;
```
OmarHatem28, line 45: final and can be moved to the anonpay_input_form directly

lib/src/screens/receive/receive_page.dart
```diff
@@ -220,11 +221,16 @@ class ReceivePage extends BasePage {
...
+            child: Observer(
+              builder: (context) {
+                return QRWidget(
+                    addressListViewModel: addressListViewModel,
+                    isAmountFieldShow: true,
+                    urlString: addressListViewModel.uri.toString(),
+                    amountTextFieldFocusNode: _cryptoAmountFocus,
+                    isLight: currentTheme.type == ThemeType.light);
+              }
+            ),
```
OmarHatem28, line 233: since we will revert changes in QR widget, no need for the Observer parent, and no need for urlString since it will be removed

lib/src/screens/receive/widgets/share_link_item.dart
OmarHatem28, line 1: change file name too

OmarHatem28 (changes requested):

OmarHatem28 (approved):

PR #827, opened 2023-03-10
CW-277-Allow-editing-nodes

Serhii-Borodenko:
Allowed to edit nodes


lib/di.dart
```diff
@@ -515,10 +515,15 @@ Future setup(
...
         type ?? getIt.get<AppStore>().wallet!.type,
         getIt.get<SettingsStore>(),
+        getIt.get<NodeListViewModel>()
     ));
 
-  getIt.registerFactory(
-      () => NodeCreateOrEditPage(getIt.get<NodeCreateOrEditViewModel>()));
+  getIt.registerFactoryParam<NodeCreateOrEditPage, List<dynamic>?, void>((List<dynamic>? args, _) {
+    final node = args?.first as Node?;
+    final isSelected = args?[1] as bool?;
```
OmarHatem28, line 523: better to use a Map instead of a List for args, cuz for any reason, if the passed args were not of the expected length, this will throw an index out of range

OmarHatem28 (approved):
clean and neat.
nice usage for the cascade operation

PR #854, opened 2023-03-30
CW-338: Currency picker UI when keyboard is showing

saltrafael:
# Description

Fix clicking to search a currency, and the keyboard layout being on top of the popup, making it harder to see fully

# Pull Request - Checklist  

- [X] Formate code


lib/src/widgets/picker.dart
```diff
@@ -200,34 +300,48 @@ class _PickerState<Item> extends State<Picker<Item>> {
...
     );
   }
 
-  Widget buildItem(int index) {
-    /// don't show selected item in the list view
-    if (widget.items[widget.selectedAtIndex] == items[index] && !widget.isGridView) {
-      return const SizedBox();
-    }
+  Widget buildItem(int index, [bool? buildSelected]) {
+    final isSelected = buildSelected == true;
```
OmarHatem28, line 317: I would say we keep them separated instead of having boolean comparisons everywhere, so it's more readable, besides, it's more scalable when separated, as whenever you want to change the selected item UI only, you won't have to go through all these boolean comparisons

OmarHatem28 (changes requested):
Overall great work ^^ just some comments for code standards and minor enhancements

lib/src/widgets/picker.dart
```diff
@@ -255,37 +401,92 @@ class _PickerState<Item> extends State<Picker<Item>> {
...
+            width: 20.0,
+          )
+        : null;
+
+    final image = images.isNotEmpty ? images[index] : icon;
+
+    return GestureDetector(
+      onTap: () {
+        Navigator.of(context).pop();
+        onItemSelected(item!);
```
OmarHatem28, line 421: **Minore:**
you can remove this, since it's already selected

OmarHatem28 (changes requested):

OmarHatem28 (approved):

PR #863, opened 2023-04-06
CW-59-New-update-highlight-popup

Serhii-Borodenko:
- [ ] Formate code


lib/utils/version_comparator.dart
```diff
@@ -0,0 +1,13 @@
...
+  static bool isVersion1Greater({required String v1, required String v2}) {
+    int v1Number = getExtendedVersionNumber(v1);
+    int v2Number = getExtendedVersionNumber(v2);
+    return v1Number > v2Number;
+  }
+
+  static int getExtendedVersionNumber(String version) {
+    List<String> stringVersionCells = version.split('.');
+    List<int> intVersionCells = stringVersionCells.map((i) => int.parse(i)).toList();
+    return intVersionCells[0] * 100000 + intVersionCells[1] * 1000 + intVersionCells[2];
```
OmarHatem28, line 11: you can instead do this
```dart
return int.parse(version.split(".").join());
```

lib/src/screens/release_notes/release_notes_screen.dart
```diff
@@ -0,0 +1,152 @@
...
+                    ),
+                    Expanded(
+                      flex: 20,
+                      child: Container(
+                        child: ListView.builder(
+                          padding: const EdgeInsets.all(0),
+                          itemCount: notesWidgetList.length,
+                          itemBuilder: (BuildContext context, int index) {
+                            return notesWidgetList[index];
+                          },
```
OmarHatem28, line 121: for future reference, it's better to use `FutureBuilder` when the UI is waiting for an async function, so that you can handle loading, and error states as well

OmarHatem28 (changes requested):
also please update the PR with the latest main

OmarHatem28 (approved):

PR #868, opened 2023-04-09
Cw 262 better handle user exchange amount below minimum or maximum trade size

Blazebrain:

# Description
Better handle user exchange amount below minimum or maximum trade size.

- [x] Using the minimum and maximum limits for each exchange, when a user types in a value that is below the minimum input or above the maximum limit, an error text  comes up, informing the user that it is above or below the limit. 

- [x] The "Convert to" value and "Convert from" value is left blank when this occurs, instead of it populating with the above/below limit values. 

- [x] The "Convert to" value is left blank when it's a variable exchange, while the "Convert from" value is left blank if it's a fixed exchange

- [x] The error text comes under the "Convert from" text field when it's a variable exchange while it comes under the "Convert to" text field if it's a fixed exchange

# Pull Request - Checklist  

- [x] Formate code


OmarHatem28 (changes requested):
Please revert all the formatting changes, and only commit the actual changes

lib/core/amount_validator.dart
```diff
@@ -3,17 +3,43 @@ import 'package:cake_wallet/generated/i18n.dart';
...
   }
 
   late final SymbolsAmountValidator symbolsAmountValidator;
 
   late final DecimalAmountValidator decimalAmountValidator;
 
-  String? call(String? value) => symbolsAmountValidator(value) ?? decimalAmountValidator(value);
+   late final AmountMinValidator amountMinValidator;
+
+   late final AmountMaxValidator amountMaxValidator;
```
OmarHatem28, line 29: indentation!

lib/src/screens/exchange/exchange_page.dart
```diff
@@ -456,9 +456,13 @@ class ExchangePage extends BasePage {
...
-          exchangeViewModel.changeDepositAmount(
-              amount: depositAmountController.text);
-          exchangeViewModel.isReceiveAmountEntered = false;
+      if (exchangeViewModel.checkIfInputMeetsMinOrMaxCondition(
+                   depositAmountController.text) ==
+               true){
+  exchangeViewModel.changeDepositAmount(
+      amount: depositAmountController.text);
+  exchangeViewModel.isReceiveAmountEntered = false;
+}
```
OmarHatem28, line 465: - format
- duplicate boolean check (i.e bool == true)

lib/src/screens/exchange/exchange_page.dart
```diff
@@ -469,9 +473,13 @@ class ExchangePage extends BasePage {
...
-          exchangeViewModel.changeReceiveAmount(
-              amount: receiveAmountController.text);
-          exchangeViewModel.isReceiveAmountEntered = true;
+         if (exchangeViewModel.checkIfInputMeetsMinOrMaxCondition(
+                   receiveAmountController.text) ==
+               true) {
+  exchangeViewModel.changeReceiveAmount(
+      amount: receiveAmountController.text);
+  exchangeViewModel.isReceiveAmountEntered = true;
+}
```
OmarHatem28, line 482: - format
- duplicate boolean check (i.e bool == true)

lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -305,7 +308,19 @@ abstract class ExchangeViewModelBase with Store {
...
+
+     if (_enteredAmount < minLimit) {
+       depositAmount = '';
+       return;
+     }
+
+     if (_enteredAmount > maxLimit && isFixedRateMode) {
+       depositAmount = '';
+       return;
+     }
```
OmarHatem28, line 323: - indentation
- if the `limits.max` is null, then it should be uncapped, you are resetting it to zero so the condition will always be true.
- also why are you checking on whether its a fixed rate or not with the maxlimit?

lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -320,6 +335,23 @@ abstract class ExchangeViewModelBase with Store {
...
+     if (_enteredAmount < minLimit) {
+       return false;
+     }
+
+     if (_enteredAmount > maxLimit && isFixedRateMode) {
+       return false;
+     }
+
+     return true;
+   }
```
OmarHatem28, line 352: - indentation
- if the limits.max is null, then it should be uncapped, you are resetting it to zero so the condition will always be true.
- also why are you checking on whether its a fixed rate or not with the maxlimit?

lib/core/amount_validator.dart
```diff
@@ -49,6 +75,75 @@ class DecimalAmountValidator extends TextValidator {
...
+     return valueInDouble > minInDouble;
+   }
+
+   double? parseToDouble(String value) {
+     final data = double.tryParse(value.replaceAll(',', '.'));
+     return data;
+   }
+ }
+
+ class AmountMaxValidator extends Validator<String> {
```
OmarHatem28, line 113: indentation

OmarHatem28 (changes requested):
will continue the review after the indentations and formatting is resolved, please double-check the rest of the added code for invalid indentation, extra empty lines (between different parts of code only 1 empty line is enough, translation files (`fr`)), and the comments I added

lib/core/amount_validator.dart
Blazebrain: Fixed

cw_monero/example/pubspec.lock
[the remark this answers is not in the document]
Blazebrain: Fixed it

lib/src/screens/exchange/exchange_page.dart
Blazebrain: Remove the duplicated boolean and fixed the formatting

lib/src/screens/exchange/exchange_page.dart
Blazebrain: Removed the double boolean and fixed the formatting

lib/view_model/exchange/exchange_view_model.dart
Blazebrain: Fixed the indentation, made it uncapped when the `limits.max` is null and removed the check

lib/view_model/exchange/exchange_view_model.dart
Blazebrain: Fixed the indentation, made it uncapped when the `limits.max` is null and removed the check

lib/core/amount_validator.dart
Blazebrain: Fixed the indentation issues

OmarHatem28 (approved):

PR #876, opened 2023-04-12
CW-273-Don't-add-node-under-Advanced-Privacy-Settings-if-it-already-exists

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [ ] Formate code


lib/view_model/node_list/node_create_or_edit_view_model.dart
```diff
@@ -109,11 +114,15 @@ abstract class NodeCreateOrEditViewModelBase with Store {
       state = IsExecutingState();
       if (editingNode != null) {
         await _nodeSource.put(editingNode.key, node);
+      } else if (existingNode != null) {
+        final node = existingNode;
```
OmarHatem28, line 118: instead of calling `existingNode` getter twice and looping through the list of nodes twice, we can just have a function that returns a boolean, whether this node exists or not, if it does then just `setAsCurrent`
```dart
else if (isExistingNode(node)) {
     setAsCurrent(node);
}
```
and you can remove the getter above `node` since there will be no use for it and we already have it as the final variable node in this function (line 106)

OmarHatem28 (approved):

PR #880, opened 2023-04-14
CW-344-Exchanging-TO-the-current-wallet-currency-should-allow-changing-the-address

Blazebrain:
Exchanging TO the current wallet currency should allow changing the address

Currently, if you are in the exchange screen, and you are converting to the currency of your current wallet, the user should be able to change the ConverTo address (by default it’s the current wallet address)

In Monero Wallet, going to the address book (from the icon on the field) and selecting an address, will just copy the address, instead, it should choose this address for Convert To address

# Pull Request - Checklist  

- [x] Formate code


lib/src/screens/exchange/widgets/exchange_card.dart
```diff
@@ -406,7 +406,7 @@ class ExchangeCardState extends State<ExchangeCard> {
                       order: NumericFocusOrder(3),
                       child: BaseTextFormField(
                           controller: addressController,
-                          readOnly: true,
+                          // readOnly: true,
```
OmarHatem28, line 409: remove, not comment

cw_monero/example/test/widget_test.dart
```diff
@@ -8,7 +8,7 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 
-import 'package:cw_monero_example/main.dart';
+import '../lib/main.dart';
```
OmarHatem28, line 11: why is this change?

lib/src/screens/contact/contact_list_page.dart
```diff
@@ -120,6 +120,7 @@ class ContactListPage extends BasePage {
         if (isCopied) {
           await Clipboard.setData(ClipboardData(text: contact.address));
           await showBar<void>(context, S.of(context).copied_to_clipboard);
+          Navigator.of(context).pop(contact);
```
OmarHatem28, line 123: this will pop the screen ok, but we need it to behave as the normal flow (i.e take the address from the selected contact and use it in the address field

OmarHatem28 (changes requested):

lib/src/screens/exchange/widgets/exchange_card.dart
Blazebrain: Fixed this

cw_monero/example/test/widget_test.dart
Blazebrain: Also, fixed

lib/src/screens/contact/contact_list_page.dart
```diff
@@ -119,7 +119,9 @@ class ContactListPage extends BasePage {
 
         if (isCopied) {
           await Clipboard.setData(ClipboardData(text: contact.address));
-          await showBar<void>(context, S.of(context).copied_to_clipboard);
+          Navigator.of(context).pop(contact);
+          // await showBar<void>(context, S.of(context).copied_to_clipboard);
+
```
OmarHatem28, line 124: please revert those, and make the `contactListViewModel.isEditable` to false this will already handle the case without further modifications

OmarHatem28 (changes requested):

OmarHatem28 (approved):

PR #886, opened 2023-04-17
CW-278 Enhance PIN timeout feature code

godilite:

# Description
Added extension to BuildContext to handle `navigateToAuthenticatedRoute` which should be used to navigate to routes that require authentication

Check need authentication logic and if route should always require authentication.

# Pull Request - Checklist  

- [X] Formate code


OmarHatem28:
Please update with the latest `main` to fix workflow issue

lib/core/extension.dart
```diff
@@ -0,0 +1,43 @@
...
+  Routes.showKeys,
+  Routes.backup,
+  Routes.setupPin,
+];
+
+final _authService = getIt.get<AuthService>();
+
+extension AuthenticateRoute on BuildContext {
+  void navigateToAuthenticatedRoute(
+      {Function(bool)? onAuthSuccess, String route = '', Object? arguments}) {
```
OmarHatem28, line 17: **Minor:**
`route` can be nullable and add an assertion instead to make sure either `onAuthSuccess` or `route`  is passed

lib/core/extension.dart
```diff
@@ -0,0 +1,43 @@
...
+
+const List<String> _alwaysAuthenticateRoutes = [
+  Routes.showKeys,
+  Routes.backup,
+  Routes.setupPin,
+];
+
+final _authService = getIt.get<AuthService>();
+
+extension AuthenticateRoute on BuildContext {
```
OmarHatem28, line 15: I would say we move this as a function in `AuthService` instead of an extension on Context, as it's not that much related to context, but more to the AuthService.
and it's not good to have global variables hanging around the code, we can move `_alwaysAuthenticateRoutes` to the AuthService too

lib/src/screens/settings/security_backup_page.dart
```diff
@@ -27,44 +27,32 @@ class SecurityBackupPage extends BasePage {
       child: Column(mainAxisSize: MainAxisSize.min, children: [
         SettingsCellWithArrow(
           title: S.current.show_keys,
-          handler: (_) => Navigator.of(context).pushNamed(Routes.auth,
-              arguments: (bool isAuthenticatedSuccessfully, AuthPageState auth) {
-            if (isAuthenticatedSuccessfully) {
-              auth.close(route: Routes.showKeys);
-            }
-          }),
+          handler: (_) => context.navigateToAuthenticatedRoute(route: Routes.showKeys),
```
OmarHatem28, line 30: and then here you can inject the AuthService dependency and call it here
`_authService.authenticateAction(route: Routes.showKeys),

lib/core/extension.dart
OmarHatem28: since it's not only navigation, we can use a more generic name
like `authenticateAction` for example

OmarHatem28 (changes requested):
also please update the branch with the latest main

OmarHatem28 (approved):
LGTM.
please look for other places with actions that need auth and apply the same function
ex. Changing/Deleting wallet and others

OmarHatem28 (approved):
Much better now, thanks

PR #900, opened 2023-04-25
CW-254-proper-bitcoin-address-validation-in-exchange-screen

saltrafael:

# Description

Tested with all currently in use (not obsolete) BTC address types from this guide: https://unchained.com/blog/bitcoin-address-types-compared/

# Pull Request - Checklist  

- [X] Format code


lib/core/address_validator.dart
```diff
@@ -1,4 +1,4 @@
-import 'package:flutter/foundation.dart';
+import 'package:cw_bitcoin/address_validate.dart' as bitcoin;
```
OmarHatem28, line 1: we can't use other packages in the main lib (`cw_bitcoin` and `cw_haven` etc...) as it will give a compilation error when building Monero.com as it does Not have bitcoin, haven, nor litecoin.
so please move the `cw_bitcoin/lib/address_validate.dart` to the project directory `lib` as it's not inclusive for BTC wallets only, its for BTC address validation in general, from any wallet
Also renaming it the file would be better to indicate that it validates bitcoin addresses

lib/core/validator.dart
```diff
@@ -1,24 +1,24 @@
-import 'package:flutter/foundation.dart';
-
 abstract class Validator<T> {
-  Validator({required this.errorMessage});
+  Validator({required this.errorMessage, this.doValidate});
 
   final String errorMessage;
+  final bool Function(T)? doValidate;
```
OmarHatem28, line 5: **Minor:**
naming can be better

OmarHatem28 (changes requested):

OmarHatem28 (changes requested):

lib/core/validator.dart
```diff
@@ -35,8 +37,13 @@ class TextValidator extends Validator<String> {
     return value.length > (minLength ?? 0) &&
         (length?.contains(value.length) ?? true) &&
         ((maxLength ?? 0) > 0 ? (value.length <= maxLength!) : true) &&
-        (pattern != null ? match(value) : true);
+        (pattern != null
+            ? (useAdditionalValidation != null &&
+                    useAdditionalValidation!(value)) ||
+                match(value)
+            : true);
```
OmarHatem28, line 44: what about the case were `useAdditionalValidation` is not null but returned false?
with the current implementation, it will proceed to check for `match(value)` and if it returned true then it will pass, although it should fail

OmarHatem28: also it would be better and more readable to separate these into variables with meaningful names and then use those variables in the return statement

OmarHatem28 (changes requested):

lib/core/validator.dart
saltrafael: I refactored a bit for better readability in the return of `isValid`, let me know if it looks better, in the `useAdditionalValidation` the idea is just to add on to the RegEx since `3...` addresses are not valid under the `bitcoin_flutter`'s validation function, so if it returns false it fallbacks to the RegEx making sure those addresses can return true

OmarHatem28 (approved):

PR #906, opened 2023-04-27
Cw 354 app issues with huge font size

saltrafael:

# Description

Following this video guide https://youtube.com/watch?v=neAn35cY8y0, in order for `SingleChildScrollView` to work with `Expanded` and `Flexible` widgets it was also needed to wrap `Column` with `IntrinsicHeight` and add a `LayoutBuilder` on top

There may be more non-scrollable pages but with this the app can be entered and used as normal and the issue closed

# Pull Request - Checklist  

- [X] Format code

### DIFF
--- a/lib/src/screens/seed/pre_seed_page.dart
+++ b/lib/src/screens/seed/pre_seed_page.dart
@@ -36,45 +36,57 @@ class PreSeedPage extends BasePage {
         child: Container(
           alignment: Alignment.center,
           padding: EdgeInsets.all(24),
-          child: ConstrainedBox(
-              constraints: BoxConstraints(maxWidth: ResponsiveLayoutUtil.kDesktopMaxWidthConstraint),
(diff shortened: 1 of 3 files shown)


OmarHatem28 (changes requested):
if you go to the `IntrinsicHeight` widget declaration, you can find that it's a very expensive widget in terms of performance.
Let's better refactor the widget tree for these screens.
What needs to be taken into consideration is the following:

1. Remove unnecessary widgets
2. The UI should be constrained in some way so that Desktop UI won't expand unconditionally
3. The UI should be scrollable when needed

OmarHatem28:
on MacOS the image expands more than needed, and is making the screen scrollable
Screen Shot 2023-05-05 at 7 21 34 PM

here is the image from the current implementation for comparison
Screen Shot 2023-05-05 at 7 21 56 PM

PR #910, opened 2023-05-01
Cw 363 back button on receive screen in bright theme is white

saltrafael:

# Description

# Pull Request - Checklist  

- [X] Format code

### DIFF
--- a/lib/src/screens/base_page.dart
+++ b/lib/src/screens/base_page.dart
@@ -26,6 +26,8 @@ abstract class BasePage extends StatelessWidget {
 
   Color? get titleColor => null;
 
+  bool get lightTitleColor => false;
+
   bool get resizeToAvoidBottomInset => true;
 
   bool get extendBodyBehindAppBar => false;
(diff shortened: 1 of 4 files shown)


lib/src/screens/base_page.dart
```diff
@@ -26,6 +26,8 @@ abstract class BasePage extends StatelessWidget {
 
   Color? get titleColor => null;
 
+  bool get lightTitleColor => false;
```
OmarHatem28, line 29: we don't need a new variable for this, it's just a misused theme style

lib/src/screens/dashboard/widgets/address_page.dart
```diff
@@ -60,16 +60,19 @@ class AddressPage extends BasePage {
   bool effectsInstalled = false;
 
   @override
-  Color get titleColor => Colors.white;
+  Color? get titleColor =>
+      currentTheme.type == ThemeType.bright ? Colors.white : null;
```
OmarHatem28, line 64: can be removed since we are overriding the leading widget entirely and `titleColor` is only used there

OmarHatem28, line 64: can be removed since we are overriding the leading widget entirely and `titleColor` is only used there

lib/src/screens/receive/receive_page.dart
```diff
@@ -53,7 +53,8 @@ class ReceivePage extends BasePage {
   final FocusNode _cryptoAmountFocus;
 
   @override
-  Color get titleColor => Colors.white;
+  Color? get titleColor =>
+      currentTheme.type == ThemeType.bright ? Colors.white : null;
```
OmarHatem28, line 57: can be removed too, having it with null will use the correct color for each theme by default

OmarHatem28 (approved):
Minor comments, can be addressed in another `in progress` PR

lib/src/screens/receive/receive_page.dart
saltrafael: Removing the white color from `receive_page` makes it look like this on the `Bright` theme:
image
Probably because the default color is meant for light background pages, like here:
image
with only `Bright` having 2 different backgrounds depending on the page, white or gradient, so this theme in specific still needs a way to have 2 different colors

OmarHatem28 (approved):

PR #919, opened 2023-05-08
CW-376-picker-ui-issue

saltrafael:

# Description

# Pull Request - Checklist  

- [X] Format code

### DIFF
--- a/lib/src/widgets/modal_widget.dart
+++ b/lib/src/widgets/modal_widget.dart
@@ -1,0 +1,64 @@
+import 'package:cake_wallet/utils/responsive_layout_util.dart';
+import 'package:flutter/material.dart';
+import 'package:cake_wallet/src/widgets/alert_background.dart';
+import 'package:cake_wallet/src/widgets/alert_close_button.dart';
+
+class ModalWidget extends StatelessWidget {
+  ModalWidget({required this.children, this.title});
+
+  final List<Widget> children;
+  final String? title;
+
+  @override
+  Widget build(BuildContext context) {
+    final double padding = 24;
+
+    final mq = MediaQuery.of(context);
+    final bottom = mq.viewInsets.bottom;
+    final height = mq.size.height - bottom;
+    final screenCenter = height / 2;
+
+    double closeButtonBottom = 60;
+    double containerHeight = height * 0.65;
+    if (bottom > 0) {
+      // increase a bit or it gets too squished in the top
+      containerHeight = height * 0.75;
+
+      final containerCenter = containerHeight / 2;
+      final containerBottom = screenCenter - containerCenter;
+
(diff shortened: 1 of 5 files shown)


lib/src/widgets/modal_widget.dart
```diff
@@ -0,0 +1,64 @@
...
+    double closeButtonBottom = 60;
+    double containerHeight = height * 0.65;
+    if (bottom > 0) {
+      // increase a bit or it gets too squished in the top
+      containerHeight = height * 0.75;
+
+      final containerCenter = containerHeight / 2;
+      final containerBottom = screenCenter - containerCenter;
+
+      final hasTitle = this.title == null || this.title!.isEmpty;
```
OmarHatem28, line 30: this is not right,
like this, the bool `hasTitle` will be true if it does NOT have title.

Also you are not using the string `title` in this class, so you can pass just the boolean

lib/src/widgets/modal_widget.dart
```diff
@@ -0,0 +1,64 @@
+import 'package:cake_wallet/utils/responsive_layout_util.dart';
+import 'package:flutter/material.dart';
+import 'package:cake_wallet/src/widgets/alert_background.dart';
+import 'package:cake_wallet/src/widgets/alert_close_button.dart';
+
+class ModalWidget extends StatelessWidget {
```
OmarHatem28, line 6: **Minor**
can we find a better descriptive name

lib/src/widgets/modal_widget.dart
```diff
@@ -0,0 +1,64 @@
...
+
+      final containerCenter = containerHeight / 2;
+      final containerBottom = screenCenter - containerCenter;
+
+      final hasTitle = this.title == null || this.title!.isEmpty;
+
+      // position the close button right below the search container
+      closeButtonBottom = closeButtonBottom -
+          containerBottom +
+          (hasTitle ? padding : padding / 1.5);
```
OmarHatem28, line 35: when you check the bool condition, please verify that this line is working as intended

OmarHatem28 (changes requested):

lib/src/widgets/picker_wrapper_widget.dart
```diff
@@ -0,0 +1,62 @@
...
+      // increase a bit or it gets too squished in the top
+      containerHeight = height * 0.75;
+
+      final containerCenter = containerHeight / 2;
+      final containerBottom = screenCenter - containerCenter;
+
+      // position the close button right below the search container
+      closeButtonBottom = closeButtonBottom -
+          containerBottom +
+          (hasTitle == null || !hasTitle! ? padding : padding / 1.5);
```
OmarHatem28, line 33: and then here you can just check if it does not have title `!hasTitle`

OmarHatem28 (approved):

PR #945, opened 2023-05-25
CW-214-Monero-wallet-rescan-bug

Serhii-Borodenko:

# Description
Clear transactionHistory before rescan
# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (changes requested):
since all wallets override the function `clear` with the same implementation, why don't we provide the implementation only in the main abstract class `cw_core/lib/transaction_history.dart`, this way we don't have to override it in each wallet and write the same code, and it will also apply on any future wallets automatically

OmarHatem28 (approved):

PR #959, opened 2023-06-02
CW-411 Update sideshift api

godilite:

# Description

# Pull Request - Checklist  

- [X] Format code

### DIFF
--- a/lib/exchange/sideshift/sideshift_exchange_provider.dart
+++ b/lib/exchange/sideshift/sideshift_exchange_provider.dart
@@ -65,12 +65,25 @@ class SideShiftExchangeProvider extends ExchangeProvider {
       if (amount == 0) {
         return 0.0;
       }
-      final fromCurrency = request.from.title.toLowerCase();
-      final toCurrency = request.to.title.toLowerCase();
-      final depositNetwork = _networkFor(request.depositMethod);
-      final settleNetwork = _networkFor(request.settleMethod);
+      final fromCurrency = from.title.toLowerCase();
+      final toCurrency = to.title.toLowerCase();
+      final depositNetwork = _networkFor(from);
+      final settleNetwork = _networkFor(to);
 
-      final url = apiBaseUrl + rangePath + '/' + fromCurrency + '-' + depositNetwork + '/' + toCurrency + '-' + settleNetwork + '?affiliateId=' + affiliateId + '&amount=' + amount;
+      final url = apiBaseUrl +
+          rangePath +
+          '/' +
+          fromCurrency +
+          '-' +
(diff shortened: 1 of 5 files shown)


lib/exchange/sideshift/sideshift_exchange_provider.dart
```diff
@@ -36,46 +37,52 @@ class SideShiftExchangeProvider extends ExchangeProvider {
...
+          '/' +
+          fromCurrency +
+          '-' +
+          depositNetwork +
+          '/' +
+          toCurrency +
+          '-' +
+          settleNetwork +
+          '?affiliateId=' +
+          affiliateId;
```
OmarHatem28, line 81: I don't see this as a query parameter for this API in their docs

lib/exchange/sideshift/sideshift_exchange_provider.dart
```diff
@@ -210,6 +248,18 @@ class SideShiftExchangeProvider extends ExchangeProvider {
...
+        from: to,
+        to: from,
+        isFixedRateMode: isFixedRateMode,
+      );
+      return Limits(
+        min: min != null ? (min * currentRate) : null,
+        max: max != null ? (max * currentRate) : null,
+      );
+    }
+
```
OmarHatem28, line 262: 1. `fetchRate` function is calling the same API, so why not just use the response we have directly?
2. why are we changing the min/max values if it's a fixed rate mode?

lib/exchange/sideshift/sideshift_exchange_provider.dart
```diff
@@ -280,28 +325,29 @@ class SideShiftExchangeProvider extends ExchangeProvider {
...
-        return 'usdtErc20';
-      case CryptoCurrency.usdttrc20:
-        return 'usdtTrc20';
-      case CryptoCurrency.usdcpoly:
-        return 'usdcpolygon';
-      case CryptoCurrency.usdcsol:
-        return 'usdcsol';
-      case CryptoCurrency.maticpoly:
+      default:
+        return currency.tag != null ? _normalizeTag(currency.tag!) : 'mainnet';
```
OmarHatem28, line 331: no need for a switch since its empty

OmarHatem28 (changes requested):

lib/exchange/sideshift/sideshift_exchange_provider.dart
SamsungGalaxyPlayer: I added this initially, but it's no longer needed after discussing with the SideShift team. It can be removed.

OmarHatem28 (approved):

PR #960, opened 2023-06-03
Cw-410 Null Issue With 2FA

Blazebrain:

# Description

# Pull Request - Checklist  

- [x] Format code


lib/src/screens/setup_2fa/setup_2fa_enter_code_page.dart
```diff
@@ -45,25 +45,31 @@ class TotpAuthCodePageState extends State<TotpAuthCodePage> {
   void initState() {
     _reaction ??= reaction((_) => widget.setup2FAViewModel.state, (ExecutionState state) {
       if (state is ExecutedSuccessfullyState) {
-        WidgetsBinding.instance.addPostFrameCallback((_) {
+       if(widget.totpArguments.onTotpAuthenticationFinished != null) {
+         WidgetsBinding.instance.addPostFrameCallback((_) {
```
OmarHatem28, line 49: why not just check for this before registering the reaction itself, so that it would only be 1 if condition

also, since all of them would need `WidgetsBinding.instance.addPostFrameCallback((_) {` then it would better make sense to move the if conditions that checks for the state to inside this function so we only have one instead of 3

OmarHatem28: and what is the case that the `onTotpAuthenticationFinished` would be passed as null? can you elaborate please

Blazebrain: Alright, will move the code around as suggested, reduces repetition.

`onTotpAuthenticationFinished` is passed as null in the initial setup, where the totp has not been previously set and the user just scans the QR Code. There is nothing to authenticate at that point as it's an initial setup and we're just activating the totp auth.

OmarHatem28 (approved):

PR #972, opened 2023-06-19
Remove defaultFiat

SamsungGalaxyPlayer:
Logic can be better handled by IP

# Description

# Pull Request - Checklist  

- [ ] Format code


SamsungGalaxyPlayer:
If someone in Australia is using the app and keeps their default fiat as USD instead of changing it to AUD, it would result in some issues with the onramps not working. Kinda silly. The IP is more reliable in practice at directing flow.

PR #978, opened 2023-06-27
Fix Bitcoin transactions not showing

OmarHatem28:

# Description

handle multiple responses coming in a single event

sometimes the events we receive in `socket!.listen((Uint8List event)` has more than one RPC response
ex.
```
{"jsonrpc": "2.0", "result": "f2c0cea5f.......1e4ca", "id": "4498"}
{"jsonrpc": "2.0", "result": "d5694e3f........87dd", "id": "4499"}
```
thus failing to parse this event as json, resulting in failure of parsing any response afterwards

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/cw_bitcoin/lib/electrum.dart
+++ b/cw_bitcoin/lib/electrum.dart
@@ -66,9 +66,14 @@ class ElectrumClient {
     socket!.listen((Uint8List event) {
       try {
         final msg = utf8.decode(event.toList());
-        final response =
-            json.decode(msg) as Map<String, dynamic>;
-        _handleResponse(response);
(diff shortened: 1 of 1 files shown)


OmarHatem28:
the other change for adding the Timeout was because if the user connection was interrupted during synchronization, the call would not complete, leading to `Future.wait` not finishing, thus no Transactions are shown.

what is currently implemented is that for any reason if a transaction failed to complete within 10 seconds, it will be considered as failed so that all the other transactions show correctly, and then when the user has a stable network connection, and does a `reconnect` the other missing transactions will show

mkyq (approved):

PR #985, opened 2023-07-03
CW-415 add coin control for xmr

godilite:

# Description

# Pull Request - Checklist  

- [ ] Format code


scripts/ios/build_monero.sh
```diff
@@ -4,7 +4,7 @@
 
 MONERO_URL="https://github.com/cake-tech/monero.git"
 MONERO_DIR_PATH="${EXTERNAL_IOS_SOURCE_DIR}/monero"
-MONERO_VERSION=release-v0.18.2.2
+MONERO_VERSION=add-coin-to-monero
```
OmarHatem28, line 7: revert this to the release branch

OmarHatem28 (changes requested):

cw_monero/lib/monero_wallet.dart
```diff
@@ -354,14 +377,93 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
...
+      address: coin.address,
+      value: coin.value,
+      vout: 0,
+      keyImage: coin.keyImage
+    );
+
+    await unspentCoinsInfo.add(newInfo);
+  }
+
+  Future<void> _refreshUnspentCoinsInfo() async {
```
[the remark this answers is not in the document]
konstantinullrich: Because the state of `isFrozen` is only stored in the Box

cw_monero/lib/monero_wallet.dart
```diff
@@ -354,14 +377,93 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
...
+          coin.isSending = coinInfo.isSending;
+          coin.note = coinInfo.note;
+        } else {
+          _addCoinInfo(coin);
+        }
+      });
+    }
+
+    await _refreshUnspentCoinsInfo();
+    _askForUpdateBalance();
```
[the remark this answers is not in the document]
konstantinullrich: The  _askForUpdateBalance is there to ensure that the frozen balance is reflected on the dashboard

lib/view_model/send/send_view_model.dart
```diff
@@ -374,7 +374,7 @@ abstract class SendViewModelBase with Store {
     final _priority = priority as TransactionPriority;
     final wallet = _wallet;
 
-    if (isElectrumWallet) {
+    if (hasCoinControl) {
```
OmarHatem28, line 377: this is not right, since it uses `bitcoin` instance, separating them into 2 variables would be better as it should stay checked by `isElectrumWallet` or see if this would be also applicable for Monero then let's add it

scripts/docker/build_all.sh
OmarHatem28, line 1: is this the correct format for this file? and the build haven scripts as well

scripts/docker/docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: '3.6'
+
+services:
+  build_deps:
+    image: build_monero_deps
+    environment:
+      MONERO_BRANCH: add-coin-to-monero
```
OmarHatem28, line 7: this should be reverted to the android release branch as well

OmarHatem28 (changes requested):

OmarHatem28:
also please check this 

> we can revert this now since it created the new externals and for the other workflow we can just modify the build_monero.sh file so that it can generate the new cache on main branch too
> 
> or better, we can add to the key of the cache, the hash of this file as well `cw_monero/ios/Classes/monero_api.cpp`
> 
> or if you have a better suggestion, I am open to discuss it

OmarHatem28:
there is something we missed, opening the `UnspentCoinsInfo` Hive box in `main.dart` has a condition that it's not MoneroOnly, i.e Cake Wallet only, but that should not be the case since we enabled that for Monero coin as well

.github/workflows/pr_test_build.yml
```diff
@@ -54,6 +54,7 @@ jobs:
             /opt/android/cake_wallet/cw_haven/ios/External
             /opt/android/cake_wallet/cw_monero/android/.cxx
             /opt/android/cake_wallet/cw_monero/ios/External
+            /opt/android/cake_wallet/cw_monero/ios/Classes/monero_api.cpp
```
OmarHatem28, line 57: nope. this will cache this file as well, but we don't need to cache this file since it's already in the git repo.

so Github stores cached data by a key, so that whenever you want to use this cached data from a workflow you just use that key, so I made that key attribute with the hash value for the script files of building monero and haven dependencies, since those dependencies are the one that are not checked in Git and are ignored

so we wanna add to the key value the hash of the `monero_api.cpp` file, since if we modified in this file then we probably modified in something that would need to build Monero dependencies again

OmarHatem28 (changes requested):

OmarHatem28 (approved):

PR #1007, opened 2023-07-20
feat: Introduce QR Scanning to add new node

Blazebrain:

# Description

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/lib/view_model/node_list/node_create_or_edit_view_model.dart
+++ b/lib/view_model/node_list/node_create_or_edit_view_model.dart
@@ -1,4 +1,5 @@
 import 'package:cake_wallet/core/execution_state.dart';
+import 'package:cake_wallet/entities/qr_scanner.dart';
 import 'package:cake_wallet/store/settings_store.dart';
 import 'package:hive/hive.dart';
 import 'package:mobx/mobx.dart';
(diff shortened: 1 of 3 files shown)


lib/view_model/node_list/node_create_or_edit_view_model.dart
```diff
@@ -152,4 +153,26 @@ abstract class NodeCreateOrEditViewModelBase with Store {
 
   @action
   void setAsCurrent(Node node) => _settingsStore.nodes[_walletType] = node;
+
+  @action
+  Future<void> scanQRCodeForNewNode() async {
+    String code = await presentQRScanner();
+
+    if (code.isEmpty) {
+      throw Exception('Unexpected scan QR code value: value is empty');
```
OmarHatem28, line 162: throwing an exception without being handled will cause the error popup to show instead of showing this error message to the user which is unreasonable.

also, you need to handle the case that the scanned QR code is not empty, but is not valid for this case

I sent you an example of the error file that will come to us if a user scanned a wrong QR code (a QR that does not comply with this code)

OmarHatem28: like wrapping the whole function in a try-catch block

OmarHatem28 (changes requested):

lib/view_model/node_list/node_create_or_edit_view_model.dart
Blazebrain: Alright,

Fixed it.

OmarHatem28 (approved):

PR #1013, opened 2023-07-26
CW-423 Use SensitiveClipboard for sensitive Stings

konstantinullrich:
# Description

Use Android 13 SensitiveClipboard for sensitive Stings

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -45,8 +45,8 @@ android {
 
     defaultConfig {
         applicationId appProperties['id']
-        minSdkVersion 21
-        targetSdkVersion 31
+        minSdkVersion 24
+        targetSdkVersion 33
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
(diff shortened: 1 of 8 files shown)


android/app/build.gradle
```diff
@@ -45,8 +45,8 @@ android {
 
     defaultConfig {
         applicationId appProperties['id']
-        minSdkVersion 21
-        targetSdkVersion 31
+        minSdkVersion 24
```
SamsungGalaxyPlayer, line 48: Should this be reverted to 21?

konstantinullrich: No, in order to set Extras to Clipboard descriptions it needs to be, at least, at API >= 24 (android N) :/

lib/src/screens/backup/backup_page.dart
```diff
@@ -52,8 +52,7 @@ class BackupPage extends BasePage {
                       child: Observer(
                           builder: (_) => GestureDetector(
                                 onTap: () {
-                                  Clipboard.setData(
-                                      ClipboardData(text: backupViewModelBase.backupPassword));
+                                  SensitiveClipboard.copy(backupViewModelBase.backupPassword);
```
OmarHatem28, line 55: check here if the platform is mobile then use this package if not, then use the old method

OmarHatem28 (changes requested):
Instead of going through all the places where you used the new package and adding the required check (if platform is mobile) you can make a utility function that handles the copy and call it instead and you can write the condition only once there

you can check the `share_util.dart` it's made for a reason close to this

OmarHatem28 (approved):

PR #1015, opened 2023-07-31
CW-438 add nano

fossephate:
Issue Number (if Applicable): Fixes #CW-438

# Description
This branch is based off of the Ethereum branch and so changes from that branch are also in this one

- [X] Import Bip39 (trust / ledger nano) seeds, as well as "nano standard" (nautilus, natrium, nault, etc.) seeds (uses balance to figure out which to import
- [X] Create new wallets (nano standard)
- [X] Delete wallets
- [X] View balances
- [X] View transaction history
- [X] View and edit RPC node
- [X] Send transactions
- [X] Receive transactions
- [X] Change Rep

# Pull Request - Checklist  

- [X] Format code


OmarHatem28 (changes requested):
These are the review comments so far, will continue the rest of the files after these are fixed

cw_core/lib/hive_type_ids.dart
```diff
@@ -1,13 +1,16 @@
...
+const TRANSACTION_TYPE_ID = 2;
+const TRADE_TYPE_ID = 3;
+const WALLET_INFO_TYPE_ID = 4;
+const WALLET_TYPE_TYPE_ID = 5;
+const TEMPLATE_TYPE_ID = 6;
+const EXCHANGE_TEMPLATE_TYPE_ID = 7;
+const ORDER_TYPE_ID = 8;
+const UNSPENT_COINS_INFO_TYPE_ID = 9;
+const ANONPAY_INVOICE_INFO_TYPE_ID = 10;
+const DERIVATION_TYPE_TYPE_ID = 15;
```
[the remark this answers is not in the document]
fossephate: this is fixed in:
https://github.com/cake-tech/cake_wallet/pull/1060

cw_nano/lib/pending_nano_transaction.dart
```diff
@@ -0,0 +1,42 @@
...
+  String get amountFormatted {
+    final String amt = NanoUtil.getRawAsUsableString(amount.toString(), NanoUtil.rawPerNano);
+    return amt;
+  }
+
+  String get accurateAmountFormatted {
+    final String amt = NanoUtil.getRawAsUsableString(amount.toString(), NanoUtil.rawPerNano);
+    final String acc = NanoUtil.getRawAccuracy(amount.toString(), NanoUtil.rawPerNano);
+    return "$acc$amt";
+  }
```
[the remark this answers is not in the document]
fossephate: this checks whether the displayed amount === RAW amount, and adds a tilde to represent that it isn't exact if necessary
basically, if there's an amount past 6 decimal places, amt = "~"
so if the amount is 100.123456 that will display as 100.123456
but if the amount is 100.1234567 it will return ~100.123456
the amount is NOT rounded

lib/view_model/restore/restore_mode.dart
```diff
@@ -1 +1 @@
-enum WalletRestoreMode { seed, keys, txids }
\ No newline at end of file
+enum WalletRestoreMode { seed, keys, txids, seedKey }
```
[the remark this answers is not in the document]
fossephate: this is just left over code from before I merged some of your later ethereum changes regarding the restore page
seedKey === mnemonic in hex form

cw_nano/lib/nano_wallet_creation_credentials.dart
```diff
@@ -0,0 +1,41 @@
...
+  String toString() => 'Failure to load the wallet.';
+}
+
+class NanoRestoreWalletFromKeysCredentials extends WalletCredentials {
+  NanoRestoreWalletFromKeysCredentials({
+    required String name,
+    required String password,
+    required this.seedKey,
+    this.derivationType,
+  }) : super(name: name, password: password);
```
OmarHatem28, line 37: shouldn't we also pass the derivationType to the base class here?

lib/view_model/dashboard/transaction_list_item.dart
```diff
@@ -57,33 +57,48 @@ class TransactionListItem extends ActionListItem with Keyable {
...
+        break;
+      case WalletType.nano:
+        final nanoTransaction = transaction as NanoTransactionInfo;
+        amount = calculateFiatAmountRaw(
+            cryptoAmount:
+                NanoUtil.getRawAsDecimal(nanoTransaction.amountRaw.toString(), NanoUtil.rawPerNano)
+                    .toDouble(),
+            price: price);
         break;
       case WalletType.ethereum:
```
OmarHatem28, line 103: duplicate case

lib/view_model/dashboard/transaction_list_item.dart
```diff
@@ -57,33 +57,48 @@ class TransactionListItem extends ActionListItem with Keyable {
...
+            price: price);
+        break;
+      case WalletType.nano:
+        final nanoTransaction = transaction as NanoTransactionInfo;
+        amount = calculateFiatAmountRaw(
+            cryptoAmount:
+                NanoUtil.getRawAsDecimal(nanoTransaction.amountRaw.toString(), NanoUtil.rawPerNano)
+                    .toDouble(),
+            price: price);
         break;
```
OmarHatem28, line 102: we cannot use any subpackage in the main code (`cw_ethereum`, `cw_nano`, etc...) as they are not defined to the project when building Monero.com, we instead use `ethereum/ethereum.dart` where the `ethereum!.` is coming from, same goes for the Nano Util, you must find another way to use it here if necessary, i.e through the nano instance that should be defined in `nano/nano.dart`

OmarHatem28 (changes requested):

cw_nano/lib/nano_subaddress_list.dart
OmarHatem28, line 1: this file is wrong on a few things:
- there is no `/api` directory in `cw_nano`
- thus there is no `subaddress_list`
- thus all the functionalities here are not working
- and the most important one, is that it's not used, so why is it here?

lib/view_model/transaction_details_view_model.dart
```diff
@@ -221,4 +232,18 @@ abstract class TransactionDetailsViewModelBase with Store {
...
     items.addAll(_items);
   }
+
+
+  void _addNanoListItems(TransactionInfo tx, DateFormat dateFormat) {
+    final _items = [
+      StandartListItem(title: S.current.transaction_details_transaction_id, value: tx.id),
+      StandartListItem(
+          title: S.current.transaction_details_date, value: dateFormat.format(tx.date)),
+      // StandartListItem(title: S.current.confirmed, value: (tx.confirmations > 0).toString()),
```
OmarHatem28, line 242: why is this commented?

lib/view_model/dashboard/transaction_list_item.dart
```diff
@@ -57,33 +55,49 @@ class TransactionListItem extends ActionListItem with Keyable {
...
+        amount = calculateFiatAmountRaw(
+            cryptoAmount: ethereum!.formatterEthereumAmountToDouble(transaction: transaction),
+            price: price);
+        break;
+      case WalletType.nano:
+        final nanoTransaction = transaction as dynamic;
+        dynamic NanoUtil = nano!.getNanoUtil();
+        amount = calculateFiatAmountRaw(
+            cryptoAmount:
+                NanoUtil.getRawAsDecimalInstance(nanoTransaction.amountRaw.toString(), NanoUtil.rawPerNanoInstance)
```
OmarHatem28, line 98: same here

lib/utils/payment_request.dart
```diff
@@ -10,16 +12,23 @@ class PaymentRequest {
...
       scheme = uri.scheme;
     }
 
+    dynamic NanoUtil = nano!.getNanoUtil();
+
+    if (address.contains("nano")) {
+      amount = NanoUtil.getRawAsUsableStringInstance(amount, NanoUtil.rawPerNanoInstance) as String;
+    } else if (address.contains("ban")) {
+      amount = NanoUtil.getRawAsUsableStringInstance(amount, NanoUtil.rawPerBananoInstance) as String;
+    }
```
OmarHatem28, line 25: this will throw an exception if nano is not enabled (ex. in Monero.com app)
also, same comment about using `getNanoUtil`

lib/src/screens/nano/rep_form.dart
OmarHatem28, line 1: unused, can be removed

lib/src/screens/nano/nano_change_rep_page.dart
```diff
@@ -0,0 +1,98 @@
...
+                                          rightButtonText: S.of(context).change,
+                                          leftButtonText: S.of(context).cancel,
+                                          actionRightButton: () => Navigator.pop(context, true),
+                                          actionLeftButton: () => Navigator.pop(context, false));
+                                    }) ??
+                                false;
+
+                            if (confirmed) {
+                              try {
+                                dynamic wallet = getIt.get<AppStore>().wallet!;
```
OmarHatem28, line 79: pass the wallet as a dependency to this class then, and please refrain from using `dynamic`
a simple `WalletBase` would do

lib/view_model/nano_account_list/nano_account_list_view_model.dart
OmarHatem28, line 1: **Minor**
you could have used the same `monero_account_list_view_model` and just rename it to be more generic

lib/di.dart
```diff
@@ -768,6 +828,16 @@ Future<void> setup({
   getIt.registerFactoryParam<WalletRestorePage, WalletType, void>(
       (type, _) => WalletRestorePage(getIt.get<WalletRestoreViewModel>(param1: type)));
 
+  getIt.registerFactoryParam<WalletRestoreChooseDerivationViewModel, List<DerivationInfo>, void>(
+      (derivations, _) =>
+          WalletRestoreChooseDerivationViewModel(derivationInfos: derivations));
+
+  getIt.registerFactoryParam<WalletRestoreChooseDerivationPage, dynamic, void>(
```
OmarHatem28, line 835: that's why I don't like dynamic.
so you are passing a `dynamic` parameter called `credentials` and is expecting a `List<DerivationInfo>` does that make sense?

OmarHatem28 (changes requested):

lib/src/screens/nodes/widgets/pow_node_list_row.dart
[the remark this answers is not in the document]
fossephate: these are all left over from when PowNode was a separate class from Node, I just forgot to remove them and change the types back to just node where necessary

lib/src/screens/restore/wallet_restore_page.dart
```diff
@@ -217,18 +232,34 @@ class WalletRestorePage extends BasePage {
...
             seedWords.length != WalletRestoreViewModelBase.electrumShortSeedMnemonicLength)) {
       return false;
     }
 
+    // bip39:
+    const validSeedLengths = [12, 15, 18, 21, 24];
+    if (walletRestoreViewModel.type == WalletType.bitcoin &&
+        !(validSeedLengths.contains(seedWords.length))) {
+      return false;
+    }
```
OmarHatem28, line 246: 15 and 21?

OmarHatem28 (changes requested):

cw_nano/lib/nano_wallet_service.dart
```diff
@@ -0,0 +1,160 @@
...
+      await file.delete(recursive: true);
+    }
+
+    final walletInfo = walletInfoSource.values
+        .firstWhere((info) => info.id == WalletBase.idFor(wallet, getType()));
+    await walletInfoSource.delete(walletInfo.key);
+  }
+
+  @override
+  Future<void> rename(String currentName, String password, String newName) async {
```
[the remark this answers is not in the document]
fossephate: It works (just tested)? is there something wrong?

cw_nano/lib/nano_wallet_service.dart
```diff
@@ -0,0 +1,160 @@
...
+        throw Exception("Invalid key length!");
+      }
+    }
+
+    DerivationType derivationType = credentials.derivationType ?? DerivationType.nano;
+    credentials.walletInfo!.derivationType = derivationType;
+
+    final wallet = await NanoWallet(
+      password: credentials.password!,
+      mnemonic: credentials.seedKey, // we can't derive the mnemonic from the key in all cases
```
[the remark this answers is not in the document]
fossephate: this case is extremely rare, basically nano-style mnemonics can go from mnemonic to hex and back, but non-nano (bip39) nano mnemonics can only go from mnemonic to hex, but you can't go from hex back to mnemonic

the majority of nano users use nano mnemonics, the only notable exceptions are ledger nano device users who have exported their seed, and trust wallet users
but even in theses cases they would have to manually convert their seed into it's hex form as these applications present them in their mnemonic form

this comment is essentially saying that if a user puts in a bip39 hex seed, we can't show them the mnemonic as going from hex->mnemonic isn't possible (and we should instead show the hex seed)
but again the percent of users this would affect is basically 0

this comment reminds me of a similar edge case I think I forgot to handle though, so I'll add that now

OmarHatem28 (changes requested):

cw_nano/lib/nano_client.dart
```diff
@@ -0,0 +1,454 @@
...
+
+      // construct the change block:
+      Map<String, String> changeBlock = {
+        "type": "state",
+        "account": ourAddress,
+        "previous": accountInfo["frontier"] as String,
+        "representative": repAddress,
+        "balance": accountInfo["balance"] as String,
+        "link": "0000000000000000000000000000000000000000000000000000000000000000",
+        "link_as_account": "nano_1111111111111111111111111111111111111111111111111111hifc8npp",
```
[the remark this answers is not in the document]
fossephate: link_as_account is just the link in account form
link here refers to the previous block in the block lattice, which normally refers to a receive block's corresponding send block
in this case it's null (represented as all 0's) because it's a change block and there is no corresponding send/receive

cw_nano/lib/nano_wallet.dart
```diff
@@ -0,0 +1,440 @@
...
+      }
+
+      runningBalance = runningBalance - amt;
+
+      final block = await _client.constructSendBlock(
+        amountRaw: amt.toString(),
+        destinationAddress: txOut.extractedAddress ?? txOut.address,
+        privateKey: _privateKey!,
+        balanceAfterTx: runningBalance,
+        previousHash: previousHash,
```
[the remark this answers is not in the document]
fossephate: this is intentional, previousHash refers to the previous block in the chain, and would be null on the first loop since we aren't constructing that block

it is super weird that this works though, this first block should have a valid previous field, but it doesn't

I think it gets filled in when the transaction submits because we don't have access to the node / getAccountInfo rpc here

I'll double check to make sure this works but pretty sure that's what's happening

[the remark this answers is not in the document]
fossephate: nvm it gets filled in immediately

TLDR: it's supposed to be null, but only on the first run, it's set after so that it's set for the subsequent runs
it's how we're able to submit multiple nano transactions at the same time from the same account, since we know the block hash of the unsubmitted blocks

cw_nano/lib/nano_wallet.dart
```diff
@@ -0,0 +1,440 @@
...
+        NanoAccountType.NANO,
+        block["account"]!,
+        block["previous"]!,
+        block["representative"]!,
+        BigInt.parse(block["balance"]!),
+        block["link"]!,
+      );
+
+      blocks.add(block);
+      runningAmount += amt;
```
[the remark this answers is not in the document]
fossephate: we do use it for the not enough balance check, it's just a few lines below

runningAmount starts at 0 at the beginning of the function and adds up to the total amount of nano trying to be sent, we then check if the total amount of nano being sent is greater than the total balance

lib/src/screens/restore/wallet_restore_from_keys_form.dart
```diff
@@ -62,7 +65,6 @@ class WalletRestoreFromKeysFromState extends State<WalletRestoreFromKeysFrom> {
     nameController.dispose();
     addressController.dispose();
     viewKeyController.dispose();
-    spendKeyController.dispose();
```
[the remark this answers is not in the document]
fossephate: I don't remember this change, but I think it's probably left over from when we both made similar changes (you made them for eth, I made them for nano) and the resulting refactoring I had to do

that's just a guess though, I don't know for sure

OmarHatem28 (approved):

PR #1045, opened 2023-08-14
Exchange improvements

SamsungGalaxyPlayer:
* Add USDT on Polygon/MATIC
* Use single USDT image (except for Tron, keeping red for now)
* Fix/add/improve exchange asset mapping


assets/images/usdt.png
OmarHatem28, line 1: ok yeah, we can remove this one

lib/exchange/simpleswap/simpleswap_exchange_provider.dart
```diff
@@ -227,13 +227,7 @@ class SimpleSwapExchangeProvider extends ExchangeProvider {
       case CryptoCurrency.zec:
         return 'zec';
       case CryptoCurrency.bnb:
-        return currency.tag!.toLowerCase();
-      case CryptoCurrency.usdterc20:
-        return 'usdterc20';
-      case CryptoCurrency.usdttrc20:
-        return 'usdttrc20';
-      case CryptoCurrency.usdcpoly:
-        return 'usdcpoly';
```
OmarHatem28, line 236: this would lead to interpreting the 3 of them as just `usdt` and as far as I can see there is no other way that simpleswap would know which network is this

OmarHatem28 (changes requested):

PR #1055, opened 2023-08-22
Add restore from private key to Ethereum

OmarHatem28:

# Description

Add restore from QR code for Ethereum in both seeds/keys

# Pull Request - Checklist  

- [x] Format code


lib/view_model/wallet_restore_view_model.dart
```diff
@@ -19,19 +16,20 @@ import 'package:cake_wallet/view_model/restore/restore_mode.dart';
...
     with _$WalletRestoreViewModel;
 
 abstract class WalletRestoreViewModelBase extends WalletCreationVM with Store {
   WalletRestoreViewModelBase(AppStore appStore, WalletCreationService walletCreationService,
       Box<WalletInfo> walletInfoSource,
       {required WalletType type})
-      : availableModes = (type == WalletType.monero || type == WalletType.haven)
-            ? WalletRestoreMode.values
-            : [WalletRestoreMode.seed],
+      : availableModes =
```
fossephate, line 26: There will be a merge conflict here with the nano branch, I think (since you can restore from keys on nano)

fossephate (approved):

lib/view_model/wallet_restore_view_model.dart
OmarHatem28: yup, we can then extract this into a getter (ex. `canRestoreWithKeys` or `hasMultipleRestoreModes` or whatever makes better name) and check for supported wallet types

PR #1058, opened 2023-08-22
Improve ChangeNOW network mapping

SamsungGalaxyPlayer:
# Description

Follows Trocador code to use the title instead of the const name, and to allow normalizing across the tag level instead of needing to map the `legacyTicker`

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/lib/exchange/changenow/changenow_exchange_provider.dart
+++ b/lib/exchange/changenow/changenow_exchange_provider.dart
@@ -68,8 +68,8 @@ class ChangeNowExchangeProvider extends ExchangeProvider {
       required CryptoCurrency to,
       required bool isFixedRateMode}) async {
     final headers = {apiHeaderKey: apiKey};
-    final normalizedFrom = normalizeCryptoCurrency(from);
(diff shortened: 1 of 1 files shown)


lib/exchange/changenow/changenow_exchange_provider.dart
```diff
@@ -68,14 +68,12 @@ class ChangeNowExchangeProvider extends ExchangeProvider {
...
     final headers = {apiHeaderKey: apiKey};
-    final normalizedFrom = normalizeCryptoCurrency(from);
-    final normalizedTo = normalizeCryptoCurrency(to);
     final flow = getFlow(isFixedRateMode);
     final params = <String, String>{
-      'fromCurrency': normalizedFrom,
-      'toCurrency': normalizedTo,
-      'fromNetwork': networkFor(from),
-      'toNetwork': networkFor(to),
+      'fromCurrency': _normalizeTitle(from),
```
OmarHatem28, line 73: **Minor:**
`_normalizeCurrency` would be more suitable as this is not a title

OmarHatem28 (approved):

PR #1080, opened 2023-09-08
Exolix integration

ProcyonLotor123:

# Description

Added Exolix exchange integration

# Pull Request - Checklist  

- [x] Format code


lib/exchange/exolix/exolix_exchange_provider.dart
```diff
@@ -0,0 +1,293 @@
...
+import 'package:cake_wallet/exchange/trade_request.dart';
+import 'package:cake_wallet/exchange/trade_state.dart';
+import 'package:cake_wallet/exchange/exolix/exolix_request.dart';
+import 'package:cake_wallet/exchange/exchange_provider_description.dart';
+
+class ExolixExchangeProvider extends ExchangeProvider {
+  ExolixExchangeProvider() : super(pairList: _supportedPairs());
+
+  static final apiKey = secrets.exolixApiKey;
+  static const apiBaseUrl = 'https://exolix.com/api';
```
OmarHatem28, line 19: this was causing an issue since you are trying to create an https uri (`Uri.https(...`) while use a base url that already includes `http`

also `/api` should not be in the base url, instead it should be in the below paths

OmarHatem28 (changes requested):
mostly looks good, those are my initial comments
1. Bug: cannot create a trade, the API response is `{"status":400,"errors":{"amount":"amount must not be less than 0"}}` although the amount sent is more than 0
2. the issue of the API base url that is mentioned below

and also if you can provide us with the documentation to your APIs that would be great.
Thanks in advance

lib/exchange/exolix/exolix_exchange_provider.dart
SamsungGalaxyPlayer: https://exolix.com/developers

OmarHatem28:
made a minor fix
and was able to create a trade, but you are only checking for the status code of `200` while the create trade API returned `201` so the app considered it a failure.
I added 201 to the create trade function check, but please verify if there are other possible status codes for it or other APIs as well

ProcyonLotor123:
possible api status codes:

`/api/v2/transactions` // 201, 400, 500
`/api/v2/rate` // 200, 400, 422

PR #1097, opened 2023-09-20
fix: Cake-2FA-setup-issue

Blazebrain:

# Description

# Pull Request - Checklist  

- [x] Format code


OmarHatem28 (commented):
what is the reason for clearing `totpSecretKey` and then check if it's empty before assigning a new value to it?

lib/view_model/set_up_2fa_viewmodel.dart
```diff
@@ -84,24 +85,17 @@ abstract class Setup2FAViewModelBase with Store {
...
-    final randomBase32Key = Utils.generateRandomBase32SecretKey(16);
-    _setBase32SecretKey(randomBase32Key);
-  }
-
   @action
-  void setUseTOTP2FA(bool value) {
-    _settingsStore.useTOTP2FA = value;
-  }
+  void generateSecretKey() {
+    totpSecretKey = Utils.generateRandomBase32SecretKey(16);
```
OmarHatem28, line 90: you are only setting the `totpSecretKey` here, and this function is only called in the setup screen, so when if a user setup the 2FA and then close the app and then re-open it, the `totpSecretKey` won't get initialized, thus it will fail to verify the code that the user enters

OmarHatem28 (changes requested):
yup, this way it's better and cleaner, thanks.

Please check the issue below and please verify the flow is working correctly in the whole flow
- setup flow
- and after closing and opening the app

OmarHatem28:
@SamsungGalaxyPlayer please can you give this another test when possible, for the whole flow, let's try closing the app and then opening it again and see if it will maintain the TOTP setup correctly

OmarHatem28 (approved):

PR #1100, opened 2023-09-22
Cw 477 add ens

fossephate:
Issue Number (if Applicable): Fixes # CW-477

# Description
Adds ENS support for ethereum

# Pull Request - Checklist  

- [ ] Format code


cw_ethereum/lib/ethereum_wallet.dart
```diff
@@ -508,4 +508,8 @@ abstract class EthereumWalletBase
   @override
   String signMessage(String message, {String? address = null}) =>
       bytesToHex(_ethPrivateKey.signPersonalMessageToUint8List(ascii.encode(message)));
+
+  Future<String> fetchEnsAddress(String name) async {
+    return await _client.checkEnsName(name);
+  }
```
SamsungGalaxyPlayer, line 514: ENS names can contain addresses for different assets other than Ethereum (example: https://app.ens.domains/vikrant.eth). Move this to `/lib/entities/ens_record.dart`

lib/entities/parse_address_from_domain.dart
```diff
@@ -104,6 +107,15 @@ class AddressResolver {
...
       }
+      if (text.contains(".")) {
+        var wallet = getIt.get<AppStore>().wallet!;
+        if (wallet.type == WalletType.ethereum) {
+          final address = await ethereum!.fetchEnsAddress(wallet, text);
+          if (address.isNotEmpty) {
+            return ParsedAddress.fetchEnsAddress(name: text, address: address);
+          }
+        }
+      }
```
SamsungGalaxyPlayer, line 118:
```
      if (text.endsWith(".eth")) {
        final address = await fetchEnsAddress(text, ticker);
        return ParsedAddress.fetchEnsAddress(name: text, address: address);
      }
```

If these services are resolved in order (top to bottom), then move ENS above OpenAlias, so we can skip an attempted ENS lookup if the domain is `.ens`. We don't need to do an OpenAlias lookup if the TLD is an Unstoppable/ENS TLD, since these TLDs don't exist on the "Regular" internet.

lib/ethereum/cw_ethereum.dart
```diff
@@ -131,4 +131,9 @@ class CWEthereum extends Ethereum {
   void updateEtherscanUsageState(WalletBase wallet, bool isEnabled) {
     (wallet as EthereumWallet).updateEtherscanUsageState(isEnabled);
   }
+
+  @override
+  Future<String> fetchEnsAddress(WalletBase wallet, String name) async {
+    return (wallet as EthereumWallet).fetchEnsAddress(name);
+  }
```
SamsungGalaxyPlayer, line 138: ENS names can contain addresses for different assets other than Ethereum (example: https://app.ens.domains/vikrant.eth). Move this to `/lib/entities/ens_record.dart`

SamsungGalaxyPlayer (dismissed):
The ENS lookup related code will need to be taken out of the Ethereum wallet, since ENS can store address records for non-ETH assets as well.

SamsungGalaxyPlayer:
This does not properly decode the non-ETH addresses, but I have made a note of this on the guide page. Fixing this now would not be worth the effort.

https://guides.cakewallet.com/docs/advanced-features/special-address-services/#ethereum-name-service-ens

cw_ethereum/lib/ethereum_client.dart
```diff
@@ -208,6 +209,25 @@ I/flutter ( 4474): Gas Used: 53000
...
+    try {
+      final ens = Ens(client: _client!);
+
+      final addr = await ens.withName(ensName).getAddress();
+      return addr.hex;
+    } catch (e) {
+      print(e);
+      return "";
+    }
+  }
```
OmarHatem28, line 225: not used since you extracted it

cw_ethereum/lib/ethereum_client.dart
```diff
@@ -208,6 +209,25 @@ I/flutter ( 4474): Gas Used: 53000
...
+
+  dynamic getWeb3Client() {
```
OmarHatem28, line 227: why dynamic?! you are already using `Web3Client` in the main lib

cw_ethereum/lib/ethereum_wallet.dart
```diff
@@ -508,4 +508,6 @@ abstract class EthereumWalletBase
   @override
   String signMessage(String message, {String? address = null}) =>
       bytesToHex(_ethPrivateKey.signPersonalMessageToUint8List(ascii.encode(message)));
+
+  dynamic getWeb3Client() => _client.getWeb3Client();
```
OmarHatem28, line 512: same
also after modifying it, please follow the flow for any modification needed (ex. casting, etc..)

lib/entities/ens_record.dart
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    if (_client == null) {
+      _client = Web3Client("https://ethereum.publicnode.com", Client());
+    }
+
+    try {
+      final ens = Ens(client: _client);
+
+      dynamic res;
```
OmarHatem28, line 23: unused

lib/entities/ens_record.dart
```diff
@@ -0,0 +1,48 @@
+import 'package:cake_wallet/ethereum/ethereum.dart';
+import 'package:cw_core/wallet_base.dart';
+import 'package:cw_core/wallet_type.dart';
+import 'package:ens_dart/ens_dart.dart';
+import 'package:http/http.dart';
+import 'package:web3dart/web3dart.dart';
+
+class EnsRecord {
+  static Future<String> fetchEnsAddress(String name, {WalletBase? wallet}) async {
```
OmarHatem28, line 9: **Enhancement:**
`wallet` can't be null, you already have the null check `!` added before using it in `parse_address_from_domain` so you can remove the nullability if condition and so

lib/entities/parse_address_from_domain.dart
```diff
@@ -96,6 +100,14 @@ class AddressResolver {
         return ParsedAddress.fetchUnstoppableDomainAddress(address: address, name: text);
       }
 
+      if (text.endsWith(".eth")) {
+        var wallet = getIt.get<AppStore>().wallet!;
+        final address = await EnsRecord.fetchEnsAddress(text, wallet: wallet);
```
OmarHatem28, line 105: also you don't need to pass the whole wallet, you only need the wallet type which is already existing here, so you can remove this and just pass the already existing wallet type

lib/entities/parse_address_from_domain.dart
```diff
@@ -1,9 +1,13 @@
 import 'package:cake_wallet/core/address_validator.dart';
 import 'package:cake_wallet/core/yat_service.dart';
+import 'package:cake_wallet/di.dart';
+import 'package:cake_wallet/entities/ens_record.dart';
 import 'package:cake_wallet/entities/openalias_record.dart';
 import 'package:cake_wallet/entities/parsed_address.dart';
 import 'package:cake_wallet/entities/unstoppable_domain_address.dart';
 import 'package:cake_wallet/entities/emoji_string_extension.dart';
+import 'package:cake_wallet/ethereum/ethereum.dart';
+import 'package:cake_wallet/store/app_store.dart';
```
OmarHatem28, line 10: no use for those and for di then

OmarHatem28 (changes requested):

lib/entities/parse_address_from_domain.dart
fossephate: I pass the wallet because that way we can use whatever the current ethereum node is if we are on an ethereum wallet, rather than just rely on the one that's hardcoded in ens_record.dart

lib/entities/ens_record.dart
fossephate: wallet should be nullable, having the ! there was a mistake
the intention being that this function be callable anywhere without the need for an existing wallet just to resolve an ens address

lib/entities/parse_address_from_domain.dart
OmarHatem28: ok, but let's not use a dependency like this, injecting it would be better for maintainability, please check the last commit

OmarHatem28 (approved):

PR #1107, opened 2023-09-29
CW-491-Send-to-Mastodon-username-addresses

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [x] Format code


OmarHatem28 (commented):
Everything looks great, just one minor thing for making the code more robust and reliable.
so now in `lib/mastodon/mastodon_user.dart` you added all the parameters for `MastodonUser` meanwhile you are only using `id` and `note` so let's keep it simple at only the ones you are using so that we minimize errors that might happen from changes in other parameters, especially that you are expecting the actual values and not nullable values

OmarHatem28:
and please solve the conflict with main

OmarHatem28 (approved):

PR #1126, opened 2023-10-13
Cw 473 exchange refactoring

konstantinullrich:
# Description

Refactor parts of the exchange code, to be simpler. Remove deprecated Exchanges like MorphToken and XMR.to

# Pull Request - Checklist  

- [x] Format code


OmarHatem28 (changes requested):

lib/exchange/provider/changenow_exchange_provider.dart
```diff
@@ -1,33 +1,31 @@
...
-                    .where((i) => i != CryptoCurrency.xhv)
-                    .map((k) => ExchangePair(from: i, to: k, reverse: true)))
-                .expand((i) => i)
-                .toList());
+  ChangeNowExchangeProvider({required SettingsStore settingsStore})
+      : _settingsStore = settingsStore, _lastUsedRateId = '',
+        super(pairList: supportedPairs(_notSupported));
+
+  static const List<CryptoCurrency> _notSupported = [
+    CryptoCurrency.zaddr,
```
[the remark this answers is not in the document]
konstantinullrich: `zaddr` refers to Shielded Zcash which is rarely supported. There is no official documentation about the support of shielded Zcash addresses, but when I tried entering a valid shielded address it threw the error Invalid address. When I used an unshielded address the exchange worked.

OmarHatem28 (approved):
Looks good, thanks for refactoring what's needed.
Please solve the conflicts with main

lib/view_model/exchange/exchange_trade_view_model.dart
```diff
@@ -115,16 +109,15 @@ abstract class ExchangeTradeViewModelBase with Store {
 
   @action
   Future<void> _updateTrade() async {
+    if (_provider == null) throw ExchangeProviderUnsupportedException(trade.provider);
+
```
OmarHatem28, line 113: you haven't added `TradeProviderUnsupportedItem` to the items in this class.

also, there is no need for this line anymore

lib/view_model/trade_details_view_model.dart
```diff
@@ -60,11 +54,12 @@ abstract class TradeDetailsViewModelBase with Store {
         break;
     }
 
-    _updateItems();
+    if (_provider != null) {
+      _updateItems();
```
OmarHatem28, line 58: I believe this should be swapped with `_updateTrade()` on line 62

lib/view_model/trade_details_view_model.dart
```diff
@@ -83,12 +78,14 @@ abstract class TradeDetailsViewModelBase with Store {
 
   @action
   Future<void> _updateTrade() async {
+    if (_provider == null) throw ExchangeProviderUnsupportedException(trade.provider);
```
OmarHatem28, line 81: and this can be removed by the change above, since we will be sure its not null

OmarHatem28 (changes requested):
I keep forgetting to tell you this, but please keep the brackets for the conditions (`if` and `else`) as they are more readable this way, even for a 1 line of code

OmarHatem28 (approved):

PR #1133, opened 2023-10-17
use working port for seths node

hundehausen:

# Description
The public node of Seth is broken on port 18089. Port 443 is working with SSL.

# Pull Request - Checklist  

- [x] Format code


SamsungGalaxyPlayer (changes requested):
This needs the toggle to enable SSL as well

hundehausen:
Hey Justin,

is it

`use_ssl: true` ?

OmarHatem28 (approved):
Thanks for fixing this.
Also, it's worth mentioning this change would be applied for new installs only, as we only read the data from these files once the app is installed, and the user has the free will to modify or delete any of them, so in order to maintain the user changes, we won't enforce this change for current users

PR #1140, opened 2023-10-20
CW-502-Add-Monero-Account-Management-to-Home-Screen

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [x] Format code


lib/src/screens/dashboard/widgets/balance_page.dart
```diff
@@ -28,47 +31,58 @@ class BalancePage extends StatelessWidget {
           !dashboardViewModel.balanceViewModel.isReversing,
       child: SingleChildScrollView(
         child: Column(
-          crossAxisAlignment: CrossAxisAlignment.start,
+          crossAxisAlignment: CrossAxisAlignment.center,
           children: [
-            SizedBox(height: 56),
-            Container(
+            Observer(
+            builder: (_) => dashboardViewModel.type != WalletType.monero
```
OmarHatem28, line 37: **Enhancement:**
for scalability, we can put this into the view model under a variable called `hasAccounts` 
this way we maintain separation of concerns as the logic would be in the view model
and we would maintain scalability so that when we get to add new wallet that has accounts feature we can just add that to the variable in the view model.

and then we can make just a simple if condition for a centered `HomeScreenAccountWidget`

something like this variable `isHomeScreenSettingsEnabled` in line `61`

OmarHatem28 (commented):
also please format the code

lib/src/screens/dashboard/widgets/balance_page.dart
Serhii-Borodenko: fixed

OmarHatem28 (approved):

PR #1177, opened 2023-11-13
CW-526-Add-Trocador-providers-option-in-privacy-settings

Serhii-Borodenko:

# Description

This update allows users to selectively enable or disable Trocador providers in the settings menu.

# Pull Request - Checklist  

- [x] Format code


lib/src/widgets/setting_actions.dart
OmarHatem28, line 1: Lets change this to be inside of the privacy page as a tile like `Domain lookups` please.
and then you can delete the added icon from the assets directory as well

lib/view_model/settings/trocador_providers_view_model.dart
```diff
@@ -0,0 +1,22 @@
...
+  final SettingsStore _settingsStore;
+
+  @computed
+  Map<String, bool> get providerStates => _settingsStore.trocadorProviderStates;
+
+  @action
+  void toggleProviderState(String providerName) {
+    final currentState = providerStates[providerName] ?? false;
+    _settingsStore.trocadorProviderStates[providerName] = !currentState;
+    _settingsStore.saveTrocadorProviderState(providerName, !currentState);
```
OmarHatem28, line 20: **Enhancement:**
this is a duplicate and can be enhanced so we dont have to call 2 functions whenever we use this.
either leave the first line and add a reaction in the settings store that will save the state
or remove the first line since the save function already updates the `trocadorProviderStates`

lib/exchange/provider/trocador_exchange_provider.dart
```diff
@@ -153,6 +170,22 @@ class TrocadorExchangeProvider extends ExchangeProvider {
...
+          firstAvailableProvider = provider as String;
+          break;
+        }
+      }
+
+      if (firstAvailableProvider.isEmpty) {
+        throw Exception('No available provider is enabled');
+      }
+
+      params['provider'] = firstAvailableProvider;
```
OmarHatem28, line 187: please fix the indentation.

and are we sure that `firstAvailableProvider` would have the best rate? or should we loop through the enabled providers and their rates and choose depending on the best rate?

lib/exchange/provider/trocador_exchange_provider.dart
```diff
@@ -12,11 +12,25 @@ import 'package:cw_core/crypto_currency.dart';
...
+    'StealthEx',
+    'Simpleswap',
+    'ChangeNow',
+    'FixedFloat',
+    'LetsExchange',
+    'Exolix',
+    'Godex',
+    'Exch',
+    'CoinCraddle'
+  ];
```
OmarHatem28, line 33: do they have an API to get this data dynamically?

OmarHatem28 (changes requested):

lib/exchange/provider/trocador_exchange_provider.dart
Serhii-Borodenko: I didn't find an API for fetching a list of providers. The new_rate method fetches a new list of providers each time for a specific pair of assets and amount.

lib/view_model/settings/trocador_providers_view_model.dart
Serhii-Borodenko: fixed

lib/exchange/provider/trocador_exchange_provider.dart
Serhii-Borodenko: The list of available providers for a trade is sorted according to the 'amount_to' parameter, placing the highest amounts at the top. Therefore, it's already sorted to highlight the best conditions. Since we fetch the first allowed provider from this list, it seems we are indeed getting the best rate.

OmarHatem28 (approved):

PR #1186, opened 2023-11-20
update wallet restore page to be less confusing for nano users

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_nano/lib/nano_wallet.dart
```diff
@@ -85,22 +85,26 @@ abstract class NanoWalletBase
...
     }
 
-    if (_seedKey == null) {
+    if (_hexSeed == null) {
       if (_derivationType == DerivationType.nano) {
-        _seedKey = bip39.mnemonicToEntropy(_mnemonic).toUpperCase();
+        _hexSeed = bip39.mnemonicToEntropy(_mnemonic).toUpperCase();
       } else {
-        _seedKey = await NanoUtil.hdMnemonicListToSeed(_mnemonic.split(' '));
+        _hexSeed = await NanoUtil.hdMnemonicListToSeed(_mnemonic.split(' '));
```
OmarHatem28, line 103: shouldn't this be the other way around?
you are checking if the derivation type is Nano, then you get the hex form from the mnemonic via bip39 ?!
and `else` (if the derivation type is bip39) you get the hex form through the NanoUtil?!!

lib/view_model/wallet_keys_view_model.dart
```diff
@@ -172,7 +171,10 @@ abstract class WalletKeysViewModelBase with Store {
     final restoreHeightResult = await restoreHeight;
     return {
       if (_appStore.wallet!.seed != null) 'seed': _appStore.wallet!.seed!,
-      if (_appStore.wallet!.privateKey != null) 'private_key': _appStore.wallet!.privateKey!,
+      if (_appStore.wallet!.seed == null && _appStore.wallet!.hexSeed != null)
+        'hexSeed': _appStore.wallet!.hexSeed!,
```
OmarHatem28, line 175: did you handle this case in restoring from QR?

OmarHatem28 (changes requested):

cw_nano/lib/nano_wallet.dart
fossephate: this is correct, the names for the derivation types are just not very good since the "nano standard" way of deriving keys still uses Bip39, it's just the derivation paths and signing algorithm are different

you can read more about it here: https://docs.nano.org/integration-guides/key-management/

lib/view_model/wallet_keys_view_model.dart
fossephate: ah nice catch, I forgot to add this
latest pushes address this

OmarHatem28 (approved):

PR #1227, opened 2023-12-11
Cw 519 tor

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/utils/proxy_wrapper.dart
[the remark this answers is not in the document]
OmarHatem28: also `getProxyInstance` is marked as future but it doesn't need to be one

OmarHatem28 (changes requested):

OmarHatem28:
also please change the name of the tile that navigates to the Tor screen, I just used `tor only` instead of adding a new name since it was a trial

OmarHatem28 (changes requested):
we can remove the `Tor only` option from Fiat API and Exchange API and make sure of handling users who had it set to `Tor only` after you remove it to be migrated to enabled and turn on `Onion only` if it wouldn't complicate things.
and please check Tux's comment on Jira

cw_core/lib/node.dart
```diff
@@ -63,6 +63,8 @@ class Node extends HiveObject with Keyable {
   @HiveField(6)
   String? socksProxyAddress;
 
+  bool connectOverTorOnly = false;
```
OmarHatem28, line 66: this will not work in many cases (switching wallets, switching nodes for example)
let's find a better way instead of having this variable as a part of the node object, or at least handle other possible cases

OmarHatem28 (changes requested):

lib/core/wallet_loading_service.dart
```diff
@@ -46,6 +53,20 @@ class WalletLoadingService {
...
+
+    if ([WalletType.bitcoin, WalletType.litecoin].contains(type)) {
+      bitcoin!.setTorOnly(wallet, mode == TorConnectionMode.torOnly);
+    }
+
+    final status = torViewModel.torConnectionStatus;
+    if (status == TorConnectionStatus.connected || status == TorConnectionStatus.connecting) {
+      // connect the node to the tor proxy:
+      await torViewModel.startTor();
+    }
```
OmarHatem28, line 68: start tor if already connected or connecting?

lib/core/wallet_loading_service.dart
```diff
@@ -46,6 +53,20 @@ class WalletLoadingService {
       await updateMoneroWalletPassword(wallet);
     }
 
+    TorConnectionMode mode = TorConnectionMode.deserialize(
+      raw: sharedPreferences.getInt(PreferencesKey.currentTorConnectionModeKey) ?? 0,
+    );
+
+    if ([WalletType.bitcoin, WalletType.litecoin].contains(type)) {
+      bitcoin!.setTorOnly(wallet, mode == TorConnectionMode.torOnly);
+    }
```
OmarHatem28, line 62: bitcoin cash would need to have the same

OmarHatem28 (changes requested):

lib/core/wallet_loading_service.dart
fossephate: the comment explains what's happening here (we call this just to connect the node to the proxy)
but I agree it should be more clear and I just forgot that I extracted that out to it's own function so the latest commit just calls that function directly

cw_core/lib/node.dart
```diff
@@ -207,19 +214,26 @@ class Node extends HiveObject with Keyable {
...
 
     String? proxy = socksProxyAddress;
 
-    if ((proxy?.isEmpty ?? true) && Tor.instance.enabled) {
+    // we have to be careful here because if Tor.instance.port is called and Tor isn't enabled
+    // we'll create a new tor instance that just eats up memory
+    // we initialize tor in the tor view model first so that this doesn't happen
+    if (!Tor.instance.enabled || Tor.instance.port == -1) {
+      return false;
+    }
```
OmarHatem28, line 228: this neglects the case where the user has an external tor running and added his own proxy address

lib/view_model/settings/tor_view_model.dart
```diff
@@ -0,0 +1,117 @@
...
+      await connectOrDisconnectNodeToProxy(connect: true);
+    } catch (e) {
+      torConnectionStatus = TorConnectionStatus.disconnected;
+    }
+  }
+
+  @action
+  Future<void> stopTor() async {
+    torInstance.disable();
+    // setting the torConnectionMode to disabled will prevent anything from actually using the proxy
```
OmarHatem28, line 111: isn't that what we want?
if so, then I believe you should set the torConnectionMode to disabled

fossephate: I think this comment is just poorly worded, this function is triggered whenever the user sets torConnectionMod to disabled

Setting it again in this function would (I think) cause an infinite loop

PR #1249, opened 2023-12-28
CW-545-Overriding-Erc20-tokens-list

Blazebrain:

# Description

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/cw_ethereum/lib/ethereum_wallet.dart
+++ b/cw_ethereum/lib/ethereum_wallet.dart
@@ -123,21 +123,25 @@ abstract class EthereumWalletBase
     ethereumErc20TokensBox = await CakeHive.openBox<Erc20Token>(
         "${walletInfo.name.replaceAll(" ", "_")}_${Erc20Token.ethereumBoxName}");
 
-    // Check if the box is not empty, if it's not empty, it means this flow has been run once.
-    // If it is empty, means it's a totally new box and we can proceed
-    if (ethereumErc20TokensBox.isNotEmpty) return;
-
     //Open the previous token configs box
     erc20TokensBox = await CakeHive.openBox<Erc20Token>(Erc20Token.boxName);
 
     // Check if it's empty, if it is, we stop the flow and return.
-    if (erc20TokensBox.isEmpty) return;
+    if (erc20TokensBox.isEmpty) {
+      // If it's empty, but the new wallet specific box is also empty,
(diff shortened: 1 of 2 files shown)


cw_ethereum/lib/ethereum_wallet.dart
```diff
@@ -114,6 +117,33 @@ abstract class EthereumWalletBase
...
+    //Open the previous token configs box
+    erc20TokensBox = await CakeHive.openBox<Erc20Token>(Erc20Token.boxName);
+
+    // Check if it's empty, if it is, we stop the flow and return.
+    if (erc20TokensBox.isEmpty) {
+      // If it's empty, but the new wallet specific box is also empty,
+      // we load the initial tokens to the new box.
+      if (ethereumErc20TokensBox.isEmpty) addInitialTokens();
+      return;
+    }
```
OmarHatem28, line 135: if the old box is empty then it has 2 cases:
1. wallet is just created/restored (and in this case `wallet.addInitialTokens()` will be called anyway so no need to call it here)
2. if it's an already existing wallet, then the user himself deleted the erc20 tokens (and in this case, you will re-add to him all the initial erc-20 tokens

so just check if the old box is empty then `return;` without any further steps.

OmarHatem28 (changes requested):

cw_ethereum/lib/ethereum_wallet.dart
Blazebrain: There is a third case or scenario:

- There are two ETH wallets A and B
- When we open the app on Wallet A, the app creates a wallet-specific Box for Wallet A, and moves the tokens list and configs in the old erc20tokens box to the wallet-specific Box for Wallet A.
- Now the Box for wallet A has the old tokens listing and their configs, while the old Box is now empty.
- When we switch now from Wallet A to Wallet B, we create a wallet-specific Box for Wallet B. 
- However, this box is empty (since it's new), and the old box that had the previous tokens list and configs is empty, which means that we have to re-add the initial list of tokens to the newly created tokens Box for Wallet B. If not, we won't see any token listed out in the tokens listing page.

OmarHatem28: yup, that's a valid case, good job, thanks for pointing it out

OmarHatem28 (approved):

PR #1269, opened 2024-01-18
Cw 540 encrypt shared

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/di.dart
+++ b/lib/di.dart
@@ -276,6 +276,7 @@ Future<void> setup({
 
   if (!_isSetupFinished) {
     getIt.registerSingletonAsync<SharedPreferences>(() => SharedPreferences.getInstance());
+    getIt.registerSingleton<FlutterSecureStorage>(secureStorage);
   }
   if (!_isSetupFinished) {
     getIt.registerFactory(() => BackgroundTasks());
@@ -302,7 +303,6 @@ Future<void> setup({
   getIt.registerFactory<Box<Node>>(() => _nodeSource);
   getIt.registerFactory<Box<Node>>(() => _powNodeSource, instanceName: Node.boxName + "pow");
 
-  getIt.registerSingleton<FlutterSecureStorage>(secureStorage);
   getIt.registerSingleton(AuthenticationStore());
   getIt.registerSingleton<WalletListStore>(WalletListStore());
   getIt.registerSingleton(NodeListStoreBase.instance);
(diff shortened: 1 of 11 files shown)


lib/di.dart
```diff
@@ -276,6 +276,7 @@ Future<void> setup({
 
   if (!_isSetupFinished) {
     getIt.registerSingletonAsync<SharedPreferences>(() => SharedPreferences.getInstance());
+    getIt.registerSingleton<FlutterSecureStorage>(secureStorage);
```
OmarHatem28, line 279: nothing to be changed, but was there a reason for this change? or even for those `if (!_isSetupFinished)` conditions?

fossephate: it's so that we have an instance of secureStorage when loading the settingsStore
I'm not totally sure what the _isSetupFinished is all about, I just assumed it should follow the same pattern that sharedpreferences does and that it'd fit here

OmarHatem28:
ok yeah I get it now, so shared prefs, secure storage, background tasks are initialized at first because they are used in the settingsStore and the reason why we have this condition `_isSetupFinished` is that, the `setup` function in `di.dart` can be called again if the user restored from backup, so to avoid getting an error that this class is already registered from `getIt` we have this condition.

lib/store/settings_store.dart
```diff
@@ -1188,6 +1228,63 @@ abstract class SettingsStoreBase with Store {
...
+        sharedPreferences.getString(SecureKey.totpSecretKey) ??
+        totpSecretKey;
+    allowBiometricalAuthentication =
+        ((await _secureStorage.read(key: SecureKey.allowBiometricalAuthenticationKey)) as bool?) ??
+            sharedPreferences.getBool(SecureKey.allowBiometricalAuthenticationKey) ??
+            allowBiometricalAuthentication;
+    selectedCake2FAPreset = Cake2FAPresetsOptions.deserialize(
+        raw: ((await _secureStorage.read(key: SecureKey.selectedCake2FAPreset)) as int?) ??
+            sharedPreferences.getInt(SecureKey.selectedCake2FAPreset) ??
+            Cake2FAPresetsOptions.normal.raw);
```
OmarHatem28, line 1246: **Minor:**
you can use the same `SecureKey.get...` as you did above

OmarHatem28 (approved):

PR #1272, opened 2024-01-21
CW-555-Add-Solana-Wallet

Blazebrain:

# Description
- Adding Solana Wallet to CakeWallet

- Create Wallet
- Restore from Seeds
- Restore from Keys
- Restore from QR
- Send Transactions
- Receive Transactions
- SPL Tokens (WIP)

# Pull Request - Checklist  

- [ ] Format code


cw_solana/lib/default_spl_tokens.dart
OmarHatem28, line 1: like this, all of them would be enabled by default

cw_solana/lib/default_spl_tokens.dart
```diff
@@ -0,0 +1,107 @@
...
+class DefaultSPLTokens {
+  final List<SPLToken> initialSPLTokens = [
+    SPLToken(
+      name: 'USDT',
+      symbol: 'USDT',
+      mintAddress: 'Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB',
+      decimal: 6,
+      mint: 'usdt',
+      logoUrl:
+          "https://raw.githubusercontent.com/solana-labs/token-list/main/assets/mainnet/Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB/logo.svg",
```
OmarHatem28, line 12: nope, do not depend on a network image, either download the asset (I only recommend this for important tokens only and that is if we don't have the icon already) else please do like Ethereum, check if the token already is defined in `CryptoCurrency.dart` and take the image from there, else leave it as null
(that's for all of the tokens)

cw_solana/lib/pending_solana_transaction.dart
```diff
@@ -0,0 +1,38 @@
...
+
+    return stringifiedAmount;
+  }
+
+  @override
+  Future<void> commit() async {}
+
+  @override
+  String get feeFormatted {
+    return '';
```
OmarHatem28, line 30: my knowledge is that Solana does have fees

lib/entities/priority_for_wallet_type.dart
```diff
@@ -27,6 +27,8 @@ List<TransactionPriority> priorityForWalletType(WalletType type) {
       return [];
     case WalletType.polygon:
       return polygon!.getTransactionPriorities();
+    case WalletType.solana:
```
OmarHatem28, line 30: doesn't solana has prioritie fees as well?

cw_solana/lib/solana_transaction_info.dart
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @override
+  String fiatAmount() => _fiatAmount ?? '';
+
+  @override
+  void changeFiatAmount(String amount) => _fiatAmount = formatAmount(amount);
+
+  @override
+  String feeFormatted() => '';
```
OmarHatem28, line 49: same

cw_solana/lib/solana_transaction_info.dart
```diff
@@ -0,0 +1,74 @@
...
+
+  Map<String, dynamic> toJson() => {
+        'id': id,
+        'amount': solAmount.toString(),
+        'direction': direction.index,
+        'blockTime': blockTime.millisecondsSinceEpoch,
+        'isPending': isPending,
+        'tokenSymbol': tokenSymbol,
+        'to': to,
+        'from': from,
```
OmarHatem28, line 72: `toJson` attribute names and types are different than `fromJson`

cw_solana/lib/pending_solana_transaction.dart
```diff
@@ -0,0 +1,38 @@
...
+
+    if (stringifiedAmount.toString().length >= 6) {
+      stringifiedAmount = stringifiedAmount.substring(0, 6);
+    }
+
+    return stringifiedAmount;
+  }
+
+  @override
+  Future<void> commit() async {}
```
OmarHatem28, line 26: this can't be empty, this one should be the function that actually triggers sending/broadcasting the transaction.
I believe you are currently sending the transaction right away, that's why Ramo mentioned that even when the transaction got an error, it broadcasted to the blockchain

OmarHatem28 (changes requested):
I still haven't reviewed the 4 main Solana files, but please fix these comments first and double check Solana main files (Wallet, client, pending transactions, transaction model)

cw_core/lib/crypto_currency.dart
OmarHatem28, line 1: I don't think this is need, please revert it

OmarHatem28 (changes requested):

cw_core/lib/crypto_currency.dart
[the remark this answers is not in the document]
Blazebrain: It's actually needed. We switched the places where we depend on either Erc20Token or SPLToken to directly depend not he parent class CryptoCurrency, removing the multiple casts we were previously doing. 

Because of that, we would not be able to have access to the enabled property, which is only available in the child classes, if we do not add this to the Parent Cryptocurrency class.

cw_solana/lib/solana_transaction_model.dart
```diff
@@ -0,0 +1,50 @@
...
+    required this.from,
+    required this.amount,
+    required this.programId,
+    required int blockTimeInInt,
+    this.isIncomingTransaction = false,
+  }) : blockTime = DateTime.fromMillisecondsSinceEpoch(blockTimeInInt * 1000);
+
+  factory SolanaTransactionModel.fromJson(Map<String, dynamic> json) => SolanaTransactionModel(
+        id: json['id'],
+        blockTimeInInt: int.parse(json["timeStamp"]) * 1000,
```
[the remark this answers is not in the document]
Blazebrain: Was a repetition, removed it. Thank you.

cw_solana/lib/solana_client.dart
```diff
@@ -0,0 +1,391 @@
...
+      throw NoAssociatedTokenAccountException(
+        destinationOwner.toBase58(),
+        mint.toBase58(),
+      );
+    }
+
+    // Input by the user
+    int userAmount = inputAmount.toInt();
+
+    int amount = int.parse('$userAmount${'0' * tokenDecimals}');
```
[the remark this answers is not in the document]
Blazebrain: Yes, this would cause some values to be lost in conversion.

Multiplying the inputAmount by 10 by tokenDecimals won't work for this case, it won't give the lamport equivalent value for this.

However, I can modify the conversion using pow instead to make it cleaner and avoid losing values in double->int conversion.

Something like this
```
Given: userAmount = 2.5 (intending to represent 2.5 units of a token), tokenDecimals = 9.

Direct multiplication: 2.5 * 10 * 9 = 225 (Won't work for this context and won't represent the conversion to the smallest token unit).

Using pow: 2.5 * pow(10, 9) = 2,500,000,000 (which correctly represents 2.5 units of the token in a unit that has 9 decimal places, such as lamports for SOL).
```

cw_solana/lib/solana_transaction_info.dart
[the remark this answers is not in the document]
Blazebrain: Was a previous issue with fees, linked it up now.

lib/view_model/send/output.dart
```diff
@@ -219,10 +218,15 @@ abstract class OutputBase with Store {
...
-      if (cryptoAmount != cryptoAmountTmp) {
-        cryptoAmount = cryptoAmountTmp;
+      //TODO(David): Dive deeper into this and see if the switch can be removed.
+      if (walletType != WalletType.solana) {
+        final cryptoAmountTmp = _cryptoNumberFormat.format(crypto);
+        if (cryptoAmount != cryptoAmountTmp) {
+          cryptoAmount = cryptoAmountTmp;
+        }
+      } else {
+        cryptoAmount = crypto.toStringAsFixed(10);
```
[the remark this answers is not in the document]
Blazebrain: The number formatting always results in the cryptoAmount rounding to zero. That's why Ramo mentioned that the crypto amount doesn't change when you enter the fiat value.

I've added a todo there to check the NumberFormat setup and see why this occurs.

cw_solana/lib/solana_wallet.dart
```diff
@@ -0,0 +1,460 @@
...
+    if (privateKey != null) {
+      final privateKeyBytes = hexToBytes(privateKey);
+      return await Wallet.fromPrivateKeyBytes(privateKey: privateKeyBytes);
+    }
+
+    return Wallet.fromMnemonic(mnemonic!, account: 0, change: 0);
+  }
+
+  @override
+  int calculateEstimatedFee(TransactionPriority priority, int? amount) => 0;
```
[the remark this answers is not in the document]
Blazebrain: This has to do with the transaction priority, and as at last discussion, we mentioned putting it on hold for now because the libraries we use currently do not provide a clear cut way for specifying the priority fee.

cw_solana/lib/solana_transaction_model.dart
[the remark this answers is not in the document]
Blazebrain: Yes, it was intended.

cw_solana/lib/solana_client.dart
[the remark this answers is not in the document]
OmarHatem28: yup, I meant pow, sorry for that confusion

cw_solana/lib/solana_client.dart
```diff
@@ -0,0 +1,406 @@
...
+                  );
+                },
+                splToken: (data) {
+                  data.parsed.map(
+                    transfer: (data) {},
+                    transferChecked: (data) {
+                      final mintAddress = tx.meta?.postTokenBalances.first.mint ?? '';
+                      final tokens = DefaultSPLTokens().initialSPLTokens;
+                      final tokenSymbol =
+                          tokens.firstWhere((token) => token.mintAddress == mintAddress).symbol;
```
OmarHatem28, line 147: if the user has some tokens that are not in Cake or not enabled in Cake, then this would throw an exception and it will result in all transactions not showing, so just wrap this in a try-catch and add it's default value as SOL if we can't get the token name

OmarHatem28 (changes requested):
Missing ChainStack Node changes

lib/core/wallet_connect/web3wallet_service.dart
```diff
@@ -14,11 +14,15 @@ import 'package:cake_wallet/src/screens/wallet_connect/widgets/connection_reques
 import 'package:cake_wallet/src/screens/wallet_connect/widgets/message_display_widget.dart';
 import 'package:cake_wallet/src/screens/wallet_connect/widgets/modals/web3_request_modal.dart';
 import 'package:cake_wallet/store/app_store.dart';
+import 'package:cw_core/wallet_type.dart';
+import 'package:cw_solana/solana_wallet.dart';
```
OmarHatem28, line 18: same

lib/src/screens/send/send_page.dart
```diff
@@ -439,10 +440,17 @@ class SendPage extends BasePage {
                                 }
 
                                 if (state is TransactionCommitted) {
+                                  String alertContent;
+                                  if (solana != null) {
```
OmarHatem28, line 444: this will be true for Cake Wallet app in general, not for solana wallets specifically

cw_solana/lib/solana_client.dart
```diff
@@ -0,0 +1,490 @@
...
+    final mainNetBetaConfig = {"uri": "api.mainnet-beta.solana.com"};
+
+    final mainNetBetaNode = Node.fromMap(mainNetBetaConfig);
+    mainNetBetaNode.type = WalletType.solana;
+
+    final mainNetBetaSolanaClient = SolanaClient(
+      rpcUrl: mainNetBetaNode.uri,
+      websocketUrl: Uri.parse('wss://${mainNetBetaNode.uriRaw}'),
+      timeout: const Duration(minutes: 2),
+    );
```
OmarHatem28, line 128: why are we still using the mainnet-beta node for this? is the ankr node not fetching the balances correctly?

cw_solana/lib/solana_client.dart
OmarHatem28, line 1: **Minor:**
**Enhancement:**
can have some minor generic code enhancements, remove of unused parameters, duplicate code, etc...

OmarHatem28 (changes requested):
also please for future localization, use this script instead
```
flutter packages pub run tool/append_translation.dart "hello_world" "Hello World!"
```

cw_solana/lib/solana_client.dart
Blazebrain: Ankr node works perfectly

I've switched to fully depend on it

lib/src/screens/send/send_page.dart
Blazebrain: Alright

Streamlined it to depend not he wallet type itself

OmarHatem28 (approved):

PR #1281, opened 2024-01-29
Cw 528 backup wallet files

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (changes requested):
good job on saving the backup file after renaming the wallet for Monero, but you missed other wallet types
also, I see no need for awaiting the `saveBackup` function, that will only slow down the wallet opening time, while it can be done asynchronously

and you can now remove the `corrupted-test` related code

OmarHatem28 (approved):

PR #1283, opened 2024-01-30
Cw 384 linux use different default storage location to avoid needing home filesystem permission for flatpak

rafael-xmr:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_core/lib/root_dir.dart
```diff
@@ -5,21 +5,31 @@ String? _rootDirPath;
...
+      } else {
+        final homePath = '/home/${Platform.environment['USER']}/.$appName';
+
+        if (await Directory(homePath).exists()) {
+          appDirPath = homePath;
+        } else {
+          final docPath = await getApplicationDocumentsDirectory();
+
+          if (await docPath.exists()) {
+            appDirPath = docPath.path;
```
OmarHatem28, line 30: shouldn't this be `= '${docPath.path}/$appName';`

cw_polygon/lib/polygon_wallet_service.dart
OmarHatem28, line 1: keep the encryptionFileUtils changes, but remove the `isFlatpak` addition

OmarHatem28 (changes requested):
a lot of unnecessary changes and reformatting
let's find a better approach than this.
at the very least, we can just import the generated `flatpak.dart` file only where it's needed
ex. use it only in `cw_core/lib/pathForWallet.dart` and `cw_core/lib/root_dir.dart`

I would say it would be easier for you to close this PR and make another one, instead of reverting all these changes, but do what you think is better for you

PR #1308, opened 2024-02-21
UI Enhancements

tuxpizza:
- Change spacing for Monero's account widget and actions buttons on the balance page.
- Remove legacyDark theme handler
- Change Cake Wallet's default theme to Dark and Monero.com's to Monero Dark
- Add Red Dark, Red Light, and Purple Dark themes


tuxpizza:
I might need to add a `final defaultTheme = ThemeType.dark.index;` and use the `defaultTheme` var instead of `ThemeType.dark.index` to avoid an error with the null check operator

tuxpizza:
Build failing bc of Solana :cry: 
```
cw_solana/lib/solana_wallet_service.dart:***:***4: Error: The method 'SolanaWalletService.create' has fewer named arguments than those of overridden method 'WalletService.create'.
  Future<SolanaWallet> create(SolanaNewWalletCredentials credentials) async {
                       ^
cw_core/lib/wallet_service.dart:1***:***: Context: This is the overridden method ('create').
  Future<WalletBase> create(N credentials, {bool? isTestnet});
```

lib/store/settings_store.dart
```diff
@@ -789,12 +790,9 @@ abstract class SettingsStoreBase with Store {
     final exchangeStatus = ExchangeApiMode.deserialize(
         raw: sharedPreferences.getInt(PreferencesKey.exchangeStatusKey) ??
             ExchangeApiMode.enabled.raw);
-    final legacyTheme = (sharedPreferences.getBool(PreferencesKey.isDarkThemeLegacy) ?? false)
-        ? ThemeType.dark.index
-        : ThemeType.bright.index;
     final savedTheme = initialTheme ??
         ThemeList.deserialize(
-            raw: sharedPreferences.getInt(PreferencesKey.currentTheme) ?? legacyTheme);
+            raw: sharedPreferences.getInt(PreferencesKey.currentTheme) ?? (isMoneroOnly ? 3 : 2));
```
OmarHatem28, line 795: use theme index better, for readability and maintainability if we change their index at any given time
ex.
`? ThemeType.bright.index : ThemeType.bright.index`

lib/store/settings_store.dart
```diff
@@ -1152,11 +1150,8 @@ abstract class SettingsStoreBase with Store {
     exchangeStatus = ExchangeApiMode.deserialize(
         raw: sharedPreferences.getInt(PreferencesKey.exchangeStatusKey) ??
             ExchangeApiMode.enabled.raw);
-    final legacyTheme = (sharedPreferences.getBool(PreferencesKey.isDarkThemeLegacy) ?? false)
-        ? ThemeType.dark.index
-        : ThemeType.bright.index;
-    currentTheme = ThemeList.deserialize(
-        raw: sharedPreferences.getInt(PreferencesKey.currentTheme) ?? legacyTheme);
+    ThemeList.deserialize(
+        raw: sharedPreferences.getInt(PreferencesKey.currentTheme) ?? (isMoneroOnly ? 3 : 2));
```
OmarHatem28, line 1154: same

lib/themes/theme_base.dart
```diff
@@ -27,7 +27,7 @@ import 'package:cake_wallet/themes/extensions/transaction_trade_theme.dart';
 import 'package:cake_wallet/themes/extensions/wallet_list_theme.dart';
 import 'package:flutter/material.dart';
 
-enum ThemeType { light, bright, dark }
+enum ThemeType { dark, light, bright }
```
OmarHatem28, line 30: I believe changing the order here will cause some issues as there are multiple places that probably depend on the index of those enum values, so since there is no need for them to be changed, please revert them

OmarHatem28 (changes requested):

lib/store/settings_store.dart
tuxpizza: I can't reference the monero orange theme with the ThemeType index. Seems to be for just a type of theme. not one of the specific new themes.

tuxpizza: Managed to do it in a cleaner way by grabbing from ThemeList.[theme].raw instead :smile:

OmarHatem28 (approved):

PR #1314, opened 2024-02-26
Update themeing

tuxpizza:
Changes Purple Dark Theme to Cake Dark Theme and makes that default


tuxpizza:
image
Not sure exactly what happened with the Locale files, I was only trying to add 1 new line. I think its related to Solana

OmarHatem28 (approved):

OmarHatem28:
> image Not sure exactly what happened with the Locale files, I was only trying to add 1 new line. I think its related to Solana

yeah that's expected because the last text addition was not added correctly (not sorted) so it just sorted it out

PR #1339, opened 2024-03-19
Litcoin bitcoin cash fix

OmarHatem28:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/core/address_validator.dart
OmarHatem28, line 1: this will cause some errors if built for Monero.com, as the user might enter a bitcoin address but the `bitcoin` instance is null in Monero.com thus `bitcoin!.` will throw an exception

PR #1347, opened 2024-03-22
Add toggle for disabling service bulletin

tuxpizza:


nahuhh:
Note that this is a privacy leak (calling home), and is specific to cakes node.
as such, it should be opt-in, and really, i don't see a need to call home for this. 

App not syncing reliably? Show a popup and tell user to switch node. < generic connectivity tests. 

Calling home to cake is the same reason monero.com didnt get listen on getmonero.org, and what instigated a 400k hostile takeover attempt of the domain.

calling home is not a feature

tuxpizza:
> Note that this is a privacy leak (calling home), and is specific to cakes node. The info being served has nothing to do with me, my node, or any other remote nodes i may be using.
> 
> its a "service status" that monitors the status if cakes nodes. as such, it should only apply to cake nodes, and opt-in, and at that point.. its overengineering. A solution in search of a problem. i don't see a need to call home no an unrelated server, to do a connectivity check in the node.
> 
> App not syncing reliably? Show a popup and tell user to switch node. = generic connectivity tests that can work for all nodes, and can enable "favorite"/"auto select".
> 
> calling home putting toxic makeup on a pic. Nothing good comes from it.
> 
> its also the same reason monero.com didnt get listen on getmonero.org, and what instigated a 400k hostile takeover attempt of the domain.
> 
> tldr: calling home is not a feature or a solution for connectivity checks

This is a general bulletin board. This isn't JUST for info about cake nodes. There has been various service changes recently and the matter of the fact is that most users who don't know to check the status pages for service information, and will instead open a support ticket for ANY service related issue and support gets flooded very quickly, like when our XMR nodes are down.

We added this as an attempt to notify users in the app of various service changes, like our nodes being down, but also for other things like our fiat API being updated, or being down aswell (which happened recently). This is no different than us providing other first party connections like or fiat API, and isn't just for node status. Another example is recently an exchange provider was down and people were having trouble getting price estimates due to the service taking forever to time out. Stuff like this.

There was an oversight adding this service status feature where there should have already been a way to disable it before we added it, so I am doing that now, as you can see.

Being on by default is same for fiat API and connection to cake node, which are first party services, but can all be disabled before a connection is even made.

nahuhh:
> > Note that this is a privacy leak (calling home), and is specific to cakes node. The info being served has nothing to do with me, my node, or any other remote nodes i may be using.
> > 
> > its a "service status" that monitors the status if cakes nodes. as such, it should only apply to cake nodes, and opt-in, and at that point.. its overengineering. A solution in search of a problem. i don't see a need to call home no an unrelated server, to do a connectivity check in the node.
> > 
> > App not syncing reliably? Show a popup and tell user to switch node. = generic connectivity tests that can work for all nodes, and can enable "favorite"/"auto select".
> > 
> > calling home putting toxic makeup on a pic. Nothing good comes from it.
> > 
> > its also the same reason monero.com didnt get listen on getmonero.org, and what instigated a 400k hostile takeover attempt of the domain.
> > 
> > tldr: calling home is not a feature or a solution for connectivity checks
> 
> This is a general bulletin board. This isn't JUST for info about cake nodes. There has been various service changes recently and the matter of the fact is that most users who don't know to check the status pages for service information, and will instead open a support ticket for ANY service related issue and support gets flooded very quickly, like when our XMR nodes are down.
> 


If im not using your node or fiat api, i dont need a fkn service.cakewallet.com connection

> We added this as an attempt to notify users in the app of various service changes, like our nodes being down, but also for other things like our fiat API being updated, or being down aswell (which happened recently).

Opt in

 > This is no different than us providing other first party connections like or fiat API,

very different. Fiar api is a tool (but without historical pricing, not a very good one).
cake wallets external services status notification on my damn wallet screen is 10000x enough reason to delete it

> and isn't just for node status. Another example is recently an exchange provider was down and people were having trouble getting price estimates due to the service taking forever to time out. 

i have exchanges disabled. Fiat either disabled or tor only. Markets disabled. Even cakepay wouldnt make any connections unless you _used_ a service. Why is cake literally calling home to tell users info they dont care about?

this is totally wrong, tux. 
Even if i was to agree with such a half assed feature, it should be implemented so that it uses the same protocols as the user sets for the service that it it checking. 

User has tor fiat apt + local node? Why is this checking over clearnet? And why is it checking cakes node status?

user has fiat disabled + local node + exchanges over clearnet. 
why is it checking if i dont actually want to makr an exchange?

> There was an oversight adding this service status feature where there should have already been a way to disable it before we added it, so I am doing that now, as you can see.
> 

no, the oversight was being lazy and adding a call home instead of a local, non centralized check

> Being on by default is same for fiat API and connection to cake node, which are first party services,

First party, third party, whatever. Those are both _external_ services. This feature isnt a service, its a tracker of services, and again again again, it can be done PROPERLY, internally, where jt works for all nodes, and smartly for cake services. 

also, is this garbage in the privacy policy?


>> but can all be disabled before a connection is even made.

Thats what i said about why monero.com is still not listed on getmonero. Fiat api was clearnet, and couldnt be disabled.
cake was going to be removed.
simply implementing this trash reverts that decision, and for the very reason it was decided in the first place. You cant parade around as monero.com, and make calls home while being the face of monero. 

have node issues? Did feather add some bs call home to tobby feature? No. He added more community nodes and has feather use them in a smart manner.

1. (now) Add more non-cake nodes (cmon. Yknow how stupid it is to have a centralized point of failure, and to solve it by telling users to scrape google for another option? It's of 0 benefit to cake fod users to use their nodes. This dumb solution doesnt fix anything)

2. (now) add favoriting nodes for autoswitching

3. (now) Add tests for favorite nodes

4. (now) if an external service requires testing, the services themselves should return errors to the app (node dropped connection, fiat api failed to connect) exchange down. And none of them should be polling. 
fkn easy as hell to poll "how many tx in the txpool" and "how long round trip to take this check this time "

tuxpizza:
Well I will once again refer to the node issue, for several days from a week ago, pretty much all the nodes in our list were unusable and simply "adding more" wouldn't necessarily help, most public nodes were struggling to keep up, only private/small ones were working well with the exception of a few. (And we have a lot listed already).

Adding some more nodes is something that can be looked into, but we don't want to just start adding nodes willy-nilly, that is also a security/privacy concern, we only want ones that are run by known good parties, and ones who are accepting of us adding our node to the list in the wallet. And besides that, the issue wasn't just with our nodes specifically, it was also because of monerod due to the P2P traffic.

Connection checks is something we can probably improve, especially once we release our own Tor implementation coming out soon (Since the needed response time for clearnet vs tor is vastly different).  However we have to be very careful with any auto node selection (in the case of an "auto-switch") because our infrastructure is massive and our userbase could easily DDOS a single one of the extra nodes. A popup after said time of stuck connecting or "attempting sync" may be reasonable, with pressing "OK" just simply goes to the node list, OR selects a random node already in the list (would be bad if every user selected the 2nd node on the list if our infrastructure is down again).
Needless to say, this issue is much more nuanced than simply "Adding more nodes to the list".

And of course if an entire network is having problem with slowness or high fees (Like how Solana was literally down for hours a couple months ago) that's entirely out of our control, and this is where something like the bulletin board is helpful. 

Support gets absolutely flooded with tickets because of issues like this which makes it hard to provide good support, and support and user-friendliness is one of the main reasons it makes our app great.

Will add an onion for the bulletin board status also. (Of course if you are using Orbot normally in the the VPN mode it won't go "outside of Tor" but will connect over clearnet through Tor)

With that being said I will formally apologize on behalf of the team, for the addition of this feature without a way to turn it off, as that was a very bad oversight on our part, and I totally understand your frustration with that.

nahuhh:
> Adding some more nodes is something that can be looked into, but we don't want to just start adding nodes willy-nilly, that is also a security/privacy concern, we only want ones that are run by known good parties, 

its a security and 0 conf concern if users are, by default, all using the same node.
also: gui uses bootstrap and simple mode. I dont agree with using randon nodes found in dark alleys, but there are plenty of community nodes to offload to

> and ones who are accepting of us adding our node to the list in the wallet. 

Feather added a few

> And besides that, the issue wasn't just with our nodes specifically, it was also because of monerod due to the P2P traffic.

p2p traffic slows down nodes proportionally, but it is rpc traffic that kills them.
being unable to limit rpc traffic speeds + having a shit-ton of users = dead cake node. 

>  However we have to be very careful with any auto node selection (in the case of an "auto-switch") because our infrastructure is massive and our userbase could easily DDOS a single one of the extra nodes.

Hence, favorites. I wouldnt be autoswitching to cakes nodes.
also, if worried about ddosing, stop pinging _every_ node as soon as you open the node screen. Only test them on demand or under stress if auto/prompt is enabled

> A popup after said time of stuck connecting or "attempting sync" may be reasonable, with pressing "OK" just simply goes to the node list,

aka manual/prompt mode

> OR selects a random node already in the list (would be bad if every user selected the 2nd node on the list if our infrastructure is down again).

again, auto + favorites vs manual prompt
> Needless to say, this issue is much more nuanced than simply "Adding more nodes to the list".

and i nuanced it more than that :)
> And of course if an entire network is having problem with slowness or high fees (Like how Solana was literally down for hours a couple months ago) that's entirely out of our control, and this is where something like the bulletin board is helpful. 

solana is a bad example. They literally turn off their blockchain repeatedly. monero has has 0 downtime and public nodes were under attack.
> Support gets absolutely flooded with tickets because of issues like this which makes it hard to provide good support, and support and user-friendliness is one of the main reasons it makes our app great.

Right. So implement a solution that is better than "try again later" (see: auto select)
> Will add an onion for the bulletin board status also. (Of course if you are using Orbot normally in the the VPN mode it won't go "outside of Tor" but will connect over clearnet through Tor)

> With that being said I will formally apologize on behalf of the team, for the addition of this feature without a way to turn it off, as that was a very bad oversight on our part, and I totally understand your frustration with that.

OmarHatem28 (approved):

nahuhh:
@OmarHatem28  remove this feature or ill ensure cake is ptomptly removed from getmonero 💯
fkn spyware

tuxpizza:
Not happening.
Not spyware. 
Btw it hits the EXACT same server that FIAT API is running on, using a similar API just through a different DNS request.

nahuhh:
> Not happening.
> Not spyware. 
> Btw it hits the EXACT same server that FIAT API is running on, using a similar API just through a different DNS request.
> 
> 4.15.3 you can disable it.
👌 bye

PR #1359, opened 2024-04-03
fixup

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/buy/dfx/dfx_buy_provider.dart
+++ b/lib/buy/dfx/dfx_buy_provider.dart
@@ -70,6 +70,24 @@ class DFXBuyProvider extends BuyProvider {
     }
   }
 
+  @override
+  List<WalletType> getSupportedWalletTypes(bool isBuy) {
+    if (isBuy) {
+      return [
+        WalletType.bitcoin,
+        WalletType.ethereum,
+        WalletType.polygon,
+        WalletType.monero,
+        WalletType.nano,
+      ];
+    }
+    return [
+      WalletType.bitcoin,
+      WalletType.ethereum,
+      WalletType.polygon,
+    ];
+  }
+
   String get walletAddress =>
       wallet.walletAddresses.primaryAddress ?? wallet.walletAddresses.address;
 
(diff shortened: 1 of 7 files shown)


lib/entities/provider_types.dart
```diff
@@ -48,61 +48,27 @@ extension ProviderTypeName on ProviderType {
...
-        return [ProviderType.askEachTime, ProviderType.onramper, ProviderType.robinhood, ProviderType.moonpay];
-      case WalletType.solana:
-        return [ProviderType.askEachTime, ProviderType.onramper, ProviderType.robinhood];
-      case WalletType.none:
-      case WalletType.haven:
-        return [];
+    final providers = <ProviderType>[];
+    for (final providerType in ProviderType.values) {
+      final dynamic p = getProviderTypeByType(providerType);
+      final supportedWalletTypes = p.getSupportedWalletTypes(true) as List<WalletType>;
```
OmarHatem28, line 54: Why dynamic, make it `BuyProvider?` thus it will give you an error when directly calling getSupportedWalletTypes as p might be null
Modify it elsewhere as well

OmarHatem28 (changes requested):

lib/entities/provider_types.dart
fossephate: Because it's not actually a BuyProvider instance that we're passing (because we wanted to avoid initializing the BuyProvider)

It's the type which inherits BuyProvider that we're passing

Afaik there's no way to do this properly without dynamic :/

lib/buy/dfx/dfx_buy_provider.dart
```diff
@@ -70,6 +70,24 @@ class DFXBuyProvider extends BuyProvider {
...
 
+  @override
+  List<WalletType> getSupportedWalletTypes(bool isBuy) {
+    if (isBuy) {
+      return [
+        WalletType.bitcoin,
+        WalletType.ethereum,
+        WalletType.polygon,
+        WalletType.monero,
+        WalletType.nano,
```
OmarHatem28, line 81: DFX doesn't support nano, can you please make sure that you have migrated the correct wallet types to each provider

PR #1361, opened 2024-04-04
CW-589-WalletConnect-Connections-Independent

Blazebrain:

# Description
Makes WalletConnect connections independent to the wallet through which the connection was created

# Pull Request - Checklist  

- [x] Format code


lib/core/wallet_connect/chain_service/solana/solana_chain_service.dart
```diff
@@ -72,7 +72,7 @@ class SolanaChainServiceImpl implements ChainService {
 
   @override
   List<String> getEvents() {
-    return [''];
+    return ['chainChanged', 'accountsChanged'];
```
OmarHatem28, line 75: what are those?

OmarHatem28 (commented):
fix conflicts by just running the translation command again with the text in the en file and it will automatically sort it

lib/core/wallet_connect/chain_service/solana/solana_chain_service.dart
Blazebrain: Some events that the Client listens to, similar to the eth_chain_service own also.

OmarHatem28 (approved):

PR #1367, opened 2024-04-05
version bump to 3.13.9, auth working on mac

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/src/widgets/validable_annotated_editable_text.dart
OmarHatem28, line 1: why is this change needed?

lib/locales/hausa_intl.dart
OmarHatem28, line 1: replace the `throw UnimplementedError();` with a dummy text so we make sure we avoid unintended behaviors

lib/entities/biometric_auth.dart
```diff
@@ -1,32 +1,32 @@
...
     try {
-      return await _localAuth.canCheckBiometrics;
-    } on PlatformException catch (e) {
-      print(e);
+      canAuthenticate = await _flutterLocalAuthenticationPlugin.canAuthenticate();
+
+      // Setup TouchID Allowable Reuse duration
+      // It works only in iOS and macOS, but it's safe to call it even on other platforms.
+      await _flutterLocalAuthenticationPlugin.setTouchIDAuthenticationAllowableReuseDuration(30);
+    } on Exception catch (error) {
```
OmarHatem28, line 25: I believe this is not needed, and you can also remove the specific `on Exception` so that it catches any error even if Dart considers the error not a descendant from the `Exception` parent class

OmarHatem28 (changes requested):

lib/src/widgets/validable_annotated_editable_text.dart
fossephate: huh, so I remember making this change because I was getting type errors complaining that TextAnnotation couldn't extend `Comparable` and I think I assumed that they just changed Comparable in the new flutter version

I can't seem to reproduce whatever type error I was getting so it must have just been my IDE clashing with the different dart versions or something weird like that

I'll remove this and make sure it works without it and get back to you

lib/locales/hausa_intl.dart
fossephate: why do these languages need their own files / fixes anyways? are they just not supported by the localization package we use?

lib/src/widgets/validable_annotated_editable_text.dart
fossephate: ok so I got the error:
```
The class 'Comparable' can't be extended outside of its library because it's an interface class.dartinvalid_use_of_type_outside_library
class Annotation extends Comparable<Annotation>
package:cake_wallet/src/widgets/validable_annotated_editable_text.dart
```

fossephate: it looks like in the more recent versions of dart you can't extend interfaces from outside of a library for whatever reason
my quick-solution was to just duplicate the class since it's relatively simple

fossephate: looking at this again though, I think my approach wouldn't work:
https://stackoverflow.com/questions/76224247/how-to-use-comparable-correctly-with-dart-3

fossephate: latest commit updates the class to use the approach in the stack overflow post linked^

lib/locales/hausa_intl.dart
OmarHatem28: yup, those are not supported by flutter

OmarHatem28 (approved):

PR #1374, opened 2024-04-09
Autogen address

tuxpizza:


tuxpizza:
Added a missing commit to always show the Plus button on the addresses page

OmarHatem28:
not sure if it's going to be that simple, at least we will need to test whether the auto-generate will work as expected even if the user manually created a few addresses

tuxpizza:
I will do some tests, if it turns out to be a problem we can have it disabled for now.

PR #1384, opened 2024-04-13
Better handle corrupted wallets

OmarHatem28:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/core/wallet_loading_service.dart
```diff
@@ -38,15 +39,43 @@ class WalletLoadingService {
...
+          final password = await keyService.getWalletPassword(walletName: walletInfo.name);
+          final wallet = await walletService.openWallet(walletInfo.name, password);
+
+          if (walletInfo.type == WalletType.monero) {
+            await updateMoneroWalletPassword(wallet);
+          }
+
+          await sharedPreferences.setString(PreferencesKey.currentWalletName, wallet.name);
+          await sharedPreferences.setInt(
+              PreferencesKey.currentWalletType, serializeToInt(wallet.type));
```
konstantinullrich, line 70: Is setting the wallet as current here necessary? It could have unintended side-effects during background sync. All the other calls to the `WalletLoadingService.load` method run `AppStore.changeCurrentWallet` which sets these shared preferences using a mobx-reaction defined in `lib/reactions/on_current_wallet_change.dart`.

konstantinullrich (changes requested):
Looks good, but please look at the minor question regarding the change current wallet.

lib/core/wallet_loading_service.dart
OmarHatem28: if I understand your concern correctly, then it should have no side-effect, as it will return the wallet that is not corrupted anyway, so `AppStore.changeCurrentWallet` will be called with the not corrupted wallet, so it will just re-assign those preferences values again with the same values, I just added it to be on the safe side.

feel free to test it tho.

I tried testing it by specifically making it throw an exception when loading the current wallet while having multiple wallets, and it successfully switched to the second wallet I had

konstantinullrich (approved):

PR #1387, opened 2024-04-15
CW-615-Add-Check-For-Contract-Address

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (changes requested):
ok this won't work

this check is only valid if the current wallet is either ethereum or polygon, which doesn't make sense and I am not sure how you tested it, because if a user is trying to make an exchange TO ethereum, he will definitely be on another wallet

so to elaborate more, if I want to exchange BTC to ETH contract address, then I will be on a BTC wallet not an ethereum wallet, thus your check wouldn't be valid

and this introduces another issue which is that we would need an Ethereum node so we can get web3dart to work, thus for Monero.com App this won't be possible.

so let's do the following, try to find another solution away from web3dart, maybe something that uses direct API calls from etherscan/polyscan or morallis or even another API no problem

if nothing is found, then enhance the current implementation to not be with current wallet, and I see no need for this, 
`getReceivingAddressFromCredentials`
you can get the address directly from the outputs, no need for an extra call to the proxy layer
```Dart
final receivingAddress =
          outputs.first.isParsedAddress ? outputs.first.extractedAddress : outputs.first.address;
```

Blazebrain:
> ok this won't work
> 
> this check is only valid if the current wallet is either ethereum or polygon, which doesn't make sense and I am not sure how you tested it, because if a user is trying to make an exchange TO ethereum, he will definitely be on another wallet
> 
> so to elaborate more, if I want to exchange BTC to ETH contract address, then I will be on a BTC wallet not an ethereum wallet, thus your check wouldn't be valid
> 
> and this introduces another issue which is that we would need an Ethereum node so we can get web3dart to work, thus for Monero.com App this won't be possible.
> 
> so let's do the following, try to find another solution away from web3dart, maybe something that uses direct API calls from etherscan/polyscan or morallis or even another API no problem
> 
> if nothing is found, then enhance the current implementation to not be with current wallet, and I see no need for this, `getReceivingAddressFromCredentials` you can get the address directly from the outputs, no need for an extra call to the proxy layer
> 
> ```dart
> final receivingAddress =
>           outputs.first.isParsedAddress ? outputs.first.extractedAddress : outputs.first.address;
> ```

Alright.

I was previously under the impression of it being specific to the evm wallets, but your explanation clears that up perfectly. 

For the issue with Monero.com App, I'll check out direct APIs we can use as anything specific would affect either of the apps.

lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -776,16 +782,70 @@ abstract class ExchangeViewModelBase extends WalletChangeListenerViewModel with
...
       if (tapRootPattern.hasMatch(payoutAddress) || tapRootPattern.hasMatch(fromWalletAddress)) {
-        return false;
+        return CreateTradeResult(
+          result: false,
+          errorMessage: S.current.thorchain_taproot_address_not_supported,
+        );
+      }
+
+      final currenciesToCheckPattern = RegExp('0x[0-9a-zA-Z]');
+      final inputAddress = trade.inputAddress ?? '';
```
OmarHatem28, line 799: this checks for the address that is owned by the exchange provider, which the user will send the funds to, to make the exchange, which is not what we want to check

we wanna check the user sending address, which would be `trade.fromWalletAddress` and also check the receiving address which would be `trade.payoutAddress` like the if condition for taproot above

how did you verify that this is working? the only way I see this being verified is that you manually changed the `trade.inputAddress` which should be returned from the exchange provider API, by then you should already know that this is not correct

lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -776,16 +782,70 @@ abstract class ExchangeViewModelBase extends WalletChangeListenerViewModel with
...
+          errorMessage: !isEOAAddress ? S.current.thorchain_contract_address_not_supported : null,
+        );
       }
     }
-    return true;
+    return CreateTradeResult(result: true);
+  }
+
+  Future<bool> _isExternallyOwnedAccountAddress(String receivingAddress) async {
+    final isEOAAddress = !(await _isContractAddress('eth', receivingAddress));
```
OmarHatem28, line 813: is hardcoding `eth` here right? it might be polygon (matic) or other L2 chains when we integrate them, why not just use the `to` cryptoCurrency's tag, which should be correct already and we don't need to hardcode it

OmarHatem28 (changes requested):

OmarHatem28 (approved):

PR #1390, opened 2024-04-16
thor name to address lookup

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/lib/entities/parse_address_from_domain.dart
+++ b/lib/entities/parse_address_from_domain.dart
@@ -5,6 +5,7 @@ import 'package:cake_wallet/entities/openalias_record.dart';
 import 'package:cake_wallet/entities/parsed_address.dart';
 import 'package:cake_wallet/entities/unstoppable_domain_address.dart';
 import 'package:cake_wallet/entities/emoji_string_extension.dart';
+import 'package:cake_wallet/exchange/provider/thorchain_exchange.provider.dart';
 import 'package:cake_wallet/mastodon/mastodon_api.dart';
 import 'package:cake_wallet/nostr/nostr_api.dart';
 import 'package:cake_wallet/store/settings_store.dart';
@@ -160,6 +161,15 @@ class AddressResolver {
           }
         }
       }
+
+      final thorChainAddress = await ThorChainExchangeProvider.lookupAddressByName(text);
+      if (thorChainAddress != null) {
+        String? address = thorChainAddress[ticker.toUpperCase()];
+        if (address != null) {
+          return ParsedAddress.thorChainAddress(address: address, name: text);
+        }
+      }
+
       final formattedName = OpenaliasRecord.formatDomainName(text);
       final domainParts = formattedName.split('.');
       final name = domainParts.last;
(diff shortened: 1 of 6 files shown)


lib/src/screens/send/widgets/extract_address_from_parsed.dart
```diff
@@ -56,6 +56,13 @@ Future<String> extractAddressFromParsed(
       profileImageUrl = parsedAddress.profileImageUrl;
       profileName = parsedAddress.profileName;
       break;
+    case ParseFrom.thorChain:
+      title = S.of(context).address_detected;
+      content = S.of(context).extracted_address_content('${parsedAddress.name} (ThorChain)');
+      address = parsedAddress.addresses.first;
+      profileImageUrl = parsedAddress.profileImageUrl;
+      profileName = parsedAddress.profileName;
```
OmarHatem28, line 64: those 2 are not valid for thorchain and will throw an exception most probably if they are not handled correctly in the app

lib/exchange/provider/thorchain_exchange.provider.dart
```diff
@@ -206,8 +208,35 @@ class ThorChainExchangeProvider extends ExchangeProvider {
     );
   }
 
+  static Future<Map<String, String>?>? lookupAddressByName(String name) async {
+    final uri = Uri.https(_baseURL, '$_nameLookUpPath$name');
+    final response = await http.get(uri);
+
+    if (response.statusCode != 200) {
+      throw Exception('Unexpected HTTP status: ${response.statusCode}');
```
OmarHatem28, line 216: just return null so it continues checking other address resolvers

lib/entities/parse_address_from_domain.dart
```diff
@@ -160,6 +161,15 @@ class AddressResolver {
           }
         }
       }
+
+      final thorChainAddress = await ThorChainExchangeProvider.lookupAddressByName(text);
+      if (thorChainAddress != null) {
+        String? address = thorChainAddress[ticker.toUpperCase()];
```
OmarHatem28, line 167: the ticker would be `rune` but the returned value from the API will be of chain `THOR` 

https://midgard.ninerealms.com/v2/thorname/lookup/cakewallet

let's add a specific case for this then, since it will probably be just the case for Thor

OmarHatem28 (changes requested):

lib/src/screens/send/widgets/extract_address_from_parsed.dart
Serhii-Borodenko: This case is handled in the app, but it is indeed unnecessary in the case of ThorChain.

lib/exchange/provider/thorchain_exchange.provider.dart
Serhii-Borodenko: done

lib/entities/parse_address_from_domain.dart
Serhii-Borodenko: Added case for the 'rune' coin.

OmarHatem28 (approved):

PR #1398, opened 2024-04-18
bio auth on mac + package updates for 3.19.3/5

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/view_model/auth_view_model.dart
```diff
@@ -105,16 +106,14 @@ abstract class AuthViewModelBase with Store {
...
-
-      if (canBiometricAuth) {
-        final isAuthenticated = await _biometricAuth.isAuthenticated();
+    final _flutterLocalAuthenticationPlugin = FlutterLocalAuthentication();
 
-        if (isAuthenticated) {
-          state = ExecutedSuccessfullyState();
-        }
+    try {
+      final authenticated = await _flutterLocalAuthenticationPlugin.authenticate();
```
OmarHatem28, line 112: you can keep using `_biometricAuth` instance, since you already modified the code there to use the new package, no need to import it here as well and use it separately

OmarHatem28 (changes requested):
ok looks, good, just 1 comment, and the fix for the biometric auth getting disabled after a restart on Mac

lib/view_model/edit_backup_password_view_model.dart
```diff
@@ -37,8 +38,6 @@ abstract class EditBackupPasswordViewModelBase with Store {
   @action
   Future<void> save() async {
     final key = generateStoreKeyFor(key: SecretStoreKey.backupPassword);
-    await secureStorage.delete(key: key);
-    await secureStorage.write(key: key, value: backupPassword);
-    secretStore.write(key: key, value: backupPassword);
```
OmarHatem28, line 42: this shouldn't be deleted, and please double check if any side-effect like this was made unintentionally through the changes

OmarHatem28 (changes requested):
and please test it with backward compatibility as well
i.e have the app already running with some wallets, and then switch to this branch and build, and see if it will cause any issues or not

OmarHatem28 (approved):

PR #1406, opened 2024-04-22
Fix estimated fee calculation for customs fee rate

Serhii-Borodenko:

# Description

Fix estimated fee calculation for customs fee rate

# Pull Request - Checklist  

- [x] Format code


lib/view_model/send/output.dart
```diff
@@ -126,8 +126,8 @@ abstract class OutputBase with Store {
 
       if (_wallet.type == WalletType.bitcoin) {
         if (_settingsStore.priority[_wallet.type] == bitcoin!.getBitcoinTransactionPriorityCustom()) {
-          fee = bitcoin!.getFeeAmountWithFeeRate(
-              _settingsStore.customBitcoinFeeRate, formattedCryptoAmount, 1, 1);
+          fee = bitcoin!.getFeeAmountWithFeeRate(_wallet,
+              _settingsStore.customBitcoinFeeRate, 1, 0);
```
OmarHatem28, line 130: shouldn't the output count be 1 so it's calculated successfully?

also are we calculating the estimated fee based on a single output? what if we have multiple outputs

OmarHatem28 (changes requested):

lib/view_model/send/output.dart
Serhii-Borodenko: I've updated the code, now the custom fee rate calculation aligns with the default priority flow. It correctly calculates inputsCount based on the amount in both cases. outputsCount depends on this condition: (amount != null ? 2 : 1).

OmarHatem28 (approved):

PR #1429, opened 2024-05-03
bitcoin restore bug fix and move restore code into view model

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/bitcoin/cw_bitcoin.dart
+++ b/lib/bitcoin/cw_bitcoin.dart
@@ -287,6 +287,10 @@ class CWBitcoin extends Bitcoin {
   }) async {
     List<DerivationInfo> list = [];
 
+    if (!validateMnemonic(mnemonic) && !bip39.validateMnemonic(mnemonic)) {
+      throw BitcoinMnemonicIsIncorrectException();
+    }
+
     List<DerivationType> types = await compareDerivationMethods(mnemonic: mnemonic, node: node);
     if (types.length == 1 && types.first == DerivationType.electrum) {
       return [
--- a/lib/view_model/wallet_restore_view_model.dart
+++ b/lib/view_model/wallet_restore_view_model.dart
@@ -1,12 +1,15 @@
 import 'package:cake_wallet/bitcoin/bitcoin.dart';
+import 'package:cake_wallet/core/execution_state.dart';
 import 'package:cake_wallet/di.dart';
 import 'package:cake_wallet/nano/nano.dart';
 import 'package:cake_wallet/ethereum/ethereum.dart';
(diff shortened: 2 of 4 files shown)


lib/bitcoin/cw_bitcoin.dart
```diff
@@ -287,6 +287,10 @@ class CWBitcoin extends Bitcoin {
   }) async {
     List<DerivationInfo> list = [];
 
+    if (!validateMnemonic(mnemonic) && !bip39.validateMnemonic(mnemonic)) {
+      throw BitcoinMnemonicIsIncorrectException();
+    }
```
OmarHatem28, line 292: did you try reproducing it again after fixing it?
cuz there is no try/catch block all the way to the UI that will catch this exception

lib/view_model/wallet_restore_view_model.dart
```diff
@@ -238,6 +241,54 @@ abstract class WalletRestoreViewModelBase extends WalletCreationVM with Store {
...
+        }
+      }
+
+      if (dInfo == null) {
+        dInfo = getDefaultDerivation();
+      }
+      return dInfo;
+    } catch (e) {
+      state = FailureState(e.toString());
+      rethrow;
```
OmarHatem28, line 288: same

OmarHatem28 (changes requested):
why instead of adding a validation to the seed field you go all the way to throwing an exception when fetching the derivation paths
why not handle it on the UI layer directly

lib/bitcoin/cw_bitcoin.dart
fossephate: yeah the fix definitely works

the main reason it's implemented this way is to match the behavior of the other wallet types but I'm also trying to minimize code duplication here
I think most of the time you're right doing form validation in the UI layer makes sense, but since we already validate (or should be if we aren't) mnemonics / seeds in the restoreFromSeed() / restoreFromKeys() functions in each wallet type, I don't see an issue with having that error be passed through to the ui

OmarHatem28: nope, my point is, if you try to reproduce this issue, you will get the `oops we got an error` popup and send a mail to support, while we don't need that, we just need to inform the user that the seeds are incorrect

fossephate: oooh that makes sense

OmarHatem28:
@fossephate since the comments have not been resolved yet and it currently has conflicts equal to the files changed anyway, so can you create a new branch based on the latest main and make these changes with the review comments in mind and test it again?
close this PR afterward and open a new one

PR #1436, opened 2024-05-08
Fix evm exchanges via thorchain

OmarHatem28:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_evm/lib/evm_chain_wallet.dart
```diff
@@ -245,12 +245,11 @@ abstract class EVMChainWalletBase
...
 
-    final String? opReturnMemo = outputs.first.memo;
+    final String? memo = outputs.first.memo;
 
     String? hexOpReturnMemo;
-    if (opReturnMemo != null) {
-      hexOpReturnMemo =
-          '0x${opReturnMemo.codeUnits.map((char) => char.toRadixString(16).padLeft(2, '0')).join()}';
+    if (memo != null) {
+      hexOpReturnMemo = '0x${memo.codeUnits.map((char) => char.toRadixString(16).padLeft(2, '0')).join()}';
```
OmarHatem28, line 252: I don't think that this is needed anymore.
isn't this the previous implementation when we thought the memo would be sent in the `data` field, but now we can skip passing the `data` parameter at all

cw_evm/lib/evm_chain_wallet.dart
```diff
@@ -323,6 +322,8 @@ abstract class EVMChainWalletBase
       contractAddress:
           transactionCurrency is Erc20Token ? transactionCurrency.contractAddress : null,
       data: hexOpReturnMemo,
```
OmarHatem28, line 324: i.e remove it from here as well, it shouldn't be needed

lib/exchange/provider/thorchain_exchange.provider.dart
```diff
@@ -72,10 +77,15 @@ class ThorChainExchangeProvider extends ExchangeProvider {
     try {
       if (amount == 0) return 0.0;
 
+      final formattedAmount =
+          (from == CryptoCurrency.usdterc20 || from == CryptoCurrency.usdc) && amount < 200
+              ? 200.00
+              : amount;
```
OmarHatem28, line 83: I would rather it fail, than fetching a wrong rate.
so for example, say the user entered 50 USDT, and ThorChain returned a better rate than ChangeNow, because it will calculate on 200 instead of 50, but when creating it will fail, so why, just let it fail and it will return 0

lib/view_model/send/send_view_model.dart
```diff
@@ -126,9 +126,14 @@ abstract class SendViewModelBase extends WalletChangeListenerViewModel with Stor
...
 
+    final pricesKeys = fiatConversionStore.prices.keys.toList();
+
+    final selectedCurr = pricesKeys.firstWhereOrNull((element) =>
+    element.title == selectedCryptoCurrency.title && element.tag == selectedCryptoCurrency.tag);
+
     try {
       final fiat = calculateFiatAmount(
-          price: _fiatConversationStore.prices[selectedCryptoCurrency]!,
+          price: fiatConversionStore.prices[selectedCurr]!,
```
OmarHatem28, line 136: what's the difference?!
why not directly use `selectedCryptoCurrency`

Serhii-Borodenko: This issue needs further investigation because the condition fiatConversionStore.prices.containsKey(selectedCryptoCurrency) is true when sending ERC20 tokens but false during the exchange flow for ERC20 tokens only. In both cases, an instance of Currency is in the prices map.

cw_evm/lib/evm_chain_client.dart
```diff
@@ -97,20 +99,39 @@ abstract class EVMChainClient {
...
+    final int gasLimit = estimatedGasLimit.toInt() < 80000 ? 80000 : estimatedGasLimit.toInt();
 
     final Transaction transaction = createTransaction(
       from: privateKey.address,
       to: EthereumAddress.fromHex(toAddress),
       maxPriorityFeePerGas: EtherAmount.fromInt(EtherUnit.gwei, priority.tip),
       amount: isNativeToken ? EtherAmount.inWei(amount) : EtherAmount.zero(),
       data: data != null ? hexToBytes(data) : null,
+      maxGas: router != null && memo != null && isNativeToken ? gasLimit : null,
+      gasPrice: router != null && memo != null && isNativeToken ? EtherAmount.inWei(adjustedGasPrice) : null,
```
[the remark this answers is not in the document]
Serhii-Borodenko: The conditions router != null and memo != null indicate that it is a Thorchain swap, and isNativeToken is ETH.

cw_evm/lib/evm_chain_client.dart
```diff
@@ -97,20 +99,39 @@ abstract class EVMChainClient {
 
     bool isNativeToken = currency == CryptoCurrency.eth || currency == CryptoCurrency.maticpoly;
 
-    final price = _client!.getGasPrice();
+    final gasPrice = await _client!.getGasPrice();
+    final adjustedGasPrice = gasPrice.getInWei * BigInt.from(15) ~/ BigInt.from(10);
```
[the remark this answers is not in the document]
Serhii-Borodenko: I increased the gas price by multiplying it by 1.5 to speed up the transaction.

cw_evm/lib/evm_chain_client.dart
```diff
@@ -280,22 +318,147 @@ abstract class EVMChainClient {
...
+        credentials: privateKey,
+      );
+
+      final txHash = await _client!.sendRawTransaction(approvalTransaction);
+
+      TransactionReceipt? receipt;
+      while (receipt == null) {
+        receipt = await _client!.getTransactionReceipt(txHash);
+        await Future.delayed(const Duration(seconds: 1));
+      }
```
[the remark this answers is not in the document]
Serhii-Borodenko: done

cw_evm/lib/evm_chain_client.dart
```diff
@@ -280,22 +318,147 @@ abstract class EVMChainClient {
...
+
+      if (receipt.status == false) {
+        log('Approval transaction failed');
+        throw Exception('Approval transaction failed');
+      }
+    }
+
+    final currentAllowanceNew =
+        await erc20.allowance(privateKey.address, EthereumAddress.fromHex(router));
+    log('New Allowance: $currentAllowanceNew');
```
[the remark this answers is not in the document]
Serhii-Borodenko: removed

cw_evm/lib/evm_chain_wallet.dart
Serhii-Borodenko: Swaps for ETH don't work without the memo/data parameter. Transactions don't arrive at the destination address, so I had to include the data parameter again.

cw_evm/lib/evm_chain_wallet.dart
Serhii-Borodenko: Swaps for ETH don't work without the memo/data parameter.

lib/exchange/provider/thorchain_exchange.provider.dart
Serhii-Borodenko: I got the point, it's useful only for fetchLimits and can be removed here.

cw_evm/lib/evm_chain_client.dart
```diff
@@ -280,22 +318,147 @@ abstract class EVMChainClient {
...
+      final feeHistory = await _client!.getFeeHistory(
+        1,
+        rewardPercentiles: [],
+      );
+
+      final baseFeePerGas = feeHistory['baseFeePerGas'].first as BigInt;
+
+      final effectiveGasPrice = gasPrice.getInWei > baseFeePerGas
+          ? gasPrice
+          : EtherAmount.inWei(baseFeePerGas * BigInt.from(2));
```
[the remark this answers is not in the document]
Serhii-Borodenko: to ensure that the transaction has a high enough effective gas price to be prioritized by the network

Blazebrain (commented):
Dropped a couple of follow up comments


Well done so far

PR #1449, opened 2024-05-13
electrum updates

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code


konstantinullrich (approved):

lib/view_model/wallet_creation_vm.dart
```diff
@@ -1,3 +1,4 @@
+import 'package:cake_wallet/bitcoin/bitcoin.dart';
```
OmarHatem28, line 1: this will have an issue when building Monero.com

cw_bitcoin/lib/electrum_derivations.dart
```diff
@@ -4,7 +4,7 @@ Map<DerivationType, List<DerivationInfo>> electrum_derivations = {
   DerivationType.electrum: [
     DerivationInfo(
       derivationType: DerivationType.electrum,
-      derivationPath: "m/0'",
+      derivationPath: "m/0'/0",
```
OmarHatem28, line 7: why not add the un-hardened 0 to the other paths as well?

OmarHatem28 (changes requested):

cw_bitcoin/lib/electrum_derivations.dart
fossephate: I'm not sure what you mean? I did?

OmarHatem28: for the other derivations some of them are without the un-hardened 0

lib/entities/default_settings_migration.dart
```diff
@@ -766,6 +769,19 @@ Future<void> changeDefaultMoneroNode(
...
+Future<void> fixBtcDerivationPaths(Box<WalletInfo> walletsInfoSource) async {
+  for (WalletInfo walletInfo in walletsInfoSource.values) {
+    if (walletInfo.type == WalletType.bitcoin ||
+        walletInfo.type == WalletType.bitcoinCash ||
+        walletInfo.type == WalletType.litecoin) {
+      if (walletInfo.derivationInfo?.derivationPath == "m/0'/0") {
+        walletInfo.derivationInfo!.derivationPath = "m/0'";
+      }
+    }
+    await walletInfo.save();
```
OmarHatem28, line 781: **Enhancement**
no need to call save for every wallet, while we only need to call it if we change the derivation path
this code is executed on app start, so it might slow the start of the app at that one time for no reason

OmarHatem28 (changes requested):

OmarHatem28 (approved):

PR #1460, opened 2024-05-22
CW-636-1-Unfocus text fields when push to next one

Serhii-Borodenko:

# Description

Implemented automatic unfocus on the current text field across various screens (Exchange, Send, Address Book, QR, Paste, etc.) to prevent the keyboard from popping up.

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/lib/src/screens/exchange/exchange_page.dart
+++ b/lib/src/screens/exchange/exchange_page.dart
@@ -100,6 +100,14 @@ class ExchangePage extends BasePage {
   AppBarStyle get appBarStyle => AppBarStyle.transparent;
 
   @override
+  Function(BuildContext)? get pushToNextWidget => (context) {
+    FocusScopeNode currentFocus = FocusScope.of(context);
+    if (!currentFocus.hasPrimaryFocus) {
+      currentFocus.focusedChild?.unfocus();
+    }
+  };
+
+  @override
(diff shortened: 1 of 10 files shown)


lib/src/screens/exchange/exchange_page.dart
```diff
@@ -99,6 +99,14 @@ class ExchangePage extends BasePage {
   @override
   AppBarStyle get appBarStyle => AppBarStyle.transparent;
 
+  @override
+  Function(BuildContext)? get pushToNextWidget => (context) {
+    FocusScopeNode currentFocus = FocusScope.of(context);
+    if (!currentFocus.hasPrimaryFocus) {
+      currentFocus.focusedChild?.unfocus();
+    }
```
OmarHatem28, line 107: wouldn't this line be enough?
```Dart
FocusScope.of(context).unfocus();
```

OmarHatem28 (changes requested):
also can you make this for wallet creation/restore (from seeds and backup)

lib/src/screens/exchange/exchange_page.dart
Serhii-Borodenko: No, this doesn't work in our case, most likely because we are using custom focusNodes.

OmarHatem28 (approved):

PR #1471, opened 2024-05-30
Sp fixes

rafael-xmr:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/src/widgets/setting_actions.dart
+++ b/lib/src/widgets/setting_actions.dart
@@ -29,6 +29,7 @@ class SettingActions {
     connectionSettingAction,
     walletSettingAction,
     addressBookSettingAction,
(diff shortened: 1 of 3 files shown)


lib/src/widgets/setting_actions.dart
```diff
@@ -29,6 +29,7 @@ class SettingActions {
     connectionSettingAction,
     walletSettingAction,
     addressBookSettingAction,
+    silentPaymentsSettingAction,
```
OmarHatem28, line 32: need to also add the check for whether this wallet supports Silent Payments or not in the page that uses this list

OmarHatem28 (changes requested):

PR #1485, opened 2024-06-08
Wownero

OmarHatem28:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/entities/seed_type.dart
```diff
@@ -10,13 +10,16 @@ class SeedType extends EnumerableItem<int> with Serializable<int> {
 
   static const legacy = SeedType(raw: 0, title: 'Legacy (25 words)');
   static const polyseed = SeedType(raw: 1, title: 'Polyseed (16 words)');
+  static const wowneroSeed = SeedType(raw: 1, title: 'Wownero (14 words)');
```
OmarHatem28, line 13: raw: 2

lib/src/screens/restore/wallet_restore_from_seed_form.dart
```diff
@@ -193,7 +209,9 @@ class WalletRestoreFromSeedFormState extends State<WalletRestoreFromSeedForm> {
                 ),
               ),
             ),
-          if (!isPolyseed && widget.displayBlockHeightSelector)
+          if ((!isPolyseed) &&
+              widget.displayBlockHeightSelector &&
+              (widget.type == WalletType.wownero && !seedTypeController.value.text.contains("14")))
```
OmarHatem28, line 214: I believe this is not correct

MrCyjaneK: wownero-seed acts like polyseed, and it stores the restore height as well

OmarHatem28: ok, but with this if condition, it has to be a wownero wallet and not a 14 word so that it shows the restore height widget, so if I am on a Monero wallet this condition would return false

nahuhh:
Missing `settings > privacy > auto generate subaddresses`

nahuhh:
Small bug

if setting up a new wallet but the selected node is unreachable, the wallet will freeze and refuse to open until the node comes online. 

only on a clean install.

also, rpc login field are missing for wow's advanced setup

PR #1534, opened 2024-07-15
update monero_c to fix unreachable wownero git hosting

MrCyjaneK:
Bumped the commit because wownero is unreachable for quite some time, and I really don't want to spend any more time copying already cloned repositories when wownero git is offline.

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/view_model/transaction_details_view_model.dart
+++ b/lib/view_model/transaction_details_view_model.dart
@@ -214,7 +214,7 @@ abstract class TransactionDetailsViewModelBase with Store {
     final addressIndex = tx.additionalInfo['addressIndex'] as int;
     final feeFormatted = tx.feeFormatted();
     final _items = [
-      StandartListItem(title: S.current.transaction_details_transaction_id, value: tx.id),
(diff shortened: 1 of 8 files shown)


OmarHatem28 (approved):

OmarHatem28:
looks good, but there are a few things changed along with these commits regarding wownero, so let's wait for the bug fixes I sent you, and update this commit hash again and test with the new dependencies with all the changes

lib/view_model/transaction_details_view_model.dart
```diff
@@ -214,7 +214,7 @@ abstract class TransactionDetailsViewModelBase with Store {
     final addressIndex = tx.additionalInfo['addressIndex'] as int;
     final feeFormatted = tx.feeFormatted();
     final _items = [
-      StandartListItem(title: S.current.transaction_details_transaction_id, value: tx.id),
+      StandartListItem(title: S.current.transaction_details_transaction_id, value: tx.txhash),
```
OmarHatem28, line 217: this should be changed then in other places in this file
for 2 reasons.
1. there are some places that uses `.id` regardless of the wallet type
2. to maintain consistency

PR #1540, opened 2024-07-18
Haven 4 update

Tay8NWWFKpz9JT4NXU0w:
This PR aims to achieve the following goals:
- Enable the build of Cake Wallet with Haven 4.0 for Android
- Adapt Cake Wallet to properly handle the new Haven Oracle format
- Fix a bug in Cake Wallet which prevented the "Send ALL" functionality to work properly

This is achieved by:
- Switching to Haven 4.0.2 in the build_haven.sh script. On Haven side, in Haven 4.0.2 the changes necessary to integrate again with Cake were already done.
- Adding cryptonote_format_utils_bacis to the linking process
- Remove unnecessary transformations in update_haven_rate.dart
- Change data types so that "amount" can be a nullable String, required when using "Send ALL"

How was this tested:
- Build process was tested under Debian on x86_64 architecture, with flutter and ndk in line with the Cake Wallet documentation
- Functional tests Cake Wallet were performed using Waydroid on x86_64
- Syncing works, proper balances and transactions were displayed, transfers of XHV, XUSD and XAU were tested and work 

Special considerations:
- INSTALL_VENDORED_LIBUNBOUND doesn't seem to work in my environment properly, so I added the copy of unbound to the haven folder in copy_monero_deps.sh. This is not part of the commits.

Out of scope:
- No tests of builds for IOS were performed, due to lack of proper environment for them. No adaptations were done to IOS build process apart from bumping the Haven version, due to lack of proper environment.
- Dependencies were not updated, but this will come in a separate PR. The reason for splitting this in two parts is to give you a chance to not perform these dependency updates at first, especially given that I am not sure how difficult it would be to perform the same steps for ios. However, in this PR, expat, openssl, sodium, zeromq will be  bumped to the versions also used by the latest Monero release.


konstantinullrich (approved):

OmarHatem28:
Thanks for taking the time and contributing to maintaining the Haven code in Cake Wallet. Unfortunately, we will be removing Haven from our codebase, as it's not being used currently and adds a lot of overload for users who are trying to build the app and for our team to maintain features with it.
again thank you for contributing and we are looking forward to having more contributions from you in the future 🙏

PR #1572, opened 2024-08-02
Monero stability and cleanup

konstantinullrich:
# Description

Clean up and get rid of unnecessary code. Also fixes a issue causing missing TX-Keys and missing Polyseeds (maybe) 

# Pull Request - Checklist  

- [x] Format code


cw_core/lib/monero_wallet_utils.dart
```diff
@@ -79,6 +79,7 @@ Future<bool> backupWalletFilesExists(String name) async {
       backupAddressListFile.existsSync();
 }
 
+// WARNING: Transaction keys and your Polyseed CANNOT be recovered if this file is deleted
 Future<void> removeCache(String name) async {
```
MrCyjaneK, line 83: I somehow think that removeCache() should remove cache, as it is (iirc) being called when wallet doesn't want to open.... so imo it still would be better to leave the wallet without polyseed / passphrase / seed offset but in a usable state..

MrCyjaneK (changes requested):
cw_wownero should also get the api/structs/* removal - as it is nearly the same code.

MrCyjaneK (approved):
lgtm

OmarHatem28 (approved):

PR #1582, opened 2024-08-07
Cw 688 avoid wallet file corruption

OmarHatem28:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_bitcoin/lib/bitcoin_wallet.dart
+++ b/cw_bitcoin/lib/bitcoin_wallet.dart
@@ -164,14 +164,11 @@ abstract class BitcoinWalletBase extends ElectrumWallet with Store {
       keysData = await WalletKeysFile.readKeysFile(name, walletInfo.type, password);
     }
 
-    walletInfo.derivationInfo ??= DerivationInfo(
-      derivationType: snp?.derivationType ?? DerivationType.electrum,
-      derivationPath: snp?.derivationPath,
-    );
+    walletInfo.derivationInfo ??= DerivationInfo();
 
     // set the default if not present:
-    walletInfo.derivationInfo!.derivationPath = snp?.derivationPath ?? electrum_path;
-    walletInfo.derivationInfo!.derivationType = snp?.derivationType ?? DerivationType.electrum;
+    walletInfo.derivationInfo!.derivationPath ??= snp?.derivationPath ?? electrum_path;
+    walletInfo.derivationInfo!.derivationType ??= snp?.derivationType ?? DerivationType.electrum;
 
     Uint8List? seedBytes = null;
     final mnemonic = keysData.mnemonic;
--- a/cw_bitcoin/lib/bitcoin_wallet.dart
+++ b/cw_bitcoin/lib/bitcoin_wallet.dart
@@ -152,7 +152,7 @@ abstract class BitcoinWalletBase extends ElectrumWallet with Store {
     if (!(await WalletKeysFile.hasKeysFile(name, walletInfo.type))) {
       final newKeysData =
           WalletKeysData(mnemonic: snp.mnemonic, xPub: snp.xpub, passphrase: snp.passphrase);
-      await WalletKeysFile.createKeysFile(name, walletInfo.type, password, newKeysData);
+      WalletKeysFile.createKeysFile(name, walletInfo.type, password, newKeysData);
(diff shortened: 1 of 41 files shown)


cw_core/lib/wallet_keys_file.dart
```diff
@@ -0,0 +1,76 @@
...
+import 'package:cw_core/pathForWallet.dart';
+import 'package:cw_core/utils/file.dart';
+import 'package:cw_core/wallet_type.dart';
+
+mixin WalletKeysFile {
+  // this needs to be overridden
+  Future<String> makePath() => throw UnimplementedError();
+
+  // this needs to be overridden
+  WalletKeysData get walletKeysData => throw UnimplementedError();
```
OmarHatem28, line 13: if needs to be overridden, then why give them a body and throw an exception, just make them abstract so they give error when using them
Screenshot 2024-08-08 at 1 27 58 AM

cw_bitcoin/lib/bitcoin_wallet.dart
```diff
@@ -143,8 +144,20 @@ abstract class BitcoinWalletBase extends ElectrumWallet with Store {
...
+    final WalletKeysData keysData;
+    // Migrate wallet from the old scheme to then new .keys file scheme
+    if (!(await WalletKeysFile.hasKeysFile(name, walletInfo.type))) {
+      final newKeysData =
+          WalletKeysData(mnemonic: snp.mnemonic, xPub: snp.xpub, passphrase: snp.passphrase);
+      await WalletKeysFile.createKeysFile(name, walletInfo.type, password, newKeysData);
+      keysData = newKeysData;
+    } else {
+      keysData = await WalletKeysFile.readKeysFile(name, walletInfo.type, password);
+    }
```
OmarHatem28, line 159: like we discussed, this should not be awaited, and preferably in a separate isolate

and btw, like this you are still depending on the snapshot file, so if it gets corrupted, the code won't reach this stage anyway

cw_bitcoin/lib/bitcoin_wallet.dart
```diff
@@ -143,49 +144,71 @@ abstract class BitcoinWalletBase extends ElectrumWallet with Store {
...
-      derivationPath: snp.derivationPath,
+      derivationType: snp?.derivationType ?? DerivationType.electrum,
+      derivationPath: snp?.derivationPath,
     );
 
     // set the default if not present:
-    walletInfo.derivationInfo!.derivationPath = snp.derivationPath ?? electrum_path;
-    walletInfo.derivationInfo!.derivationType = snp.derivationType ?? DerivationType.electrum;
+    walletInfo.derivationInfo!.derivationPath = snp?.derivationPath ?? electrum_path;
+    walletInfo.derivationInfo!.derivationType = snp?.derivationType ?? DerivationType.electrum;
```
OmarHatem28, line 176: derivation type and path should be in .keys files as they are sensitive also to how wallet would work and generate addresses and sign them

cw_bitcoin/lib/litecoin_wallet.dart
```diff
@@ -101,19 +102,38 @@ abstract class LitecoinWalletBase extends ElectrumWallet with Store {
...
+    } catch (e) {
+      if (!hasKeysFile) rethrow;
+    }
+
+    final WalletKeysData keysData;
+    // Migrate wallet from the old scheme to then new .keys file scheme
+    if (!hasKeysFile) {
+      final newKeysData =
+          WalletKeysData(mnemonic: snp!.mnemonic, xPub: snp.xpub, passphrase: snp.passphrase);
+      WalletKeysFile.createKeysFile(name, walletInfo.type, password, newKeysData);
```
OmarHatem28, line 120: why call the createKeysFile here when you are calling `saveKeysFile` in the `save` function?

(that's for all wallets)

cw_bitcoin/lib/bitcoin_wallet.dart
konstantinullrich: True
Will do

konstantinullrich: I'm not sure, because it is stored in WalletInfo as well (which is stored in the database). It does not make sense to override data with the same data from a different source

PR #1595, opened 2024-08-12
Update support links

tuxpizza:
Update support links + provider icons


lib/src/screens/support_other_links/support_other_links_page.dart
```diff
@@ -37,16 +41,16 @@ class SupportOtherLinksPage extends BasePage {
                 if (item is RegularListItem) {
                   return SettingsCellWithArrow(title: item.title, handler: item.handler);
                 }
-
                 if (item is LinkListItem) {
+                  bool lightIcon = false;
```
OmarHatem28, line 45: **Minor:**
better naming for readability can be `hasLightIcon`

OmarHatem28 (changes requested):

assets/images/dfx_dark.png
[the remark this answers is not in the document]
tuxpizza: This is correct, I had checked how it would look on Buy/Sell screen and they are sized correctly and look fine.
The reason for this change is so we're using original and non-modified partner icons, and also are higher quality than some of the ones that were blurry before.

lib/src/screens/support_other_links/support_other_links_page.dart
tuxpizza: Thanks, change made

OmarHatem28 (approved):

PR #748, opened 2023-01-26
Tagalog

nahuhh:
Issue Number (if Applicable): Fixes #704 

# Description

Please include a summary of the changes and which issue is fixed / feature is added. 

add tagalog
(tl, tgl, phl)
Flag already exist in repo

# Pull Request - Checklist  

- [ ] Initial Manual Tests Passed

PR #1611, opened 2024-08-15
Major overhaul of the Tagalog localization

rbrunner7:
This is a major overhaul of the Tagalog localization. Some of the more important systematic changes:

* Proper use and capitalization of product names, e.g. *Cake Wallet* and not *Cake wallet* or even *wallet ng cake*
* Proper acronyms, e.g. *PIN* and not *pin*
* In general, keeping the capitalization of the English strings intact
* Avoidance of "over-translation" of very special technical terms to keep them in English, e.g. *seed*, *view key*, *wallet* (and not *pitaka*)
* Use of *fee* instead of *bayad* because the latter, while maybe linguistically acceptable, is very easy to misunderstand in my opinion
* Removal of blanks in front of minus signs that connect word parts, e.g. *I-edit* instead of *I -edit*
* Re-adding of blanks at the end of some strings where the app will dynamically append other strings at runtime

I compared strings from the following 2 sources, took the better one in each case and often improved it further:
* [This PR](https://github.com/cake-tech/cake_wallet/pull/748) that was never merged for unknown reasons, carefully prepared "by hand" by a native speaker
* [This PR](https://github.com/cake-tech/cake_wallet/pull/1086), maybe from a Cake Wallet employee

For working on this I built an Excel sheet with English strings and the Tagalog strings from those two PRs side-by-side. If anybody is interested, I can provide that.


OmarHatem28 (approved):

OmarHatem28:
yup, thanks for taking the time and making the translation way better 🙏

rbrunner7:
It was a pleasure. Thanks to you for the fast merge.

PR #1618, opened 2024-08-16
Update flags

tuxpizza:
Updating country flags with higher resolution images


OmarHatem28 (approved):
if tested and doesn't cause any UI issue, feel free to merge it

PR #1628, opened 2024-08-20
make analyzer less red

MrCyjaneK:
# Description

Remove cw_*.dart files from analyzer
So my IDE is full of red, and I have to go and confirm that I do want to debug the app without fixing the errors (VSCode) or hide prompts warning me about that (Android Studio), because of errors in files that are not being used in the wallet.

Pros:
- no analyzer smalltalk for working code

Cons:
- no analyzer smalltalk for when something is wrong in the cw_*.dart files (it will still complain while compiling/running so it should be fine imo)

Also: `dart run tool/download_moneroc_prebuilds.dart` now runs gen_framework.sh to make running iOS app easier.

| With PR | Without |
| ---- | ---- |
| obraz | obraz |

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -10,7 +10,18 @@ analyzer:
     lib/generated/*.dart,
     cw_monero/ios/External/**,
     cw_shared_external/**,
-    shared_external/**]
+    shared_external/**,
+    lib/bitcoin/cw_bitcoin.dart,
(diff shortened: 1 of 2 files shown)


OmarHatem28 (approved):
oh it's only hiding them from the files in the main lib,
ok I thought you also excluded `cw_bitcoin`, `cw_ethereum` etc... from the analyzer.

all good then, thanks for adding them

PR #1620, opened 2024-08-17
subaddress fix

MrCyjaneK:
Issue Number (if Applicable): Fixes #

# Description

Please include a summary of the changes and which issue is fixed / feature is added. 

# Pull Request - Checklist  

- [ ] Initial Manual Tests Passed
- [ ] Double check modified code and verify it with the feature/task requirements
- [ ] Format code
- [ ] Look for code duplication
- [ ] Clear naming for variables and methods

PR #1630, opened 2024-08-21
CW-611-Refactor-Address-Handling

MrCyjaneK:
# Description

See jira ticket for details

# Pull Request - Checklist  

- [ ] This PR contains https://github.com/cake-tech/cake_wallet/pull/1620
- [ ] Format code


cw_monero/lib/monero_wallet_addresses.dart
```diff
@@ -27,6 +29,40 @@ abstract class MoneroWalletAddressesBase extends WalletAddresses with Store {
   @observable
   String address;
 
+  @override
+  String get latestAddress {
+    final accountIndex = 0;
```
OmarHatem28, line 34: shouldn't we be using the index of the account that the user is on?

cw_monero/lib/monero_wallet_addresses.dart
```diff
@@ -27,6 +29,40 @@ abstract class MoneroWalletAddressesBase extends WalletAddresses with Store {
...
+    }
+    return address;
+  }
+
+  @override
+  String get addressForExchange {
+    final accountIndex = 0;
+    var addressIndex = subaddress_list.numSubaddresses(accountIndex) - 1;
+    var address = getAddress(accountIndex: accountIndex,addressIndex: addressIndex);
+    while (hiddenAddresses.contains(address) || manualAddresses.contains(address)) {
```
OmarHatem28, line 49: I believe both `latestAddress` and `addressForExchange` should check against both `hiddenAddresses` and `manualAddresses` so there shouldn't be a difference here and they can be merged into one getter `latestUnusedAddress` for example

cw_monero/lib/monero_wallet_addresses.dart
```diff
@@ -27,6 +29,40 @@ abstract class MoneroWalletAddressesBase extends WalletAddresses with Store {
...
+      addressIndex++;
+      address = getAddress(accountIndex: 0, addressIndex: addressIndex);
+    }
+    return address;
+  }
+
+  String getLastAddress(int accountIndex) {
+    var addressIndex = subaddress_list.numSubaddresses(accountIndex) - 1;
+    var address = getAddress(accountIndex: accountIndex,addressIndex: addressIndex);
+    while (hiddenAddresses.contains(address)) {
```
OmarHatem28, line 59: maybe I am misunderstanding what the lastAddress is all about, cuz you seem to be skipping the check for manualAddresses every time for the `lastAddress`

OmarHatem28 (changes requested):

cw_monero/lib/api/transaction_history.dart
```diff
@@ -27,7 +33,6 @@ List<Transaction> getAllTransactions() {
   List<Transaction> dummyTxs = [];
 
   txhistory ??= monero.Wallet_history(wptr!);
-  monero.TransactionHistory_refresh(txhistory!);
```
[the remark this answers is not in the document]
MrCyjaneK: To not call it multiple times, it will be called only in one function with a mutex to prevent the crash that occurs when the wallet is large, and refresh is happening on a separate thread

lib/monero/cw_monero.dart
```diff
@@ -61,7 +61,13 @@ class CWMoneroSubaddressList extends MoneroSubaddressList {
   ObservableList<Subaddress> get subaddresses {
     final moneroWallet = _wallet as MoneroWallet;
     final subAddresses = moneroWallet.walletAddresses.subaddressList.subaddresses
-        .map((sub) => Subaddress(id: sub.id, address: sub.address, label: sub.label))
+        .map((sub) => Subaddress(
+          id: sub.id,
+          address: sub.address,
+          label: sub.label,
+          received: sub.balance??"unknown",
```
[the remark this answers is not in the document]
MrCyjaneK: The UI accepts string, and it is already formatted before passing - otherwise we would have to calculate the string on the UI insteaed of the wallet code

cw_monero/lib/monero_wallet_addresses.dart
MrCyjaneK: Correct, when user creates a new address I assume that they want to use it - so I don't exclude them from the list

lib/monero/cw_monero.dart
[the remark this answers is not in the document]
MrCyjaneK: received is a string, to account for formating so this is expected.

lib/src/screens/receive/receive_page.dart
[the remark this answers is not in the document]
MrCyjaneK: ShareUtil was a mistake, it shouldn't be latest - fixed in next commit.

Yeah, it will be better to just reset it when users enters the receive page

cw_monero/lib/api/wallet.dart
```diff
@@ -66,9 +66,14 @@ String getSeedLegacy(String? language) {
   return legacy;
 }
 
-String getAddress({int accountIndex = 0, int addressIndex = 0}) =>
-    monero.Wallet_address(wptr!,
+String getAddress({int accountIndex = 0, int addressIndex = 0}) {
+  while (monero.Wallet_numSubaddresses(wptr!, accountIndex: accountIndex)-1 < addressIndex) {
+    print("adding subaddress");
+    monero.Wallet_addSubaddress(wptr!, accountIndex: accountIndex);
+  }
```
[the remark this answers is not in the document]
MrCyjaneK: in case when the local state will be different than what's in the monero code, or when we hide excahnge address - then we will just do +1 to the index, and this function should make sure that the address exist (it would return it, but changes to for example label would cause errors)

cw_monero/lib/monero_wallet_addresses.dart
MrCyjaneK: We shouldn't use manual addresses for exchange refunds, that's why there are 2 getters, we want to display manual addresses, but we don't want to use them for exchanges.

cw_monero/lib/monero_wallet_addresses.dart
```diff
@@ -27,6 +29,30 @@ abstract class MoneroWalletAddressesBase extends WalletAddresses with Store {
   @observable
   String address;
 
+  @override
+  String get latestAddress {
+    var addressIndex = subaddress_list.numSubaddresses(account?.id??0) - 1;
+    var address = getAddress(accountIndex: account?.id??0, addressIndex: addressIndex);
+    while (hiddenAddresses.contains(address)) {
+      addressIndex++;
+      address = getAddress(accountIndex: 0, addressIndex: addressIndex);
```
OmarHatem28, line 38:
```suggestion
      address = getAddress(accountIndex: account?.id??0, addressIndex: addressIndex);
```

cw_monero/lib/monero_wallet_addresses.dart
```diff
@@ -27,6 +29,30 @@ abstract class MoneroWalletAddressesBase extends WalletAddresses with Store {
...
+    return address;
+  }
+
+  @override
+  String get addressForExchange {
+    var addressIndex = subaddress_list.numSubaddresses(account?.id??0) - 1;
+    var address = getAddress(accountIndex: account?.id??0, addressIndex: addressIndex);
+    while (hiddenAddresses.contains(address) || manualAddresses.contains(address) || subaddress_list.getRawLabel(accountIndex: account?.id??0, addressIndex: addressIndex).isNotEmpty) {
+      addressIndex++;
+      address = getAddress(accountIndex: 0, addressIndex: addressIndex);
```
OmarHatem28, line 50:
```suggestion
      address = getAddress(accountIndex: account?.id??0, addressIndex: addressIndex);
```

cw_monero/lib/monero_wallet_addresses.dart
```diff
@@ -109,7 +136,10 @@ abstract class MoneroWalletAddressesBase extends WalletAddresses with Store {
         accountIndex: accountIndex,
         defaultLabel: defaultLabel,
         usedAddresses: usedAddresses.toList());
-    subaddress = (subaddressList.subaddresses.isEmpty) ? Subaddress(id: 0, address: address, label: defaultLabel) : subaddressList.subaddresses.last;
+    subaddress = (subaddressList.subaddresses.isEmpty) ? Subaddress(id: 0, address: address, label: defaultLabel, balance: '0', txCount: 0) : subaddressList.subaddresses.last;
```
OmarHatem28, line 139: `allAddresses` and `getAllUnusedAddresses` are reversed, let's check if the list is also affected or not (i.e should it be `.first` or keep it as is `.last`)

MrCyjaneK: It is a normal ObservableList<> so shouldn't be reversed - keeping it last makes sense.

MrCyjaneK: https://github.com/cake-tech/cake_wallet/blob/b6089a2dd8d57533f5d7252333c11bb49c9dd466/cw_monero/lib/monero_subaddress_list.dart#L64-L65

cw_wownero/lib/api/transaction_history.dart
```diff
@@ -252,15 +253,13 @@ Future<PendingTransactionDescription> createTransactionMultDest(
 
 class Transaction {
   final String displayLabel;
-  String get subaddressLabel => wownero.Wallet_getSubaddressLabel(wptr!, accountIndex: 0, addressIndex: 0);
-  String get address => wownero.Wallet_address(
-    wptr!,
+  late String subaddressLabel = wownero.Wallet_getSubaddressLabel(wptr!, accountIndex: 0, addressIndex: 0);
+  late String address = getAddress(
```
OmarHatem28, line 257: shouldn't this take the accountIndex and addressIndex instead of 0s?

also why did you change the getters to late in some files?
late will throw an exception if it tried to initialize the variable again, and it's not even final, so it won't give any error if someone tries to set its value again, why not leave them as getters?

cw_monero/lib/monero_wallet_addresses.dart
OmarHatem28: ok thanks for looking into it

cw_wownero/lib/api/transaction_history.dart
MrCyjaneK: fixed the 0 and added proper accountIndex/addressIndex variable there. I've made them final as well - the idea is for the code to be future proof - with `late` it won't getn initialized until it is being used - and next uses are from memory instead of constantly calling a function - which should be more performant.

OmarHatem28 (approved):
overall looks good
some minor things are missing in Wownero, but not blockers, will commit the minor code suggestions I added and wait for some final tests

PR #1631, opened 2024-08-21
Cw 488 seed offset

konstantinullrich:
# Description

- Add switcher to select between a BIP39 and Electrum Wallet.
- Add Passphrase option on wallet creation.
- Implements Electrum Passphrases based on the official implementation [here](https://github.com/spesmilo/electrum/blob/bd845080b852ab19c54c2c4bd5ec8a36255b9e35/electrum/mnemonic.py#L175)
- Hide Length-selector on advanced settings from restore flow

# Pull Request - Checklist  

- [x] Format code


konstantinullrich:
Still some visual kinks to workout. eg. Passphrase is missing when reentering the advanced settings. The Passphrase does not propagate to previous screen

lib/store/settings_store.dart
```diff
@@ -587,6 +597,9 @@ abstract class SettingsStoreBase with Store {
   @observable
   SeedType moneroSeedType;
 
+  @observable
+  DerivationTypeSetting bitcoinDerivationType;
```
OmarHatem28, line 601: **Minor:**
would be easier to rename it to bitcoinSeedType (everywhere)

OmarHatem28 (changes requested):
let's change the UI as mentioned in ux channel
Seed type at the top, followed by the same UI components (exchange and fiat API settings)
then passphrase
then toggles

lib/view_model/advanced_privacy_settings_view_model.dart
```diff
@@ -52,7 +52,14 @@ abstract class AdvancedPrivacySettingsViewModelBase with Store {
     }
   }
 
-  bool get hasSeedTypeOption => type == WalletType.monero || type == WalletType.wownero;
+  bool get hasSeedTypeOption => [WalletType.monero, WalletType.wownero].contains(type);
+
+  bool get hasPassphraseOption => [
```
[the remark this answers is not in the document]
konstantinullrich: I added it to the credentials, but I did not implement it yet

lib/store/settings_store.dart
konstantinullrich: But it it does not change the seed type, but only the key derivation

lib/view_model/wallet_creation_vm.dart
```diff
@@ -115,11 +117,23 @@ abstract class WalletCreationVMBase with Store {
...
+              .getElectrumDerivations()[DerivationType.bip39]!
+              .firstWhere((element) => element.description == "Standard BIP84 native segwit");
+        }
+        return bitcoin!.getElectrumDerivations()[DerivationType.electrum]!.first;
       case WalletType.litecoin:
+        if (useBip39) {
+          return bitcoin!
+              .getElectrumDerivations()[DerivationType.bip39]!
+              .firstWhere((element) => element.description == "Default Litecoin");
+        }
```
OmarHatem28, line 136: just hard code the derivation path here better than comparing with a description string

OmarHatem28 (approved):

PR #1651, opened 2024-08-26
CW-695: Fix Zone Mismatch Error Triggered By Restoring Backup

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -81,7 +81,7 @@ Future<void> runAppWithZone() async {
 }
 
 Future<void> initializeAppAtRoot({bool reInitializing = false}) async {
-  if (reInitializing) await setDefaultMinimumWindowSize();
+  if (!reInitializing) await setDefaultMinimumWindowSize();
   await CakeHive.close();
   await initializeAppConfigs();
 }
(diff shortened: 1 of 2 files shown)


lib/main.dart
```diff
@@ -62,49 +65,27 @@ Future<void> main() async {
...
+        TopLevelErrorWidget(error: error, stackTrace: stackTrace),
       );
     }
 
     ExceptionHandler.onError(FlutterErrorDetails(exception: error, stack: stackTrace));
   });
 }
 
+Future<void> initializeAppAtRoot({bool reInitializing = false}) async {
+  if (reInitializing) await setDefaultMinimumWindowSize();
```
OmarHatem28, line 84: this was set when app is initialized normally, why change it to be only when restore?

OmarHatem28 (commented):
let's test this thoroughly tho

lib/main.dart
Blazebrain: Ohh, error from my end.

It's supposed to be inverse. I'll adjust it now.

The check is, if it's not a restore flow, don't call the defaults again.

Blazebrain: Done

OmarHatem28 (approved):

PR #1701, opened 2024-09-24
sync status eta

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_core/lib/sync_status.dart
+++ b/cw_core/lib/sync_status.dart
@@ -1,6 +1,10 @@
 abstract class SyncStatus {
   const SyncStatus();
   double progress();
+
+  String formattedProgress() {
+    return "${(progress() * 100).toStringAsFixed(2)}%";
+  }
 }
 
 class StartingScanSyncStatus extends SyncStatus {
@@ -12,10 +16,12 @@ class StartingScanSyncStatus extends SyncStatus {
 }
 
 class SyncingSyncStatus extends SyncStatus {
-  SyncingSyncStatus(this.blocksLeft, this.ptc);
+  SyncingSyncStatus(this.blocksLeft, this.ptc) {
+    updateEtaHistory(blocksLeft);
+  }
 
-  final double ptc;
-  final int blocksLeft;
+  double ptc;
+  int blocksLeft;
 
   @override
   double progress() => ptc;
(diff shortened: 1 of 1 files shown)


cw_core/lib/sync_status.dart
```diff
@@ -31,10 +37,162 @@ class SyncingSyncStatus extends SyncStatus {
...
+    // Apply smoothing to prevent ETA jumping
+    duration = _applySmoothing(duration);
+    lastEtaDuration = duration;
+
+    return _formatDuration(duration);
+  }
+
+  Duration getEtaDuration() {
+    DateTime now = DateTime.now();
+    DateTime? completionTime = calculateEta();
```
OmarHatem28: not nullable
```suggestion
    DateTime completionTime = calculateEta();
```

cw_core/lib/sync_status.dart
```diff
@@ -31,10 +37,162 @@ class SyncingSyncStatus extends SyncStatus {
...
+      return Duration(seconds: adjustedSeconds);
+    } else if (diff > 60) {
+      // If it's more than 1 minute difference, it's a small change so we move by max 30 seconds
+      final direction = newSeconds > currentSeconds ? 1 : -1;
+      final maxChange = 30;
+      final adjustedSeconds = currentSeconds + (direction * maxChange);
+      return Duration(seconds: adjustedSeconds);
+    }
+
+    return lastEtaDuration!;
```
OmarHatem28: this probably explains the issue Rotten reported, I believe this should return the `newDuration`
```suggestion
    return newDuration;
```

cw_core/lib/sync_status.dart
```diff
@@ -31,10 +37,162 @@ class SyncingSyncStatus extends SyncStatus {
...
+    for (int i = 0; i < sortedData.length - 1; i++) {
+      final current = sortedData[i];
+      final next = sortedData[i + 1];
+
+      final blocksProcessed = current.value - next.value;
+
+      if (blocksProcessed <= 0) continue; // Skip invalid data
+
+      final timeDifference = next.key.difference(current.key);
+      final timeSeconds = timeDifference.inMicroseconds / 1000000;
```
OmarHatem28: why not just directly use inSeconds?
```suggestion
      final timeSeconds = timeDifference.inSeconds;
```

cw_core/lib/sync_status.dart
Blazebrain: The reason we returned `lastEtaDuration`, not `newDuration` is that we're already checking for changes as low as 1 minute, and if we return `newDuration` here, then the eta would be jumping for "seconds" changes, for very small time differences.

But yeah, we can return `newDuration` and see if it gives better clarity and solves Rotten's scenario.

PR #1719, opened 2024-10-02
Add passphrase support for Eth, Polygon, and Tron

OmarHatem28:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_bitcoin_cash/lib/src/bitcoin_cash_wallet.dart
+++ b/cw_bitcoin_cash/lib/src/bitcoin_cash_wallet.dart
@@ -83,7 +83,7 @@ abstract class BitcoinCashWalletBase extends ElectrumWallet with Store {
       unspentCoinsInfo: unspentCoinsInfo,
       initialAddresses: initialAddresses,
       initialBalance: initialBalance,
-      seedBytes: await MnemonicBip39.toSeed(mnemonic, passphrase: passphrase),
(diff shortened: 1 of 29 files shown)


cw_bitcoin_cash/lib/src/bitcoin_cash_wallet.dart
```diff
@@ -83,7 +83,7 @@ abstract class BitcoinCashWalletBase extends ElectrumWallet with Store {
       unspentCoinsInfo: unspentCoinsInfo,
       initialAddresses: initialAddresses,
       initialBalance: initialBalance,
-      seedBytes: await MnemonicBip39.toSeed(mnemonic, passphrase: passphrase),
+      seedBytes: MnemonicBip39.toSeed(mnemonic, passphrase: passphrase),
```
konstantinullrich, line 86: You are right, the `await` here is not needed. But now `static Future<BitcoinCashWallet> create(` can be `static BitcoinCashWallet create(` or am I wrong?

OmarHatem28: You are right but it's minor can be skipped for now
Some code enhancements needs to be done in this flow anyway but I said I would do it in a separate pr better

konstantinullrich (approved):

konstantinullrich (approved):

PR #1724, opened 2024-10-04
Clean up Silent Payments copy

sethforprivacy:
Our copy for Silent Payments is quite inconsistent right now, this PR improves it across the board but should be tested to be sure strings display properly.


OmarHatem28 (approved):
**Minor:**
it would be better to make the translation changes through the localization scripts, so it translates the new strings to all other languages, but for these changes I guess it won't matter much so all good, just for future references

PR #1728, opened 2024-10-07
Fix issues when restoring wallets from backup

MrCyjaneK:
# Description

TODO

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/core/backup_service.dart
+++ b/lib/core/backup_service.dart
@@ -2,6 +2,7 @@ import 'dart:convert';
 import 'dart:io';
 import 'dart:typed_data';
 import 'package:cake_wallet/core/secure_storage.dart';
+import 'package:cake_wallet/entities/get_encryption_key.dart';
 import 'package:cake_wallet/entities/transaction_description.dart';
 import 'package:cake_wallet/themes/theme_list.dart';
 import 'package:cw_core/root_dir.dart';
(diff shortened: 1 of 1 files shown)


lib/core/backup_service.dart
```diff
@@ -206,6 +230,16 @@ class BackupService {
...
+
+    if (!_transactionDescriptionBox.isOpen) {
+      final transactionDescriptionsBoxKey = await getEncryptionKey(secureStorage: secureStorageShared, forKey: TransactionDescription.boxKey);
+      final transactionDescriptionBox = await CakeHive.openBox<TransactionDescription>(
+        TransactionDescription.boxName,
+        encryptionKey: transactionDescriptionsBoxKey,
+      );
+      await transactionDescriptionBox.putAll(descriptionsMap);
+      return;
+    }
```
OmarHatem28, line 242: why would it be closed? it gets opened in main before the app starts
and if indeed it's necessary, please use `this.secureStorage` instead

lib/core/backup_service.dart
```diff
@@ -148,14 +157,29 @@ class BackupService {
     await _importPreferencesDump();
   }
 
+  // checked with .endsWith - so this should be the last part of the filename
+  static const ignoreFiles = [
+    "flutter_assets/kernel_blob.bin",
+    "flutter_assets/vm_snapshot_data",
+    "flutter_assets/isolate_snapshot_data",
+    ".lock",
+  ];
```
OmarHatem28, line 166: out of curiosity, why do we ignore them?

lib/core/backup_service.dart
OmarHatem28: ok so this function `_reloadHiveWalletInfoBox` is the one that closes it, but it also closes all boxes and just open the walletInfo box, so let's make sure that's the expected behavior, cuz I don't think it is

MrCyjaneK: It is an issue that [Ramo reported](https://cakewallet.slack.com/archives/C04CSQESA58/p1728607357598899?thread_ts=1728577025.568509&cid=C04CSQESA58), and I was able to reproduce later, with this fix there is no error

lib/core/backup_service.dart
MrCyjaneK: as for "flutter_assets/kernel_blob.bin", "flutter_assets/vm_snapshot_data", "flutter_assets/isolate_snapshot_data" if we backup/restore these files the app will crash in runtime when on debug build
and .lock files are not required - so I just included them while also adding the list - so we have even less files in backup.

PR #1732, opened 2024-10-07
fix secure storage keys not being cleared on a fresh install

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (commented):
please test this on an already existing app with some wallets and see if it will have any strange behavior

also please fix the build issue with linux

OmarHatem28:
ok so while reviewing this, I noticed that the main issue that this PR was made for, is not gonna be solved by this.

so the main issue is, on new installs, the app opens the pin screen and asks for pin, which from what I see can happen only if the shared prefs has a value for this key `PreferencesKey.currentWalletName`, or something else that is not very clear, but this doesn't seem to be a fix for it

fossephate:
IIRC, shared preferences are cleared on uninstalls, but secure storage is not (only on iOS)

This PR should fix this by deleting ALL secure storage entries when a fresh install is detected

If I'm understanding correctly you're saying we should delete the shared preference key: currentWalletName as well? I don't think it's necessary since uninstalls should clear the shared preferences but it probably can't hurt

OmarHatem28:
what I am saying is that the issue won't be solved with this change, because the app shows the PIN screen based on the shared prefs value, and since shared prefs value already gets cleared, then the issue is something else and not related to secure storage

fossephate:
🤔 that's weird because the issue did appear to be fixed in my testing; were you able to reproduce the issue / did the fix not work for you?

Also: this was an iOS specific bug right? Or did we have this issue pop up on android as well?

Because on iOS the shared_preferences package just wraps NSUSerDefaults and that should definitely be cleared on uninstall, the only exception I found being related to app groups:

https://stackoverflow.com/questions/24985825/nsuserdefaults-not-cleared-after-app-uninstall-on-simulator (see the 2nd reply about app groups)

OmarHatem28:
let's discuss it when you get back no problem, enjoy your vacation 🙏

PR #1733, opened 2024-10-08
CW 781 replace all print statements with printV

MrCyjaneK:
# Description

Goals:
- Replace all print statements with printV
- Show caller (path/file:line:column functionName)
- Add an action that makes sure that no normal print statement is being used in dart files (printV is a drop-in replacement and using print statements would make debugging harder in the future)
- Optional: At little cost I can implement logging to file, with timestamps and details - we could use that in error reporting screen)

# Pull Request - Checklist  

- [ ] Format code


cw_core/lib/utils/print_verbose.dart
```diff
@@ -0,0 +1,84 @@
+void printV(dynamic content) {
+  CustomTrace programInfo = CustomTrace(StackTrace.current);
+  printV("${programInfo.fileName}#${programInfo.lineNumber}:${programInfo.columnNumber} ${programInfo.callerFunctionName}: $content");
```
OmarHatem28, line 3: recursive call

```suggestion
  print("${programInfo.fileName}#${programInfo.lineNumber}:${programInfo.columnNumber} ${programInfo.callerFunctionName}: $content");
```

cw_core/lib/utils/print_verbose.dart
```diff
@@ -0,0 +1,84 @@
...
+  String? functionName;
+  String? callerFunctionName;
+  int? lineNumber;
+  int? columnNumber;
+
+  CustomTrace(this._trace) {
+    try {
+      _parseTrace();
+    } catch (e) {
+      printV("Unable to parse trace (printV): $e");
```
OmarHatem28, line 21: same
```suggestion
      print("Unable to parse trace (printV): $e");
```

PR #1739, opened 2024-10-09
CW-782: Show error report popup without cooldown

MrCyjaneK:
# Description

improve exception throwing on broken wallets
- put _lastOpenedWallet to avoid issues on windows (file is currently open by)
- don't throw corruptedWalletsSeed - instead store it inside of secureStorage
- await ExceptionHandler.onError calls where possible to makse sure that popup won't be canceled by some UI element
- adjust BaseAlertDialog to be scrollable if the text is too long
- add ExceptionHandler.resetLastPopupDate - that can be called when we want to show error report screen (bypassing cooldown)
# Pull Request - Checklist  

- [ ] Format code


lib/utils/exception_handler.dart
```diff
@@ -93,7 +99,7 @@ class ExceptionHandler {
       return;
     }
 
-    _saveException(
+    await _saveException(
```
OmarHatem28, line 102: we don't want to await saving the exception, we can still show the user the popup until it gets saved

lib/utils/exception_handler.dart
```diff
@@ -264,31 +275,31 @@ class ExceptionHandler {
...
+      final seeds = await secureStorageShared.read(key: "corruptedWalletsSeed");
+      await secureStorageShared.delete(key: "corruptedWalletsSeed");
       final shouldCopy = await showPopUp<bool?>(
         context: navigatorKey.currentContext!,
         builder: (context) {
           return AlertWithTwoActions(
             isDividerExist: true,
             alertTitle: S.of(context).error,
-            alertContent: content,
+            alertContent: content+"\n"+(seeds??""),
```
OmarHatem28, line 302: here I believe it would be duplicated, as the content is the error message, which contains the seeds as well

lib/src/screens/wallet_list/wallet_list_page.dart
```diff
@@ -454,8 +456,12 @@ class WalletListBodyState extends State<WalletListBody> {
             });
           }
         } catch (e) {
+          await ExceptionHandler.resetLastPopupDate();
+          final err = e.toString() + ((await secureStorageShared.read(key: "corruptedWalletsSeed"))??"unable to retrieve seeds");
+          await secureStorageShared.delete(key: "corruptedWalletsSeed");
+          await ExceptionHandler.onError(FlutterErrorDetails(exception: err));
```
OmarHatem28, line 462: same,
same

lib/di.dart
OmarHatem28, line 1: haven't reviewed this file yet, but double check it as well please

OmarHatem28 (changes requested):

lib/core/wallet_loading_service.dart
```diff
@@ -94,13 +97,15 @@ class WalletLoadingService {
...
-            corruptedWalletsSeeds += "\nFailed to  fetch $name seeds: $e";
+            corruptedWalletsSeeds += "\nFailed to fetch $name seeds: $e";
           }
+          secureStorageShared.write(key: "corruptedWalletsSeed", value: corruptedWalletsSeeds);
         }
       }
 
       // if all user's wallets are corrupted throw exception
-      throw error.toString() + "\n\n" + corruptedWalletsSeeds;
+      secureStorageShared.write(key: "corruptedWalletsSeed", value: corruptedWalletsSeeds);
```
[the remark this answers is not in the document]
MrCyjaneK: Otherwise the error report may contain the seed - which wouldn't be ideal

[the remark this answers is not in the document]
OmarHatem28: Well we already ignore the error and don't save it, if the error string contains "corrupted seeds"

xximwon (commented):
G

PR #1744, opened 2024-10-11
Mweb enhancements 3

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -2127,19 +2128,26 @@ abstract class ElectrumWalletBase
...
       case ConnectionStatus.failed:
-        syncStatus = LostConnectionSyncStatus();
+        if (syncStatus is! LostConnectionSyncStatus) {
+          syncStatus = LostConnectionSyncStatus();
+        }
         break;
       case ConnectionStatus.connecting:
-        syncStatus = ConnectingSyncStatus();
+        if (syncStatus is! ConnectedSyncStatus) {
+          syncStatus = ConnectingSyncStatus();
```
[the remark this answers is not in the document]
fossephate: yeah this was just a mistake

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -2142,30 +2158,44 @@ abstract class ElectrumWalletBase
...
       case ConnectionStatus.connected:
         if (syncStatus is NotConnectedSyncStatus ||
             syncStatus is LostConnectionSyncStatus ||
             syncStatus is ConnectingSyncStatus) {
-          syncStatus = AttemptingSyncStatus();
-          startSync();
+          // make sure we're not triggering this more than necessary:
+          if (this.syncStatus is NotConnectedSyncStatus ||
+              this.syncStatus is LostConnectionSyncStatus ||
+              this.syncStatus is ConnectingSyncStatus) {
```
OmarHatem28, line 2171: not needed any more

cw_mweb/lib/cw_mweb.dart
```diff
@@ -54,7 +54,7 @@ class CwMweb {
     log("Attempting to connect to server on port: $_port");
 
     // wait for the server to finish starting up before we try to connect to it:
-    await Future.delayed(const Duration(seconds: 5));
+    await Future.delayed(const Duration(seconds: 8));
```
OmarHatem28, line 57: 8? isn't that a bit much?

cw_bitcoin/lib/litecoin_wallet.dart
```diff
@@ -95,6 +96,35 @@ abstract class LitecoinWalletBase extends ElectrumWallet with Store {
...
+        } catch (_) {
+          nodeHeight = 0;
+        }
+
+        if (nodeHeight == 0) {
+          // we aren't connected to the litecoin node, so the current electrum_wallet reactions will take care of this case for us
+        } else {
+          // we're connected to the litecoin node, but we failed to connect to mweb, try again after a few seconds:
+          await Future.delayed(const Duration(seconds: 2));
+          startSync();
```
OmarHatem28, line 114: should we call `mweb.stop()` here?

cw_bitcoin/lib/litecoin_wallet.dart
```diff
@@ -915,10 +950,31 @@ abstract class LitecoinWalletBase extends ElectrumWallet with Store {
...
+          case UnspentCoinType.any:
+            return true;
+        }
+      }).toList();
 
+      for (var input in availableInputs) {
+        if (input.bitcoinAddressRecord.type == SegwitAddresType.mweb) {
+          hasMwebInput = true;
+        }
+      }
```
OmarHatem28, line 976: this will be always true in any case except peg-in, as long as the user has mweb funds, regardless of whether they will be used in this transaction or not

OmarHatem28 (changes requested):

cw_bitcoin/lib/litecoin_wallet.dart
fossephate: we filter the availableInputs list to only include coins the user has selected:
```
if (!utx.isSending || utx.isFrozen) {
```

so this case should be covered, but this highlights that the variable should probably renamed to something like `selectedInputs` to be less confusing

OmarHatem28: I guess you missed my point,
at the end, you are setting `hasMwebInput` to true, if you find an mweb input in the user's selected inputs, but that doesn't mean that the tx will contain it, like it's selected yeah, but the transaction might make it with only normal ltc inputs

ex.
let's say I have 3 unspent coins
1. segwit 0.5
2. mweb 0.3
3. mweb 0.2

and I have all of them selected
then I go ahead and send 0.4
with this logic, it will mark it as `hasMwebInput`, while the actual inputs that will be used in tx might be only the first one (segwit 0.5), so that means the tx doesn't have mweb inputs, but the bool will be true regardless

fossephate: oh I see
I'm not sure how to tell if we're using mweb inputs then
the CreateRequest response from mwebd doesn't list mweb inputs (only non-mweb ones) in tx2.inputs
and the transactionCredentials object only has the outputs

OmarHatem28 (approved):

PR #1747, opened 2024-10-14
Cw 660 add ledger monero

konstantinullrich:
# Description

This adds support for ledger hardware wallets.

# Pull Request - Checklist  

- [ ] Format code


lib/src/screens/wallet_list/wallet_list_page.dart
OmarHatem28, line 1: please make sure you set the line width to 100 as we agreed, because I tried reformatting this page, and it didn't change except 2 lines

and then reformat the code again so that it removes all these changes

Screenshot 2024-11-04 at 5 07 26 AM

lib/src/screens/connect_device/monero_hardware_wallet_options_page.dart
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  bool _effectsInstalled = false;
+
+  void _setEffects(BuildContext context) {
+    if (_effectsInstalled) return;
+
+    reaction((_) => _walletHardwareRestoreVM.error, (String? error) {
+      if (error != null) {
+        if (error == S.current.ledger_connection_error) Navigator.of(context).pop();
```
OmarHatem28, line 199: why do we just pop? is it displayed to the user somewhere else?

lib/src/screens/connect_device/connect_device_page.dart
```diff
@@ -247,11 +261,27 @@ class ConnectDevicePageBodyState extends State<ConnectDevicePageBody> {
...
         ),
       ),
     );
   }
+
+  void _onChangeWallet() {
+    Navigator.of(context).pushNamed(
+      Routes.walletList,
+      arguments: (BuildContext context) =>
+          Navigator.of(context).pushNamed(Routes.dashboard),
```
OmarHatem28, line 284: shouldn't this be `pushAndRemoveUntil`?

OmarHatem28 (changes requested):
why are the wownero changes needed?

lib/src/screens/wallet_list/wallet_list_page.dart
konstantinullrich: I reverted the changes, but I assume `dart format` has not been run in this file. because my format changes have the line length set to 100.

lib/src/screens/wallet_list/wallet_list_page.dart
```diff
@@ -436,26 +415,50 @@ class WalletListBodyState extends State<WalletListBody> {
...
+          }
+
+          showPopUp<void>(
+            context: context,
+            builder: (BuildContext context) => AlertWithOneAction(
+                alertTitle: S.of(context).proceed_on_device,
+                alertContent: S.of(context).proceed_on_device_description,
+                buttonText: S.of(context).cancel,
+                buttonAction: () => Navigator.of(context).pop()),
+          );
```
OmarHatem28, line 442: wouldn't this show up in normal wallets as well?

konstantinullrich: Nope, requireHardwareWalletConnection is only true if it is a monero hardware wallet

lib/src/screens/connect_device/monero_hardware_wallet_options_page.dart
konstantinullrich: It is shown right after so we co back to the connection screen and then we show the popup. Navigator.pop() is not a return.

lib/src/screens/connect_device/connect_device_page.dart
OmarHatem28: the second one, push to dashboard, so that after the user goes to the dashboard, if he tapped back, he would get out of the app not go back to the wallet list page

lib/src/screens/wallet_list/wallet_list_page.dart
OmarHatem28: yes, but this is out of the if condition scope

OmarHatem28 (approved):

PR #1749, opened 2024-10-15
fix P2wpkhAddress btc ltc validation

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (commented):
did you test this?
would it work with just providing the Regex object? shouldn't we pass the regex pattern not the whole object?
i.e
```Dart
RegExp(r'(bc|tb)1q[ac-hj-np-z02-9]{25,39}}').pattern
```

Serhii-Borodenko:
I tested it, and it was working, but this is the correct way to use a pattern.

OmarHatem28 (approved):

PR #1751, opened 2024-10-15
CW-766 fix coin freezing

MrCyjaneK:
# Description

# Pull Request - Checklist  

- [ ] Format code


MrCyjaneK:
@OmarHatem28 I thought this one was merged, it may will fix some of the issues

cw_monero/lib/monero_wallet.dart
```diff
@@ -729,9 +725,8 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
 
   void _askForUpdateBalance() {
     final unlockedBalance = _getUnlockedBalance();
-    final fullBalance = _getFullBalance();
-    final frozenBalance = _getFrozenBalance();
-
+    final fullBalance = _getUnlockedBalance() + _getFrozenBalance();
+    final frozenBalance = 0; // this is calculated on the monero side now
```
OmarHatem28, line 729: why this change? I don't think it's correct, and it will change the balance with the `frozenBalance` variable, which would be 0

lib/view_model/send/send_view_model.dart
```diff
@@ -530,6 +531,17 @@ abstract class SendViewModelBase extends WalletChangeListenerViewModel with Stor
       throw Exception('Priority is null for wallet type: ${wallet.type}');
     }
 
+    if (hasCoinControl) {
+      final vm = getIt.get<UnspentCoinsListViewModel>(param1: coinTypeToSpendFrom);
```
OmarHatem28, line 535: please add this as a dependency in di.dart for the send_view_model
so that if we modify this, it's gonna be only 1 file to look into

OmarHatem28 (changes requested):

cw_wownero/lib/api/transaction_history.dart
```diff
@@ -6,13 +6,16 @@ import 'package:cw_wownero/api/exceptions/creation_transaction_exception.dart';
...
+import 'package:cw_wownero/exceptions/wownero_transaction_creation_exception.dart';
 import 'package:ffi/ffi.dart';
 import 'package:monero/wownero.dart' as wownero;
 import 'package:monero/src/generated_bindings_wownero.g.dart' as wownero_gen;
 
 
 String getTxKey(String txId) {
-  return wownero.Wallet_getTxKey(wptr!, txid: txId);
+  final ret = wownero.Wallet_getTxKey(wptr!, txid: txId);
+  wownero.Wallet_status(wptr!);
```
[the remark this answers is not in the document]
MrCyjaneK: To clear the error string so it won't get caught by other functions, such as sync status check.

cw_monero/lib/monero_wallet.dart
MrCyjaneK: Otherwise balance will be incorrect, monero is calculating the frozen balance now

PR #1754, opened 2024-10-16
Fix fiat rate on receive screen.

konstantinullrich:
# Description

This fixes a divide with 0 issue on the receive screen

# Testing

Please test for the following:
Go to the receive screen and select a fiat currency other than your default. 
Now enter a amount and scan the QR code with a other wallet. The fiat value of the amount encoded should correspond with the fiat value entered.  

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/lib/view_model/wallet_address_list/wallet_address_list_view_model.dart
+++ b/lib/view_model/wallet_address_list/wallet_address_list_view_model.dart
@@ -42,12 +42,11 @@ abstract class PaymentURI {
 }
 
 class MoneroURI extends PaymentURI {
-  MoneroURI({required String amount, required String address})
-      : super(amount: amount, address: address);
+  MoneroURI({required super.amount, required super.address});
 
   @override
   String toString() {
-    var base = 'monero:' + address;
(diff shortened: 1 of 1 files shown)


OmarHatem28 (commented):
too many reformatting changes for just a small change
let's make the reformatting in a separate PR in future instances

and let's handle the case we discussed about entering the amount then selecting the currency

lib/view_model/wallet_address_list/wallet_address_list_view_model.dart
```diff
@@ -624,13 +625,15 @@ abstract class WalletAddressListViewModelBase
...
         _fiatRate = value;
+        _convertAmountToCrypto();
       });
     }
   }
 
   @action
   void changeAmount(String amount) {
     this.amount = amount;
+    this._rawAmount = amount;
```
OmarHatem28, line 636: why new variable? why not use `this.amount` ?

OmarHatem28 (approved):

PR #1755, opened 2024-10-16
transaction description key id address

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/view_model/send/send_view_model.dart
+++ b/lib/view_model/send/send_view_model.dart
@@ -472,18 +472,17 @@ abstract class SendViewModelBase extends WalletChangeListenerViewModel with Stor
         nano!.updateTransactions(wallet);
       }
 
-      final walletAddress = wallet.type == WalletType.bitcoin
-          ? wallet.walletAddresses.primaryAddress
-          : wallet.walletAddresses.address;
 
       if (pendingTransaction!.id.isNotEmpty) {
+
+        final descriptionKey = '${pendingTransaction!.id}_${wallet.walletAddresses.primaryAddress}';
         _settingsStore.shouldSaveRecipientAddress
             ? await transactionDescriptionBox.add(TransactionDescription(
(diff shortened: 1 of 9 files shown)


lib/view_model/send/send_view_model.dart
```diff
@@ -472,12 +472,19 @@ abstract class SendViewModelBase extends WalletChangeListenerViewModel with Stor
         nano!.updateTransactions(wallet);
       }
 
+      final walletAddress = wallet.type == WalletType.bitcoin
+          ? wallet.walletAddresses.primaryAddress
+          : wallet.walletAddresses.address;
```
OmarHatem28, line 477: this will be different for Monero each time

lib/view_model/transaction_details_view_model.dart
```diff
@@ -110,9 +110,14 @@ abstract class TransactionDetailsViewModelBase with Store {
           } catch (e) {}
         }));
 
+    final walletAddress = wallet.type == WalletType.bitcoin
+        ? wallet.walletAddresses.primaryAddress
+        : wallet.walletAddresses.address;
```
OmarHatem28, line 115: same

OmarHatem28 (changes requested):
in case of Monero this would fail, so you can either add a getter to Monero to get the primary address and use that like we do in Electrum, or if you have another idea in mind

Serhii-Borodenko:
I think it would be better to add a primary address for all wallets

OmarHatem28 (approved):

OmarHatem28:
> I think it would be better to add a primary address for all wallets

could have been better to set it from the parent `wallet_address.dart` like this
```
String get primaryAddress => address;
```
since it will be like this for all of the wallets, without having to modify each wallet file, and you would only need to override it in the needed wallets, but anyway that's minor, just thought I would let you know about it for future instances

PR #1757, opened 2024-10-18
Automatically upgrade non-SSL nodes to SSL if they fail

MrCyjaneK:
# Description

This PR automatically upgrades non-SSL nodes to SSL if it detects a Location header or an error that indicates that HTTPS might be an option.

# Pull Request - Checklist  

- [ ] Format code


cw_core/lib/node.dart
```diff
@@ -200,9 +200,29 @@ class Node extends HiveObject with Keyable {
...
       );
-
       client.close();
 
+      if (
+        response.body.contains("400 Bad Request") // Some other generic error
+        || response.body.contains("plain HTTP request was sent to HTTPS port") // Cloudflare
+        || response.headers["location"] != null // Generic reverse proxy
+        || response.body.contains("301 Moved Permanently") // Poorly configured generic reverse proxy
+      ) {
```
OmarHatem28, line 210: shouldn't we also `&& !useSSL` to avoid any possibility for a recursion

cw_core/lib/node.dart
```diff
@@ -200,9 +200,29 @@ class Node extends HiveObject with Keyable {
...
+        useSSL = true;
+        try {
+          final ret = await requestMoneroNode();
+          if (ret == true) {
+            await save();
+            return ret;
+          }
+          useSSL = oldUseSSL;
+        } catch (e) {
+          useSSL = false;
```
OmarHatem28, line 222: I believe better return it to the old state
i.e
```Dart
useSSL = oldUseSSL;
```

OmarHatem28 (approved):

PR #1772, opened 2024-10-24
Cw 649 rbf improvements opt in

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -1640,25 +1637,103 @@ abstract class ElectrumWalletBase extends WalletBase<
...
+          int outputAmount = output.value.toInt();
 
           if (outputAmount > _dustAmount) {
             int deduction = (outputAmount - _dustAmount >= remainingFee)
                 ? remainingFee
                 : outputAmount - _dustAmount;
+
             outputs[i] = BitcoinOutput(
-                address: outputs[i].address, value: BigInt.from(outputAmount - deduction));
+                address: output.address, value: BigInt.from(outputAmount - deduction));
```
OmarHatem28, line 1736: why do we keep the output with the dust amount?
why not just check if the `outputAmount - remainingFee > _dustAmount` then simply deduct the remaining fee from the output and break

else, then deduct the whole output amount from remainingFee and remove the whole output, and decrement the variable `i`
i.e
```Dart
if (outputAmount - remainingFee > _dustAmount) {
    _remainingFee = 0;
    outputs[i] = BitcoinOutput(
                address: output.address, value: BigInt.from(outputAmount - remainingFee));
   break;
} else {
    _remainingFee -= outputAmount;
    outputs.remove(i);
    i--;
}
```

OmarHatem28 (changes requested):

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -1696,17 +1771,11 @@ abstract class ElectrumWalletBase extends WalletBase<
...
         final key =
-            privateKeys.firstWhereOrNull((element) => element.getPublic().toHex() == publicKey);
-
+        privateKeys.firstWhereOrNull((element) => element.getPublic().toHex() == publicKey);
         if (key == null) {
           throw Exception("Cannot find private key");
         }
-
-        if (utxo.utxo.isP2tr()) {
-          return key.signTapRoot(txDigest, sighash: sighash);
```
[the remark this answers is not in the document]
Serhii-Borodenko: this was accidentally removed during a merge. I’ve restored it to handle Taproot signing

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -1543,14 +1543,17 @@ abstract class ElectrumWalletBase extends WalletBase<
     final bundle = await getTransactionExpanded(hash: txId);
     final outputs = bundle.originalTransaction.outputs;
 
-    final changeAddresses = walletAddresses.allAddresses.where((element) => element.isHidden);
+    final receiverAmount = outputs.isNotEmpty ? outputs[0].amount.toInt() : null;
```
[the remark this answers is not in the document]
Serhii-Borodenko: Receiver amount is what the receiver should get, not just the sum of all outputs. I’ve improved this to sum outputs that aren’t change outputs, as using the first output isn’t reliable

cw_bitcoin/lib/electrum_wallet.dart
Serhii-Borodenko: That makes sense, I’ve updated the code accordingly

Serhii-Borodenko: I’ve double-checked, and if I remove the output completely, I get the error “Sum value of utxo not spending” because the transaction becomes unbalanced—the inputs no longer equal the outputs plus the fee.

OmarHatem28: but like this it might also error out because of the dust amount

my solution is to remove it, and you can add the remaining funds (which would be a dust amount) to the fees, so that the transaction becomes balanced again

OmarHatem28 (approved):

PR #1782, opened 2024-10-30
Cw 805 fix bch refund address

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [x] Format code


lib/view_model/contact_list/contact_list_view_model.dart
```diff
@@ -99,18 +99,22 @@ abstract class ContactListViewModelBase with Store {
...
 
   bool _isValidForCurrency(ContactBase element) {
-    return _currency == null ||
-        element.type == _currency ||
-        element.type.title == _currency!.tag ||
-        element.type.tag == _currency!.tag;
+    if (element.name.contains('(Silent Payments)')) return false;
+    if (element.name.contains('MWEB)')) return false;
+
+    return element.type == _currency ||
```
OmarHatem28, line 113: why not just keep `_currency == null ||` before this, and you can remove the checks you have in `get walletContactsToShow` and `get contactsToShow`

lib/view_model/contact_list/contact_list_view_model.dart
```diff
@@ -99,18 +99,22 @@ abstract class ContactListViewModelBase with Store {
 
   Future<void> delete(ContactRecord contact) async => contact.original.delete();
 
-  @computed
-  List<ContactRecord> get contactsToShow =>
-      contacts.where((element) => _isValidForCurrency(element)).toList();
+  ObservableList<ContactRecord> get contactsToShow => _currency != null ?
```
OmarHatem28, line 102: shouldn't this be annotated with computed?

OmarHatem28 (commented):
I wouldn't say that `_formatBchAddress` is a good name since it's not for only BCH addresses and you just make an if condition there

so I would prefer if you rename it to be something like `normalizeAddress` or something and add an if condition there, so that if we want to normalize more addresses, it would be generic and we can just add to this function

lib/view_model/contact_list/contact_list_view_model.dart
Serhii-Borodenko: moved _currency == null to  _isValidForCurrency. This check acts as a flag for the exchange screen, if it’s true, we don’t need validated contacts at all. So, it works this way as well.

lib/view_model/contact_list/contact_list_view_model.dart
Serhii-Borodenko: Since I modified it to an observable list, I don’t think it needs to be computed.

lib/view_model/contact_list/contact_list_view_model.dart
OmarHatem28: yeah I agree, but just a bit cleaner, since the loop isn't a big one anyway so it won't matter

OmarHatem28 (approved):

PR #1858, opened 2024-12-05
CW-829 Solana Enhancements

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (changes requested):
are all the errors caught correctly?
also, please throw an exception class and handle the message in the send_view_model as we do for electrum wallets, so that they are translated

Blazebrain:
> are all the errors caught correctly? also, please throw an exception class and handle the message in the send_view_model as we do for electrum wallets, so that they are translated

Yes, they are all being caught correctly. It displays the right error to the user. 

We can do the exception classes also, makes for more clarity.

OmarHatem28 (approved):

PR #1865, opened 2024-12-09
CW-854 Monero Wallet Group Fix

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/view_model/wallet_list/wallet_list_view_model.dart
```diff
@@ -99,9 +100,29 @@ abstract class WalletListViewModelBase with Store {
...
+                wallet.derivationInfo?.derivationType == DerivationType.electrum;
+
+        // Exclude if any of these conditions are true
+        return isNonBIP39Wallet || isNanoDerivationType || isElectrumDerivationType;
+      });
+
+      if (shouldAddToSingleWallets) {
+        for (var wallet in group.wallets) {
+          singleWalletsList.add(convertWalletInfoToWalletListItem(wallet));
+        }
```
OmarHatem28, line 123: you are adding all the wallets in this group to single wallet list if `any` wallet was wrong?

so if the whole group was correctly grouped, but only 1 wallet was grouped by mistake, this code, will add all wallets in this group to single wallet list, why not just specifically add the odd wallet

i.e 
```Dart
singleWalletsList.addAll(group.wallets.where((wallet) { // your checks }).map((e) => convertWalletInfoToWalletListItem(e)));
```

Blazebrain: Yeah, that does make sense

I went with the logic that if one is disrupted, there is a high probability that everything in that group is disrupted. But it's better to handle them individually instead of adding them as a group.

PR #1867, opened 2024-12-10
Salvium Support

0xskydb:


MrCyjaneK:
Hey! We are currently in process of removing old build scripts, now monero (and monero-like) currencies are being built using [monero_c](https://github.com/MrCyjaneK/monero_c/) repository.

Simplest way to get started is to
1. add a submodule in monero_c repo
2. adjust workflow file to build it for all targets.
3. Create patches (if required)
Once monero_c will have support for Salvium adding support for it to cake is much easier.

https://github.com/MrCyjaneK/monero_c/pull/83 Here's PR for integration of zano in monero_c.

0xskydb:
Thanks for your comment. @MrCyjaneK 
Can you share any guide/link how to add support for cake after monero_c work done?

MrCyjaneK:
I didn't check this pr in detail, but if you are using wallet2_api.h or any other header directly then you don't have much to do (if you have dart UI ready), mostly replacing native calls with calls from monero_c repo (which are essentially the same but more unified).
You can take a look at https://github.com/cake-tech/cake_wallet/pull/1793. If you would have any issues I'm happy to help you

PR #1873, opened 2024-12-12
Cw 853 mweb utxos are not displayed in coin control screen

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/cw_bitcoin/lib/electrum.dart
+++ b/cw_bitcoin/lib/electrum.dart
@@ -235,21 +235,23 @@ class ElectrumClient {
         return [];
       });
 
-  Future<List<Map<String, dynamic>>> getListUnspent(String scriptHash) =>
-      call(method: 'blockchain.scripthash.listunspent', params: [scriptHash])
-          .then((dynamic result) {
-        if (result is List) {
-          return result.map((dynamic val) {
-            if (val is Map<String, dynamic>) {
-              return val;
-            }
+  Future<List<Map<String, dynamic>>> getListUnspent(String scriptHash) async {
+    final result = await call(method: 'blockchain.scripthash.listunspent', params: [scriptHash]);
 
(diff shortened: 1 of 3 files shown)


cw_bitcoin/lib/electrum.dart
```diff
@@ -235,21 +235,23 @@ class ElectrumClient {
...
-          return result.map((dynamic val) {
-            if (val is Map<String, dynamic>) {
-              return val;
-            }
+  Future<List<Map<String, dynamic>>> getListUnspent(String scriptHash) async {
+    final result = await call(method: 'blockchain.scripthash.listunspent', params: [scriptHash]);
 
-            return <String, dynamic>{};
-          }).toList();
+    if (result == null) throw Exception('Failed to get listunspent');
```
OmarHatem28, line 241: this will have side effects
ex:
1. it's called in a `Future.wait`, so, throwing an exception would cause the whole Future.wait to fail
2. the second place it's called, has a try and catch, but it will make it not continue the flow
3. and also would make it report that error to the user through the send error report popup, thus giving the support more unnecessary tickets for connection issues

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -1444,6 +1450,9 @@ abstract class ElectrumWalletBase
         updatedUnspentCoins.add(coin);
       } catch (_) {}
     }));
+    } catch (_) {
+      rethrow;
+    }
```
OmarHatem28, line 1455: I don't think we should rethrow while it's not handled correctly, and will just show the `send error report` popup to the user

OmarHatem28 (changes requested):

OmarHatem28 (approved):

PR #1876, opened 2024-12-13
Generic fixes

tuxpizza:


OmarHatem28 (approved):
I would say `guides` is better than `docs` in this context, but I will leave it up to you, merge it if you see it's good like this

PR #1902, opened 2024-12-26
fix cakepay text encoding

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (commented):
`utf8.decode(utf8.encode(text));`?! encoding then decoding! what would be the purpose?

Serhii-Borodenko:
It looks like the initial issue was that, due to some symbols, Dart decoded the text to Latin initially, which was wrong because the text is actually UTF-8. Forcing the encoding fixed the issue

OmarHatem28 (approved):

PR #1941, opened 2025-01-06
decred: Add used address history. 

JoeGruffins:
Part of #1938 

This adds an address history.  Currently very simple but can be built upon if necessary.

It changes the sync check back to 5 seconds while syncing only. I feel like 30 seconds is rather long in the beginning. Hope this is alright.

It reverts wisdom's pubspec changes, they tell me the downgrade was a problem with their flutter version after all. Also added testnet by them which can be used for testing, I can provide testnet funds please message.

There were some formatting errors apparently, although I thought I had been updating those.


JoeGruffins:
I guess the formatting problem is on our side. I am using `dart format` but I guess this is not correct, what command should I use?

Also, there is currently a problem with wallet groups. Because of putting the birthday in the mnemonic, we are unfortunately not compatible with other wallets atm...

JoeGruffins:
Added another commit. We do use the bip39 word list, which is what I think this is about.

JoeGruffins:
Now formatted with 100 width lines.

JoeGruffins:
I think maybe we don't want to add ourselves to `isBIP39Wallet` so we can stay a "single seed wallet" which solves the encoded date for now.

Also the file `assets/decred_node_list.yml` didn't have any useful info in it and we don't need the nodelist so removed.

JoeGruffins:
Now populating transaction history before sync.

JoeGruffins:
`decred: Fix send all` updates the libwallet and allows for sending all inputs. The bug that may stall sync is also fixed with the dependency update.

JoeGruffins:
In my local commit history I have a 
```
commit db067929d03ad139e94b2df30494f7c133d19172
Author: Czarek Nakamoto <cyjan@mrcyjanek.net>
Date:   Thu Jan 16 11:34:43 2025 +0100

    [skip slack] [run tests] clone by sha
```
But for some reason I do not see it here. My last commit hash is the same though.

JoeGruffins:
I updated the logo colors.

assets/decred_node_list.yml
OmarHatem28, line 1: why was this and it's usage, removed? merge conflict?

JoeGruffins: Sorry, I probably made a mistake. Thanks for fixing.

JoeGruffins:
The last three commits clean up the code a little and ensuring selected addresses are showed on the receive screen.

assets/decred_node_list.yml
JoeGruffins: Oh wait, my bad. The node list was not used. It is only for default nodes correct? The wallet has a list of seeders that it uses to find nodes. This is hardcoded in dcrwallet -> libwallet

JoeGruffins: actually imported from our coin daemon https://github.com/decred/dcrd/blob/172eeb5fe1493e1d5cff6dc304c94de5befd0d21/chaincfg/mainnetparams.go#L576-L581

OmarHatem28: it adds the nodes that will show in the app, I know that the wallet has a list that it will go through, but if we decided to host our node or the user wanted to add his node, he still will be able to connect through it right?

that's why we have the node_list.yml, so we can keep it organized if we want to add a new node that the users can connect to

JoeGruffins: Users can add persistent peers currently, which would be one node to connect to.

We can add it back if you want but its much better to leave finding peers to the wallet software by default. The wallet verifies all headers so there is no need for trust here.

JoeGruffins: These are results from one seeder https://mainnet-seed-1.decred.org/api/addrs

If someone sets a persistent peer, it can no longer use any of those peers, it only connects to the set one.

JoeGruffins:
The last two add back the default node list and adds the ability to generate addresses manually. It also sets the number of unused addrs returned regularly from 1 to 3. Two things that may be wrong still, we can't really tell which are unused. Maybe should add more data eventually. Also the screen that shows all the addresses does not update unless you leave it. Maybe missing something somewhere.

OmarHatem28 (approved):
Looks good overall, will merge it and modify any minor things, thank you 🙏

PR #1950, opened 2025-01-09
CW-519 Enable built-in Tor

MrCyjaneK:
# Description

# Pull Request - Checklist  

- [ ] Format code


cw_bitcoin/lib/electrum.dart
```diff
@@ -73,17 +75,9 @@ class ElectrumClient {
     socket = null;
 
     try {
-      if (useSSL == false || (useSSL == null && uri.toString().contains("btc-electrum"))) {
```
OmarHatem28, line 76: please add this condition for useSSL for now as it's not guaranteed that `useSSL` is set correctly for all nodes

cw_core/lib/node.dart
```diff
@@ -184,23 +184,22 @@ class Node extends HiveObject with Keyable {
     final body = {'jsonrpc': '2.0', 'id': '0', 'method': "getinfo"};
 
     try {
-      final authenticatingClient = HttpClient();
+      final authenticatingClient = ProxyWrapper().getHttpClient();
+
       authenticatingClient.badCertificateCallback =
           ((X509Certificate cert, String host, int port) => true);
```
OmarHatem28, line 190: not used

OmarHatem28 (changes requested):

cw_core/lib/solana_rpc_http_service.dart
```diff
@@ -1,26 +1,25 @@
...
   @override
   final String url;
-  final Client client;
   final Duration defaultRequestTimeout;
 
   @override
   Future<Map<String, dynamic>> call(SolanaRequestDetails params, [Duration? timeout]) async {
-    final response = await client.post(
-      Uri.parse(url),
+    final response = await ProxyWrapper().post(
```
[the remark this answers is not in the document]
MrCyjaneK: you mean to create ProxyWrapper as a part of di and use GetIt to access it?

lib/cake_pay/cake_pay_api.dart
OmarHatem28, line 1: we have tor disabled for CakePay APIs anyway, so instead of showing the send error popup, let's catch that 403 status code, and show a meaningful message to the user

lib/core/open_crypto_pay/open_cryptopay_service.dart
```diff
@@ -5,14 +5,15 @@ import 'package:cake_wallet/core/open_crypto_pay/exceptions.dart';
...
+import 'package:cw_core/utils/proxy_wrapper.dart';
 
 class OpenCryptoPayService {
   static bool isOpenCryptoPayQR(String value) =>
       value.toLowerCase().contains("lightning=lnurl") ||
       value.toLowerCase().startsWith("lnurl");
 
-  final Client _httpClient = Client();
+  // ignore: deprecated_member_use
+  final _httpClient = ProxyWrapper().getHttpClient();
```
OmarHatem28: unused
```suggestion
```

cw_monero/lib/api/wallet.dart
```diff
@@ -228,7 +245,7 @@ void storeSync({bool force = false}) async {
     return monero.Wallet_synchronized(Pointer.fromAddress(addr));
   });
   if (lastStorePointer == addr &&
-      lastStoreHeight + 5000 > currentWallet!.blockChainHeight() &&
+      lastStoreHeight + 75000 > currentWallet!.blockChainHeight() &&
```
OmarHatem28, line 248: why is this change?

lib/view_model/start_tor_view_model.dart
```diff
@@ -0,0 +1,95 @@
...
+    if (didStartTor) {
+      return;
+    }
+    await ensureTorStarted(context: null);
+    while (true) {
+      await Future.delayed(Duration(milliseconds: 250));
+      if (CakeTor.instance.port != -1 && CakeTor.instance.started) {
+        break;
+      }
+    }
```
OmarHatem28, line 65: wouldn't this loop keep working even if the user selects to disableTor?
or is this viewmodel disposed anyway so we are sure that this loop will be closed?

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -493,11 +493,13 @@ abstract class ElectrumWalletBase
...
-            .timeout(Duration(seconds: 5));
-
-        final result = json.decode(response.body) as Map<String, dynamic>;
+        final req = await ProxyWrapper()
+            .getHttpClient()
+            .getUrl(Uri.parse("https://mempool.cakewallet.com/api/v1/fees/recommended"))
+            .timeout(Duration(seconds: 15));
+        final response = await req.close();
+        final stringData = await response.transform(utf8.decoder).join();
+        final result = json.decode(stringData) as Map<String, dynamic>;
```
OmarHatem28: why use a deprecated function?
this should do the same right?
```suggestion
        final response = await ProxyWrapper()
            .get(clearnetUri: Uri.parse("https://mempool.cakewallet.com/api/v1/fees/recommended"))
            .timeout(Duration(seconds: 15));
        final result = json.decode(response.body) as Map<String, dynamic>;
```

OmarHatem28: same for other 3 places in this file

lib/utils/image_utill.dart
```diff
@@ -1,13 +1,16 @@
+import 'package:cw_core/utils/proxy_wrapper.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_svg/svg.dart';
 
 class ImageUtil {
   static Widget getImageFromPath({required String imagePath, double? height, double? width}) {
+    if (CakeTor.instance.enabled) {
+      imagePath = "assets/images/tor_logo.svg";
+    }
```
OmarHatem28: move it below the next bool `isNetworkImage` and add it only when it's a network image

cw_core/lib/utils/tor/android.dart
OmarHatem28, line 1: since there is no file for iOS, does this entail it won't work on iOS?

lib/utils/tor.dart
```diff
@@ -0,0 +1,80 @@
...
+  await CakeTor.instance.start();
+  printV("Tor started");
+  while (!CakeTor.instance.started) {
+    printV("Waiting for tor to start (part 1)");
+    await Future.delayed(const Duration(seconds: 1));
+  }
+  while (CakeTor.instance.port == -1) {
+    printV("Waiting for tor to start (listening on port)");
+    await Future.delayed(const Duration(seconds: 1));
+  }
```
OmarHatem28, line 46: shouldn't we put a timeout or something so it doesn't continue in an infinite loop if tor didn't work?

cw_monero/lib/api/wallet.dart
MrCyjaneK: To make it a bit less frequent - 5000 is a bit overkill, especially noticeable with huge slowdown with sync over tor

cw_core/lib/utils/tor/android.dart
MrCyjaneK: Yes, though this should be rather easy - just building tor binary for iOS and starting it - that being said probably better implement it later, or revisit the way we do it as even Onion Browser now require 3rd party app to connect to Tor.

lib/utils/tor.dart
MrCyjaneK: No need to, function should only return if tor state changed, if loop doesn't return during app start there is UI element to disable tor / wait in background.

lib/view_model/start_tor_view_model.dart
MrCyjaneK: if it hangs for too long it will allow user to take action (disable Tor / ignore the wait) - the loop will run until Tor is enabled though, but unless Tor would be completely broken it should never happen - this loop makes sure that tor gets as much time as it wants to start properly

OmarHatem28 (approved):

pubspec_base.yaml
```diff
@@ -162,10 +166,20 @@ dependency_overrides:
       url: https://github.com/cake-tech/bitcoin_base
       ref: cake-update-v9
   ffi: 2.1.0
+  web_socket_channel: ^3.0.2
+  freezed_annotation: 2.4.4
```
OmarHatem28:
```suggestion
```

PR #1999, opened 2025-02-03
fix tests not working in latest main

MrCyjaneK:
# Description

Caused by new UI

# Pull Request - Checklist  

- [ ] Format code


MrCyjaneK:
disable exchange_flow_test.dart - it fails mostly because of API keys. I cannot fix or debug it locally - that's why I disabled it.

PR #2038, opened 2025-02-18
Cw 939 whitelist known tokens

fossephate:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/view_model/dashboard/home_settings_view_model.dart
```diff
@@ -189,6 +188,40 @@ abstract class HomeSettingsViewModelBase with Store {
...
+      case WalletType.polygon:
+        defaultTokenAddresses = polygon!.getDefaultTokenContractAddresses();
+        break;
+      case WalletType.solana:
+        defaultTokenAddresses = solana!.getDefaultTokenContractAddresses();
+        break;
+      case WalletType.tron:
+        defaultTokenAddresses = tron!.getDefaultTokenContractAddresses();
+        break;
+      case WalletType.zano:
```
OmarHatem28, line 207: like this, any zano asset will be marked as a scam, let's think of a way to handle that other than ignoring just for zano

fossephate: we only mark it as a potential scam if the token address isn't in the whitelist and the token is that of a base currency:
```
if (baseCurrencySymbols.contains(tokenSymbol) && !isWhitelisted) {
    isPotentialScam = true;
}
```

OmarHatem28 (approved):

lib/view_model/dashboard/home_settings_view_model.dart
OmarHatem28: oh right, my bad, all good then, will wait on testers to verify and we can merge it

PR #2040, opened 2025-02-20
Cw 855 transactions not cleared correctly when switching wallets

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (commented):
there is another `connectMapToListWithTransform` in the constructor so probably you should convert it too
but I want to make sure that this approach will indeed trigger the reaction for all change types like it's currently doing in `connectMapToListWithTransform`
compare the code in there and see if it will achieve the same or would it have any side effect

Serhii-Borodenko:
I have converted the connectMapToListWithTransform to a reaction in the constructor as well. I’ve tested adding, removing, and updating transactions, and it appears to work as expected.
Although the MobX documentation isn’t very detailed on this matter, my understanding is that using .values.toList() on an observable list (or map) creates a new list instance—but MobX still registers dependencies on all observables accessed within the lambda. This means the reaction will be triggered for all change types (add, remove, update), similarly to how connectMapToListWithTransform works, without introducing side effects.

OmarHatem28 (approved):

PR #2046, opened 2025-02-23
Update Italian (IT) translations

saloniamatteo:
Greetings,
this PR fixes many translation errors included within `strings_it.arb`.

All 1000 string entries were checked against the upstream `strings_en.arb` file for any obvious changes. For some reason some string entries were in Croatian (is the OG team croatian? ;P), so those were translated as well.

Please contact me if any issues arise!


OmarHatem28 (approved):
Looking good, thanks for contributing 🙏

Issue #986, opened 2023-07-04
multi: Add decred.


JoeGruffins:
I'm looking into the possibility of adding [decred](https://decred.org/) to cake wallet. Would this be supported by the cake wallet maintainers? I can do all of the work of adding but require review and merging ofc. Will add to this issue with a more fleshed out plan if the endeavor will be supported.

SamsungGalaxyPlayer:
Hi Joe, please email justin@cakewallet.com and we can chat about it later this week.

JoeGruffins:
#1322 Adds decred. This is meant to add everything we need. There are still a few things we want to update after this.
 
 TODO list:

- [ ] Make the UI around persistent peers better. It looks like you can set multiple but currently it will only look at the last set I guess. We don't need one set at all but cake seems to barf if you have none, not saving the wallet at all. So it's currently just a port if we are letting the libwallet find peers for us.
- [ ]  We need to add address discover as a sync topic because this can take a while. Should add fetching headers and rescanning as well.
- [ ] We don't want to give out an address before sync is complete, so need to disable the receive screen or something until synced. Currently it returns nothing and shows the default screen.
- [ ] We don't want the user to go to the send screen when using a watching only wallet.
- [ ] There needs to be a way to tell that one is using a watching only wallet rather than a "normal" one.
- [ ] The next wallet version will have the ability to sync from a birthday, so the initial sync time will be greatly reduced. We will want to wait for a dcrwallet release so it may be a bit before we can use it here though.
- [ ] Related to the above but rescans should just start from the wallet birthday. The current date -> block method is temporary and not precise.

JoeGruffins:
Some of the solutions may be debatable.
- Make the UI around persistent peers better.
It now says default-spv-peers which is fine I think.

- We need to add address discover as a sync topic
Says "Processing" now, which is better than nothing

- There needs to be a way to tell that one is using a watching only wallet
Sync is greyed out.

- The wallet groups do not work with decred's mnemonic
This is just disabled for decred now, which is fine imo. We want users to use the 15 word seed because it contains the birthday.

PR #2067, opened 2025-03-05
decred: Close libwallet when app paused.

JoeGruffins:
Leaving the libwallet running causes a crash when coming back from sleeping on mobile. This prevents that, however it also stops sync and all other wallet functions. The wallet only works while in the foreground.

Part of #986

### DIFF
--- a/cw_decred/lib/wallet.dart
+++ b/cw_decred/lib/wallet.dart
@@ -26,6 +26,7 @@ import 'package:cw_core/sync_status.dart';
 import 'package:cw_core/node.dart';
 import 'package:cw_core/unspent_coins_info.dart';
 import 'package:cw_core/unspent_transaction_output.dart';
+import 'package:cake_wallet/src/screens/root/root.dart';
 
 part 'wallet.g.dart';
 
@@ -70,7 +71,6 @@ abstract class DecredWalletBase
   // synced is used to set the syncTimer interval.
   bool synced = false;
   bool watchingOnly;
-  bool connecting = false;
   String persistantPeer = "default-spv-nodes";
   FeeCache feeRateFast = FeeCache(defaultFeeRate);
   FeeCache feeRateMedium = FeeCache(defaultFeeRate);
@@ -98,6 +98,8 @@ abstract class DecredWalletBase
     if (watchingOnly) {
       return null;
     }
+    // TODO: Remove when sleep panic bug is fixed.
+    loadWallet();
     return libdcrwallet.walletSeed(walletInfo.name, _password);
   }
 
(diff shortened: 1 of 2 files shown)


MrCyjaneK:
hey @JoeGruffins

Do we know the reason for which it crashes? Our current mweb implementation is written in Golang and it works just fine (in both foreground and background). I am aware that OS tends to harm processes in many ways (usually by restricting network, freezing the process or just straight up killing the entire app), but OS shouldn't under any circumstances free actively allocated memory, if that memory is really being allocated to something else then most likely Go GC freed the memory.

Couple blind guesses/thoughts that I have are
- Maybe we start the process in an isolate that we later close - thus freeing the memory?
- Wallet is opened in a goroutine and pointer is only kept on dart side so Go GC just free that memory because it doesn't remember it?
- It should be possible to check for accidental GC by doing `debug.SetGCPercent(-1); debug.SetMemoryLimit(math.MaxInt64)` in Go.

Could you share stacktrace of that crash?

Also according to the comment the app sometimes won't wake up, and I believe that is pretty bad because many times when paying online one switches between a web browser and wallet multiple times, with this workaround there is a risk that people wouldn't be able to spend funds reliably in any way other than clicking a link or scanning QR code.

JoeGruffins:
>stacktrace

This is all I've been able to get so far, from logcat. Unsure how to get a better log atm. https://pastebin.com/3PXQ0c0u

JoeGruffins:
I tried putting `debug.SetGCPercent(-1); debug.SetMemoryLimit(math.MaxInt64)` in main in our cgo code and it still crashed.

Looking now, we are freeing the arguments for libwallet (I think) but not the returns. I wonder if that could be part of it.

MrCyjaneK:
I'm currently trying to run decred from `decred-updated-main`, I'm putting the app in the background and it doesn't crash on my end. Could you please share steps to reproduce? Wallet does disconnect on it's own when in background (and connects back once I put it back in the foreground)

JoeGruffins:
So... its real annoying to cause so far for me. The most surefire way seems to be to let the app sit open for ~10 minutes. After the screen goes dark and you have to open your phone again, coming back from that usually causes it, but not always. So, its a longer sleep, usually.

JoeGruffins:
The timer to detect if resumed was also janky. This is the best patch I can do. Will continue looking into the bug but maybe this is ok for now.

JoeGruffins:
Trying your first suggestion #2077

Still testing to see if the bug is fixed, but may be a good change regardless.

JoeGruffins:
I think I finally found the reason, and it is a problem with the library. Will probably close this and update the deps once we have the change solidified.

JoeGruffins:
Yeah, I'm pretty confident to have found the issue, closing this.

PR #2070, opened 2025-03-06
Update InfoBase.plist

tuxpizza:
Add CADisableMinimumFrameDurationOnPhone to unlock refresh rate for iPhones with ProMotion display


MrCyjaneK (approved):
Made a built, tested it with @tuxpizza and it works great.

PR #2072, opened 2025-03-07
CW-901 add app logo to center of QR code

MrCyjaneK:
# Description

While this PR works there are some things that we need to take care of

- [ ] Different icon (with background) so it will look nicer
- [ ] Use a lower resolution image to prevent lagging when creating QR code
    - [ ] If that doesn't help then maybe cache the QR code widgets

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28:
Please add a white background to it if possible from the code.
If not, then let Tux know, and he will get you a suitable one

OmarHatem28 (changes requested):
it should use Monero.com logo for monero.com app

PR #2106, opened 2025-03-22
decred: Do not create log file. 

JoeGruffins:


cw_decred/lib/wallet.dart
```diff
@@ -380,7 +380,7 @@ abstract class DecredWalletBase
         totalAmt = totalIn;
       } else if (out.cryptoAmount != null) {
         final coins = double.parse(out.cryptoAmount!);
-        amt = (coins * 1e8).toInt();
```
JoeGruffins, line 383: This doesn't seem to be as accurate so changed to `round`
The following will print 29999
```
    final coins = .0003;
    final thing = (coins * 1e8).toInt();
    print("${thing}");
```

cw_decred/lib/wallet_service.dart
```diff
@@ -27,17 +27,16 @@ class DecredWalletService extends WalletService<
   static final pubkeyRestorePathTestnet = "m/44'/1'/0'";
   final mainnet = "mainnet";
   final testnet = "testnet";
-  Libwallet? libwallet;
```
JoeGruffins, line 30: Was seeing the error
```
I/flutter (20589): cw_decred/api/libdcrwallet.dart#317:9 Libwallet: decred libwallet returned an error for method initlibdcrwallet: Exception: duplicate initialization
```

But making this static looks to fix that.

Issue #2120, opened 2025-03-25
large tx history wallet causes bad UI lags / stutters


nahuhh:
**To Reproduce**

Restore CCS view wallet & scroll

**Platform:**
 - OS: android
 - Device: [e.g. iPhone 14, Galaxy S21]
 - Cake Wallet Version: 1.21.0

PR #2128, opened 2025-03-27
CW-1024 Improve performance of xmr wallet

MrCyjaneK:
- enabled development options in CI builds.
- Implemented caching for account retrieval.
- refactor transaction handling in `dashboard_view_model.dart` to improve efficiency and reduce unnecessary updates in xmr.
- `DevMoneroCallProfilerPage`, for profiling performance of xmr,wow,zano wallet calls.

Issue Number (if Applicable): Fixes https://github.com/cake-tech/cake_wallet/issues/2120

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_monero/lib/api/transaction_history.dart
+++ b/cw_monero/lib/api/transaction_history.dart
@@ -42,6 +42,7 @@ Future<void> refreshTransactions() async {
   await Isolate.run(() {
     monero.TransactionHistory_refresh(Pointer.fromAddress(ptr));
   });
+  await Future.delayed(Duration.zero);
   txHistoryMutex.release();
   isRefreshingTx = false;
 }
@@ -54,7 +55,18 @@ Future<List<Transaction>> getAllTransactions() async {
   await txHistoryMutex.acquire();
   txhistory ??= monero.Wallet_history(wptr!);
   int size = countOfTransactions();
-  final list = List.generate(size, (index) => Transaction(txInfo: monero.TransactionHistory_transaction(txhistory!, index: index)));
+  final list = <Transaction>[];
+  for (int index = 0; index < size; index++) {
+    if (index % 25 == 0) {
+      // Give main thread a chance to do other things.
+      await Future.delayed(Duration.zero);
+    }
+    final txInfo = monero.TransactionHistory_transaction(txhistory!, index: index);
+    final txHash = monero.TransactionInfo_hash(txInfo);
+    txCache[wptr!.address] ??= {};
+    txCache[wptr!.address]![txHash] = Transaction(txInfo: txInfo);
+    list.add(txCache[wptr!.address]![txHash]!);
+  }
   txHistoryMutex.release();
   final accts = monero.Wallet_numSubaddressAccounts(wptr!);
   for (var i = 0; i < accts; i++) {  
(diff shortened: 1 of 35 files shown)


cw_monero/lib/monero_unspent.dart
```diff
@@ -7,27 +7,29 @@ class MoneroUnspent extends Unspent {
...
-    }
-  }
-
-  @override
-  bool get isFrozen {
-    printV("get isFrozen");
-    final coinId = getCoinByKeyImage(keyImage!);
-    if (coinId == null) throw Exception("Unable to find a coin for address $address");
-    final coin = getCoin(coinId);
-    return monero.CoinsInfo_frozen(coin);
```
OmarHatem28, line 30: Shouldn't this still exist, but returning the new `_frozen` variable you made?

cw_monero/lib/monero_unspent.dart
```diff
@@ -7,27 +7,29 @@ class MoneroUnspent extends Unspent {
   MoneroUnspent(
       String address, String hash, String keyImage, int value, bool isFrozen, this.isUnlocked)
       : super(address, hash, value, 0, keyImage) {
+    getCoinByKeyImage(keyImage).then((coinId) {
+      if (coinId == null) throw Exception("Unable to find a coin for address $address");
```
OmarHatem28, line 11: is this exception gracefully handled? or would it cause a gray screen or so?

cw_zano/devtools_options.yaml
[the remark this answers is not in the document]
MrCyjaneK: I'm using them all the time, but I don't commit the config file, let me correct that

cw_monero/lib/monero_unspent.dart
MrCyjaneK: the behavior is more or less the same, except it would throw it in constructor instead of later in getter.

cw_monero/lib/monero_wallet.dart
```diff
@@ -569,6 +570,7 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
     walletInfo.restoreHeight = height;
     walletInfo.isRecovery = true;
     monero_wallet.setRefreshFromBlockHeight(height: height);
+    setupBackgroundSync(password, wptr!);
```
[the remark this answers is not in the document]
MrCyjaneK: yes, otherwise restore height won't get stored in background cache and it will get overwritten the next time user opens wallet.

cw_monero/lib/monero_wallet.dart
```diff
@@ -709,8 +712,17 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
...
+      final currentIds = transactionHistory.transactions.keys.toSet();
+      final newIds = transactions.keys.toSet();
+      
+      // Remove transactions that no longer exist
+      currentIds.difference(newIds).forEach((id) => 
+          transactionHistory.transactions.remove(id));
+      
+      // Add or update transactions
+      transactions.forEach((key, tx) => 
+          transactionHistory.transactions[key] = tx);
```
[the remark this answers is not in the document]
MrCyjaneK: clearing and adding them again is what caused updated _transactionDisposer to freak out, when we update we trigger rebuild only when needed

cw_monero/lib/monero_unspent.dart
MrCyjaneK: correct, I've missed that

cw_monero/lib/monero_unspent.dart
MrCyjaneK: ah no wait

cw_monero/lib/monero_wallet.dart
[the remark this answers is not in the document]
OmarHatem28: but if no transaction was removed, but some got updated, then I believe it won't trigger the rebuild, right?
but when we were doing clear then addAll, it will trigger a rebuild

cw_monero/lib/api/transaction_history.dart
```diff
@@ -13,15 +14,23 @@ import 'package:monero/src/generated_bindings_monero.g.dart' as monero_gen;
...
+  txKeys[wptr!.address] ??= {};
+  if (txKeys[wptr!.address]![txId] != null) {
+    return txKeys[wptr!.address]![txId]!;
+  }
   final txKey = monero.Wallet_getTxKey(wptr!, txid: txId);
   final status = monero.Wallet_status(wptr!);
   if (status != 0) {
-    final error = monero.Wallet_errorString(wptr!);
+    monero.Wallet_errorString(wptr!);
+    txKeys[wptr!.address]![txId] = "";
```
OmarHatem28, line 27: shouldn't we try again the next time?
like this if this errors out, it will cache it as empty string, thus it will be always empty, while it should try again every time

MrCyjaneK: If we don't have the key it won't change in the future, this is to save times and not log too much to console.

MrCyjaneK: We call these functions to not leak errors in other places

PR #2165, opened 2025-04-06
add/remove coins from readme & matic -> pol on add wallet page

nahuhh:

# Description

- Readme was missing ckme coins, and still listed haven
- The add wallet screen still listed polygon as matic

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,13 @@ Cake Wallet includes support for several cryptocurrencies, including:
 * Ethereum (ETH)
 * Litecoin (LTC)
 * Bitcoin Cash (BCH)
-* Polygon (Pol)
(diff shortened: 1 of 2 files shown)


nahuhh:
I only touched the obvious places.

i think trocador still needs matic -> POL, and there is a bunch of dead haven code that could use a clean up `grep -r haven`

OmarHatem28:
will re-check with our providers on whether they migrated to POL or not, because most of them at that time still used matic.

For Haven, yeah, some code is still there, so we don't break the app for people who have an already existing Haven wallet, and it should just fail to open it instead of crashing the whole app

PR #2172, opened 2025-04-07
Cw 1032 verify on ramper integration

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/buy/onramper/onramper_buy_provider.dart
+++ b/lib/buy/onramper/onramper_buy_provider.dart
@@ -201,7 +201,6 @@ class OnRamperBuyProvider extends BuyProvider {
       required String cryptoCurrencyAddress,
       String? countryCode}) async {
     final actionType = isBuyAction ? 'buy' : 'sell';
-    final prefix = actionType == 'sell' ? actionType + '_' : '';
 
     final primaryColor = getColorStr(Theme.of(context).primaryColor);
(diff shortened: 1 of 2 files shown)


lib/buy/onramper/onramper_buy_provider.dart
```diff
@@ -276,6 +279,10 @@ class OnRamperBuyProvider extends BuyProvider {
   String _getNormalizeCryptoCurrency(Currency currency) {
     if (currency is CryptoCurrency) {
       if (!mainCurrency.contains(currency)) {
+
+        final networkByCurrency = _getNetworkByCrypto(currency);
+        if (networkByCurrency != null) return '${currency.title}_$networkByCurrency';
```
OmarHatem28, line 284: so this will be only the case of `CryptoCurrency.eos`?

can we re-write this function in a way that is more readable and friendly, even if it will duplicate some code

also doesn't this code mean that in the case of Monero it will return `XMR_MONERO` is this the expected value?
because for Bitcoin it will return just `BTC`

Serhii-Borodenko: Did some refactoring. Yes, for Monero the expected value is XMR_MONERO.

OmarHatem28 (approved):

PR #2180, opened 2025-04-09
 ui: show EUR offers for SEPA currencies on onramping screen

MrCyjaneK:
# Description

# Pull Request - Checklist  

- [ ] Format code


lib/view_model/buy/buy_sell_view_model.dart
```diff
@@ -149,13 +149,12 @@ abstract class BuySellViewModelBase extends WalletChangeListenerViewModel with S
...
   bool get isReadyToTrade {
-    final hasSelectedQuote = selectedQuote != null;
+    // final hasSelectedQuote = selectedQuote != null;
     final hasSelectedPaymentMethod = selectedPaymentMethod != null;
     final isPaymentMethodLoaded = paymentMethodState is PaymentMethodLoaded;
     final isBuySellQuotLoaded = buySellQuotState is BuySellQuotLoaded;
 
-    return hasSelectedQuote &&
-        hasSelectedPaymentMethod &&
+    return hasSelectedPaymentMethod &&
```
OmarHatem28, line 157: why this change? it shouldn't be ready until the user selects a quote

OmarHatem28 (changes requested):

lib/view_model/buy/buy_sell_view_model.dart
```diff
@@ -409,44 +446,67 @@ abstract class BuySellViewModelBase extends WalletChangeListenerViewModel with S
...
     }
 
-    validQuotes.sort((a, b) => a.rate.compareTo(b.rate));
+    validQuotesNative.sort((a, b) => a.rate.compareTo(b.rate));
+    validQuotesSepa.sort((a, b) => a.rate.compareTo(b.rate));
 
     final Set<String> addedProviders = {};
-    final List<Quote> uniqueProviderQuotes = validQuotes.where((element) {
+
+    final List<Quote> uniqueProviderQuotesNative = validQuotesNative.where((element) {
```
OmarHatem28, line 486: and I feel like there is something missing in this part and the one below, but I am not sure, so I will let @Serhii-Borodenko chime in

lib/view_model/buy/buy_sell_view_model.dart
MrCyjaneK: With this change we do not have any details regarding the quote because we don't select the quote by default from EUR-only providers (if the currency is SEPA enabled), as this would be unreliable. We just do not calculate anything in terms of price because
- provider cannot give use price in local currency
- we cannot calculate the price on our own because we do not include SEPA fees and provider fees, so using for example priceapi wouldn't work.

Yet we still need to mark the trade as ready to allow the user to click the button - so if there is any quote (even in foreign currency) we unlock the button

lib/view_model/buy/buy_sell_view_model.dart
Serhii-Borodenko: So far, the only issue I see is with the “Recommended” section. With the current implementation, only successRateQuotes are added there, which means only Onramper quotes with the successRate label. And if we don’t want to duplicate providers that already have native quotes with SEPA quotes in the ‘Recommended’ section, then we don’t need uniqueProviderQuotesSepa

MrCyjaneK: Yeah it is messed up, I'll resolve this

OmarHatem28:
@MrCyjaneK what's the state of this?

kathleenmotley49-maker (approved):

PR #2235, opened 2025-04-29
Cw 935 get exchange payment method recommendations from on ramper

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/buy/onramper/onramper_buy_provider.dart
```diff
@@ -71,9 +73,14 @@ class OnRamperBuyProvider extends BuyProvider {
       if (response.statusCode == 200) {
         final Map<String, dynamic> data = jsonDecode(response.body) as Map<String, dynamic>;
         final List<dynamic> message = data['message'] as List<dynamic>;
-        return message
+
+        final allAvailablePaymentMethods = message
             .map((item) => PaymentMethod.fromOnramperJson(item as Map<String, dynamic>))
             .toList();
+
+        recommendedPaymentType = allAvailablePaymentMethods.first.paymentMethodType;
```
OmarHatem28, line 81: if the recommended is the first avilable payment method, then what's customPaymentMethod? why is it returned separately?

OmarHatem28 (commented):
also is the switch from defaulting to `PaymentType.all` to `PaymentType.unknown` handled everywhere?

lib/buy/onramper/onramper_buy_provider.dart
Serhii-Borodenko: At first, I thought the first available payment method was the recommended one. It worked for most locations I tested, but it turned out not to work for all, so I had to add an extra API call to getDefaults to fetch the recommended method.

Serhii-Borodenko:
I hardcoded some common payment methods used by most providers, like Google Pay, SEPA, and credit card. But, most of Onramper’s payment methods are unique, which is why I’m using PaymentType.unknown along with customPaymentMethod.

If we don’t map the fetched method to one of the predefined options, we store it as a string in customPaymentMethod, display it to the user, and pass it as a parameter if the user selects it. It works for Onramper and Moonpay.

OmarHatem28 (approved):

lib/buy/onramper/onramper_buy_provider.dart
```diff
@@ -131,13 +167,13 @@ class OnRamperBuyProvider extends BuyProvider {
...
       String? countryCode}) async {
     String? paymentMethod;
 
-    if (paymentType != null && paymentType != PaymentType.all) {
-      paymentMethod = normalizePaymentMethod(paymentType);
-      if (paymentMethod == null) paymentMethod = paymentType.name;
-    }
+    if (paymentType == PaymentType.all && recommendedPaymentType != null) paymentMethod = recommendedPaymentType!;
+    else if (paymentType == PaymentType.unknown) paymentMethod = customPaymentMethodType;
+    else if (paymentType != null) paymentMethod = normalizePaymentMethod(paymentType);
```
OmarHatem28, line 176: if payment type is credit card for example. (this is what the user selected)

the first if condition is going to be false.
the second one would probably be true if we already have a custom payment type available, and thus we will set the `paymentMethod` with the custom one, instead of the credit card the user selected

Serhii-Borodenko: In the case of a payment method like credit card, there’s a specific `PaymentType.creditCard,` so this condition (paymentType == PaymentType.unknown) will be false.

PR #2247, opened 2025-05-02
CW-959: Swap Status on Transaction Screen

Blazebrain:
- Introduce SwapManager for automatic tracking of active-wallet swaps.
- Automatically queues new or updated trades from the Hive box.
- Periodically fetch and persist swap statuses via the corresponding trade provider.
- Implement start(wallet, providers), stop(), and dispose() for lifecycle control.
- Apply user's ExchangeApiMode(disabled, tor-only, enabled) when fetching updates.
- Remove swaps from the watchlist on any final state (completed, expired, failed).
- Dispose SwapManager in AppState.dispose() to cancel polling and the Hive subscription.

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/di.dart
```diff
@@ -497,8 +498,30 @@ Future<void> setup({
...
+  getIt.registerSingleton(
+    SwapManager(
+      tradesStore: getIt.get<TradesStore>(),
+      settingsStore: getIt.get<SettingsStore>(),
+    ),
+  );
+
   getIt.registerFactory(() => DashboardViewModel(
       balanceViewModel: getIt.get<BalanceViewModel>(),
+      exchangeViewModel: getIt.get<ExchangeViewModel>(),
```
OmarHatem28, line 523: this will trigger the functions and API calls of the exchange view model while it's not needed, why even use ExchangeViewModel, you already have each provider that created a swap, so just initiate this provider when you find a trade for it, and check its status

lib/di.dart
```diff
@@ -497,8 +498,30 @@ Future<void> setup({
...
+      getIt.get<FeesViewModel>(),
+    ),
+  );
+
+  getIt.registerSingleton(
+    SwapManager(
+      tradesStore: getIt.get<TradesStore>(),
+      settingsStore: getIt.get<SettingsStore>(),
+    ),
+  );
```
OmarHatem28, line 519: I don't see a need for this also, just go through the trades when listing them and if the trade was created in less than 24 hours, then initiate its provider and get its status and add a timer for it until its creation time exceeds 24 hours or its status is completed (`success`, `expired`, `failed`, etc....)

OmarHatem28 (changes requested):
will not continue the review as the solution might be way simpler and less API calls aggressive

lib/core/trade_monitor.dart
```diff
@@ -0,0 +1,171 @@
...
+
+  void _startTradeMonitoring(Trade trade) {
+    if (_tradeTimers.containsKey(trade.id)) return;
+
+    final provider = _getProviderByDescription(trade.provider);
+    if (provider == null) {
+      printV('No provider found for trade ${trade.id}');
+      return;
+    }
+
```
OmarHatem28, line 95: you are already checking that in `_checkTradeStatus`
```suggestion
```

lib/core/trade_monitor.dart
```diff
@@ -0,0 +1,171 @@
...
+        ..stateRaw = updated.state.raw
+        ..receiveAmount = updated.receiveAmount
+        ..outputTransaction = updated.outputTransaction;
+      printV('Trade ${trade.id} updated: ${trade.state}');
+      await trade.save();
+
+      if (_isFinalState(updated.state)) {
+        printV('Trade ${trade.id} is in final state');
+        _cancelTradeTimer(trade.id);
+      }
```
OmarHatem28, line 137: here you only cancel it when it gets to a final state, but we should also cancel if it exceeds 24 hours.

lib/main.dart
```diff
@@ -281,6 +282,12 @@ class App extends StatefulWidget {
 }
 
 class AppState extends State<App> with SingleTickerProviderStateMixin {
+  @override
+  void dispose() {
+    getIt.get<TradeMonitor>().dispose();
+    super.dispose();
+  }
```
OmarHatem28, line 289: I believe this will only get called when closing the app, and if so then it doesn't make sense to dispose anything then, it will already be disposed automatically?

Unless, this is getting called when switching wallets, is that the case? and if so, did we test if there would be a race condition of whether the timers of the new wallet would get registered before this function is called, and then this will remove the timers?

lib/core/trade_monitor.dart
Blazebrain: Alright, I'll add that check

lib/core/trade_monitor.dart
```diff
@@ -0,0 +1,171 @@
...
+      final trade = item.trade;
+
+      if (trade.walletId != walletId) continue;
+
+      final createdAt = trade.createdAt;
+      if (createdAt == null) continue;
+
+      if (now.difference(createdAt).inHours > 24) continue;
+
+      if (_isFinalState(trade.state)) continue;
```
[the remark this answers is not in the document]
Blazebrain: The timers have not been created here, timers get created when a trade passes all these checks

But for trades from other wallets, we could add a general dispose for trades in the outgoing wallet, I'll modify that

OmarHatem28 (approved):
the trade monitor can get some enhancements for readability and easier maintainability, but let's make that in a separate PR

PR #2261, opened 2025-05-13
feat: bitcoin view-only

MrCyjaneK:
# Description

Adds view-only support for bitcoin wallets

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_bitcoin/lib/electrum_wallet.dart
+++ b/cw_bitcoin/lib/electrum_wallet.dart
@@ -280,11 +280,26 @@ abstract class ElectrumWalletBase
   }
 
   @override
-  BitcoinWalletKeys get keys => BitcoinWalletKeys(
-        wif: WifEncoder.encode(hd.privateKey.raw, netVer: network.wifNetVer),
-        privateKey: hd.privateKey.toHex(),
-        publicKey: hd.publicKey.toHex(),
-      );
+  BitcoinWalletKeys get keys {
+    String? wif;
+    String? privateKey;
+    String? publicKey;
+    try {
+      wif = WifEncoder.encode(hd.privateKey.raw, netVer: network.wifNetVer);
+    } catch (_) {}
+    try {
+      privateKey = hd.privateKey.toHex();
+    } catch (_) {}
+    try {
+      publicKey = hd.publicKey.toHex();
+    } catch (_) {}
+    return BitcoinWalletKeys(
+      wif: wif ?? '',
+      privateKey: privateKey ?? '',
+      publicKey: publicKey ?? '',
+      xpub: xpub,
+    );
+  }
 
   String _password;
   List<BitcoinUnspent> unspentCoins;
--- a/lib/view_model/wallet_keys_view_model.dart
+++ b/lib/view_model/wallet_keys_view_model.dart
@@ -1,3 +1,4 @@
+import 'package:cake_wallet/bitcoin/bitcoin.dart';
 import 'package:cake_wallet/generated/i18n.dart';
 import 'package:cake_wallet/monero/monero.dart';
 import 'package:cake_wallet/src/screens/transaction_details/standart_list_item.dart';
@@ -162,18 +163,21 @@ abstract class WalletKeysViewModelBase with Store {
       case WalletType.bitcoin:
       case WalletType.litecoin:
       case WalletType.bitcoinCash:
+          final keys = bitcoin!.getWalletKeys(_appStore.wallet!);
+        
+          items.addAll([
+            if ((keys['wif']??'').isNotEmpty)
+              StandartListItem(title: "WIF", value: keys['wif']!),
+            if ((keys['privateKey']??'').isNotEmpty)
+              StandartListItem(title: S.current.private_key, value: keys['privateKey']!),
+            if (keys['publicKey'] != null)
+              StandartListItem(title: S.current.public_key, value: keys['publicKey']!),
+            if (keys['xpub'] != null)
+              StandartListItem(title: "xPub", value: keys['xpub']!),
+          ]);
+        break;
       case WalletType.none:
       case WalletType.haven:
-        //   final keys = bitcoin!.getWalletKeys(_appStore.wallet!);
-        //
-        //   items.addAll([
-        //     if (keys['wif'] != null)
-        //       StandartListItem(title: "WIF", value: keys['wif']!),
-        //     if (keys['privateKey'] != null)
-        //       StandartListItem(title: S.current.private_key, value: keys['privateKey']!),
-        //     if (keys['publicKey'] != null)
-        //       StandartListItem(title: S.current.public_key, value: keys['publicKey']!),
-        //   ]);
         break;
     }
 
--- a/lib/view_model/wallet_restore_view_model.dart
+++ b/lib/view_model/wallet_restore_view_model.dart
@@ -32,17 +32,7 @@ abstract class WalletRestoreViewModelBase extends WalletCreationVM with Store {
   WalletRestoreViewModelBase(AppStore appStore, WalletCreationService walletCreationService,
       Box<WalletInfo> walletInfoSource, SeedSettingsViewModel seedSettingsViewModel,
       {required WalletType type, this.restoredWallet})
-      : hasSeedLanguageSelector =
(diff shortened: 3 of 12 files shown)


cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -1105,7 +1120,8 @@ abstract class ElectrumWalletBase
           network: network,
           hasChange: estimatedTx.hasChange,
           isSendAll: estimatedTx.isSendAll,
-          hasTaprootInputs: false, // ToDo: (Konsti) Support Taproot
+          hasTaprootInputs: false, // ToDo: (Konsti) Support Taproot,
+          isViewOnly: false,
```
OmarHatem28, line 1131: technically, HWW is also a view only wallet, so we might want to discuss the flow and unify both view only and HWW flow into one

lib/view_model/wallet_restore_view_model.dart
```diff
@@ -82,9 +72,34 @@ abstract class WalletRestoreViewModelBase extends WalletCreationVM with Store {
...
+    WalletType.nano,
+    WalletType.banano,
+    WalletType.solana,
+    WalletType.tron
+  ].contains(type);
+
+  late final bool onlyViewKeyRestore = [
+    WalletType.bitcoin,
+    WalletType.litecoin,
+    WalletType.bitcoinCash,
```
OmarHatem28: not yet implemented for them.
```suggestion
```

lib/view_model/wallet_keys_view_model.dart
```diff
@@ -162,18 +163,21 @@ abstract class WalletKeysViewModelBase with Store {
...
+        
+          items.addAll([
+            if ((keys['wif']??'').isNotEmpty)
+              StandartListItem(title: "WIF", value: keys['wif']!),
+            if ((keys['privateKey']??'').isNotEmpty)
+              StandartListItem(title: S.current.private_key, value: keys['privateKey']!),
+            if (keys['publicKey'] != null)
+              StandartListItem(title: S.current.public_key, value: keys['publicKey']!),
+            if (keys['xpub'] != null)
+              StandartListItem(title: "xPub", value: keys['xpub']!),
```
OmarHatem28, line 176: since we are using the native segwit derivation path (84), wouldn't that mean that BDK would generate a zpub not an xpub?

cw_bitcoin/lib/bitcoin_wallet.dart
```diff
@@ -364,14 +368,19 @@ abstract class BitcoinWalletBase extends ElectrumWallet with Store {
         outputOrdering: BitcoinOrdering.none,
         enableRBF: true,
         publicKeys: tx.publicKeys!,
-        masterFingerprint: Uint8List(0));
+        masterFingerprint: Uint8List.fromList([0, 0, 0, 0]));
```
OmarHatem28, line 381: Previously, it was considerably not provided, but now it defaults to a specific 0 value, Why is this change?

cw_bitcoin/lib/bitcoin_wallet.dart
```diff
@@ -430,6 +440,44 @@ abstract class BitcoinWalletBase extends ElectrumWallet with Store {
...
+      }
+      final result = (ur.result as UR);
+      final cbor = result.cbor;
+      final cborDecoder = CBORDecoder(cbor);
+      final out = cborDecoder.decodeBytes();
+      final bytes = out.$1;
+      final base64psbt = base64Encode(bytes);
+      final psbt = PsbtV2()..deserializeV0(base64Decode(base64psbt));
+
+      // psbt.finalize();
```
OmarHatem28, line 465: is this no longer needed?

cw_bitcoin/lib/bitcoin_wallet.dart
```diff
@@ -430,6 +440,44 @@ abstract class BitcoinWalletBase extends ElectrumWallet with Store {
...
+
+      return PendingBitcoinTransaction(
+        btcTx,
+        type,
+        electrumClient: electrumClient,
+        amount: 0,
+        fee: 0,
+        feeRate: "",
+        network: network,
+        hasChange: true,
```
OmarHatem28, line 479: We can get these values, why are we hardcoding them?

lib/src/screens/ur/animated_ur_page.dart
```diff
@@ -160,6 +179,24 @@ class _URQRState extends State<URQR> {
...
+  }
+
+  late String nextLabel = widget.urqr.keys.toList()[(selectedInt + 1) % widget.urqr.length] ;
+
+  void next() {
+    final keys = widget.urqr.keys.toList();
+    selectedInt++;
+    setState(() {
+      nextLabel = keys[(selectedInt) % keys.length];
+      selected = keys[(selectedInt) % keys.length];
```
OmarHatem28: I think this is not what you intended right?

cw_bitcoin/lib/bitcoin_wallet.dart
konstantinullrich: Previosly that was a Bug. 
The masterfingerprint should be 0x00000000, but since we did Address matching before we just never noticed that the psbt was slightly wrong by having a null masterfingerprint

cw_bitcoin/lib/electrum_wallet.dart
MrCyjaneK: I've opted to use abstraction made for monero in cw_core (commitUR and similar functions) in bitcoin, since hardware wallet flow is significantly different in XMR. Using hardware wallet check would require us to treat cupcake as different kind of hardware wallet, duplicating the screens that we use, this together with walletinfo hive box sometimes getting corrupted made me go with a different approach.

lib/view_model/wallet_keys_view_model.dart
MrCyjaneK: Due to issues with derivation paths in cake, I had to adjust BDK function calls to derive xpub using the same derivation path that we use.

cw_bitcoin/lib/bitcoin_wallet.dart
MrCyjaneK: most likely it's not - since everything is working as expected.
cc: @konstantinullrich

cw_bitcoin/lib/bitcoin_wallet.dart
MrCyjaneK: that's just `commitPsbt` with the addition of UR parsing, I didn't change how the transaction object is being constructed.

lib/src/screens/ur/animated_ur_page.dart
MrCyjaneK: `selectedInt++;` should be in setState - other than that it is good

lib/view_model/wallet_keys_view_model.dart
OmarHatem28: I would understand that for other derivation paths, but since we are using the correct derivation path for native segwit (84) I believe it would generate a zpub, no?

MrCyjaneK: BDK prefers to generate xpub as far as I can tell (https://github.com/bitcoindevkit/bdk/discussions/977).

> BDK approach follows rust-bitcoin and doesn't use z/y/v pubs.

We could manually generate zpub but it is no what is recommended as far as I can tell (all sources are in discord and I can't find the messages linked so uh..) - actually bdk doesn't support zpub at all

lib/view_model/wallet_address_list/wallet_address_list_view_model.dart
```diff
@@ -603,6 +603,11 @@ abstract class WalletAddressListViewModelBase extends WalletChangeListenerViewMo
   bool get isSilentPayments =>
       wallet.type == WalletType.bitcoin && bitcoin!.hasSelectedSilentPayments(wallet);
 
+  @computed
+  bool get isCupcake =>
```
OmarHatem28, line 607: **Enhancement**
I would prefer renaming it to `isViewOnly` since this check on the `privateKey` is also valid for HWW

OmarHatem28 (approved):

PR #2313, opened 2025-06-09
Improving spanish translation

ktecho:


MrCyjaneK (approved):
took a quick glance with help of google translate, seems good.

PR #2314, opened 2025-06-10
CW-1090-ledger-issues

konstantinullrich:
# Description

This PR should improve the error handling of Ledger exceptions 
# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/lib/reactions/on_authentication_state_change.dart
+++ b/lib/reactions/on_authentication_state_change.dart
@@ -9,6 +9,7 @@ import 'package:cake_wallet/routes.dart';
 import 'package:cake_wallet/src/screens/connect_device/connect_device_page.dart';
 import 'package:cake_wallet/src/screens/wallet_connect/services/bottom_sheet_service.dart';
 import 'package:cake_wallet/src/widgets/alert_with_one_action.dart';
+import 'package:cake_wallet/src/widgets/alert_with_two_actions.dart';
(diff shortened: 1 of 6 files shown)


lib/reactions/on_authentication_state_change.dart
```diff
@@ -61,16 +62,49 @@ void startAuthenticationStateChange(
...
+                },
               );
-              await loadCurrentWallet();
+              bool tryOpening = true;
+              while (tryOpening) {
+                try {
+                  await loadCurrentWallet();
+                  tryOpening = false;
+                } on Exception catch (e) {
+                  if (e.toString().contains('0x5515')) {
```
OmarHatem28: if it's another exception, it would continue trying forever, shouldn't we add a limitation and then rethrow the error

lib/utils/exception_handler.dart
```diff
@@ -183,6 +186,56 @@ class ExceptionHandler {
     _hasError = false;
   }
 
+  static const List<String> _ledgerErrors = [
+    'Wrong Device Status',
+    'PlatformException(133, Failed to write: (Unknown Error: 133), null, null)',
```
OmarHatem28, line 191: isn't this too generic? or is it something that only comes from Ledger?

konstantinullrich: As far as I know this only comes from universal_ble

OmarHatem28 (approved):

PR #2317, opened 2025-06-11
CW-1091-payjoin-error-handeling

konstantinullrich:
# Description

Initialize Payjoin after successful sync and improve error handling  

# Pull Request - Checklist  

- [x] Format code


cw_bitcoin/lib/payjoin/manager.dart
```diff
@@ -176,15 +189,7 @@ class PayjoinManager {
     }
   }
 
-  Future<void> spawnNewReceiver({
-    required Receiver receiver,
-    bool isTestnet = false,
-  }) async {
-    await _payjoinStorage.insertReceiverSession(receiver, _wallet.id);
```
OmarHatem28: you no longer need to store the receiver sessions when spawned?
then who are you iterating over `_payjoinSessionSources`?
or because you don't need to check on the receiving sessions anyway and only need to check on the sending sessions

konstantinullrich: No longer needed. All instances of receiver are created using initReceiver which saves the session anyways.

PR #2323, opened 2025-06-13
feat: add optional parameter to customize address extraction pattern

konstantinullrich:
# Description

This should fix the address extraction on restoring from QRCode

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/lib/core/address_validator.dart
+++ b/lib/core/address_validator.dart
@@ -284,7 +284,7 @@ class AddressValidator extends TextValidator {
   static String get silentPaymentAddressPattern => SilentPaymentAddress.regex.pattern;
   static String get mWebAddressPattern => MwebAddress.regex.pattern;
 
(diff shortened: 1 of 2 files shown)


lib/core/address_validator.dart
```diff
@@ -336,7 +336,9 @@ class AddressValidator extends TextValidator {
     }
 
     if (pattern != null) {
-      return "$BEFORE_REGEX($pattern)$AFTER_REGEX";
+      if (requireSurroundingWhitespaces)
+        return "$BEFORE_REGEX($pattern)$AFTER_REGEX";
```
OmarHatem28: I believe this regex should already handle whether it is preceded by space or not, so this should work as well, if not then the regex is wrong, can you check that instead

konstantinullrich: If that check would be optional, what would be the point of having it in the regex at all?

OmarHatem28 (approved):

PR #2329, opened 2025-06-19
CW-1086: Polygon Issues

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_polygon/lib/polygon_wallet.dart
+++ b/cw_polygon/lib/polygon_wallet.dart
@@ -45,7 +45,12 @@ class PolygonWallet extends EVMChainWallet {
     final initialErc20Tokens = DefaultPolygonErc20Tokens().initialPolygonErc20Tokens;
 
     for (final token in initialErc20Tokens) {
-      if (!evmChainErc20TokensBox.containsKey(token.contractAddress)) {
+      if (evmChainErc20TokensBox.containsKey(token.contractAddress)) {
+        final existingToken = evmChainErc20TokensBox.get(token.contractAddress);
+        if (existingToken?.tag != token.tag) {
+          evmChainErc20TokensBox.put(token.contractAddress, token);
+        }
+      } else {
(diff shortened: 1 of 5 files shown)


cw_polygon/lib/polygon_wallet.dart
```diff
@@ -45,7 +45,12 @@ class PolygonWallet extends EVMChainWallet {
     final initialErc20Tokens = DefaultPolygonErc20Tokens().initialPolygonErc20Tokens;
 
     for (final token in initialErc20Tokens) {
-      if (!evmChainErc20TokensBox.containsKey(token.contractAddress)) {
+      if (evmChainErc20TokensBox.containsKey(token.contractAddress)) {
+        final existingToken = evmChainErc20TokensBox.get(token.contractAddress);
+        if (existingToken?.tag != token.tag) {
+          evmChainErc20TokensBox.put(token.contractAddress, token);
+        }
```
OmarHatem28, line 52: why was this case introduced?

OliveKiddies: It fixes the issue with the Polygon wallets not coming up in address books

The tag previously associated with the default tokens is "POLY" while the one associated with Cryptocurrencies and other related items is "POL", so when we choose a token, it doesn't correspond to the tag for any currency or wallet type.

This case updates the saved ERC20 Tokens in the wallet with the previous tag to the updated tag "POL".

Blazebrain: It fixes the issue with the Polygon wallets not coming up in address books

The tag previously associated with the default tokens is "POLY" while the one associated with Cryptocurrencies and other related items is "POL", so when we choose a token, it doesn't correspond to the tag for any currency or wallet type.

This case updates the saved ERC20 Tokens in the wallet with the previous tag to the updated tag "POL".

PR #2361, opened 2025-07-04
CW-974: Add support for automatic node switching.

Blazebrain:

# Description

This change:
- Adds isTrustedNodeForSwitching property to Node class.
- Adds UI Checkbox to mark a node as trusted for automatic switching in Node create or edit form.
- Implements checkNodeHealth method in wallet classes to verify node connectivity.
- Adds NodeSwitchingService for managing automatic node switching based on health checks.
- Adds setting option for user to enable/disable automated node switching

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_evm/lib/evm_chain_wallet.dart
+++ b/cw_evm/lib/evm_chain_wallet.dart
@@ -115,6 +115,8 @@ abstract class EVMChainWalletBase
   int? gasBaseFee = 0;
   int estimatedGasUnits = 0;
 
+  Timer? _updateFeesTimer;
+
   bool _isTransactionUpdating;
 
   // TODO: remove after integrating our own node and having eth_newPendingTransactionFilter
@@ -377,6 +379,7 @@ abstract class EVMChainWalletBase
   Future<void> close({bool shouldCleanup = false}) async {
     _client.stop();
     _transactionsUpdateTimer?.cancel();
+    _updateFeesTimer?.cancel();
   }
 
   @action
@@ -418,6 +421,9 @@ abstract class EVMChainWalletBase
       await _updateTransactions();
       await _updateEstimatedGasFeeParams();
 
+      _updateFeesTimer ??= Timer.periodic(const Duration(seconds: 30), (timer) async {
+        await _updateEstimatedGasFeeParams();
+      });
       syncStatus = SyncedSyncStatus();
     } catch (e) {
       syncStatus = FailedSyncStatus();
@@ -855,10 +861,9 @@ abstract class EVMChainWalletBase
       _transactionsUpdateTimer!.cancel();
     }
 
-    _transactionsUpdateTimer = Timer.periodic(const Duration(seconds: 30), (_) {
+    _transactionsUpdateTimer = Timer.periodic(const Duration(seconds: 20), (_) {
       _updateTransactions();
       _updateBalance();
-      _updateEstimatedGasFeeParams();
     });
   }
 
--- a/lib/entities/default_settings_migration.dart
+++ b/lib/entities/default_settings_migration.dart
@@ -512,8 +512,7 @@ Future<void> defaultSettingsMigration(
           );
           break;
         case 50:
-          await migrateExistingNodesToUseAutoSwitching(nodes: nodes, powNodes: powNodes);
-          await removeDuplicateElectrumCakewalletNodes(nodes: nodes);
+          migrateExistingNodesToUseAutoSwitching(nodes: nodes, powNodes: powNodes);
           break;
         default:
           break;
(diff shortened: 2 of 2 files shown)


assets/bitcoin_electrum_server_list.yml
OmarHatem28, line 1: we need to allow it for electrum.cakewallet.com as well

lib/src/screens/nodes/widgets/node_form.dart
```diff
@@ -153,6 +154,25 @@ class NodeForm extends StatelessWidget {
...
+            child: Row(
+              mainAxisAlignment: MainAxisAlignment.start,
+              mainAxisSize: MainAxisSize.max,
+              children: [
+                Observer(
+                  builder: (_) => StandardCheckbox(
+                    value: nodeViewModel.isTrustedNodeForSwitching,
+                    borderColor: Theme.of(context).colorScheme.surfaceContainerHighest,
+                    iconColor: Theme.of(context).colorScheme.primary,
+                    onChanged: (value) => nodeViewModel.isTrustedNodeForSwitching = value,
```
OmarHatem28: better use the setter as well and make the variable private

lib/src/screens/root/root.dart
```diff
@@ -161,6 +163,10 @@ class RootState extends State<Root> with WidgetsBindingObserver {
         }
 
         widget.tradeMonitor.resumeTradeMonitoring();
+
+        // Trigger node health check when app resumes
+        getIt.get<NodeSwitchingService>().performHealthCheck();
```
OmarHatem28: **Minor:**
better pass it as a dependency

cw_evm/lib/evm_chain_wallet.dart
```diff
@@ -828,6 +858,7 @@ abstract class EVMChainWalletBase
     _transactionsUpdateTimer = Timer.periodic(const Duration(seconds: 15), (_) {
       _updateTransactions();
       _updateBalance();
+      _updateEstimatedGasFeeParams();
```
OmarHatem28: previously it was called every 30 seconds, which I believe is sufficient for this function, now that I see our daily usage, I believe we should check which functions needs to be called frequently and which should be called less frequently so we can decide better and put periodic time accordingly

lib/view_model/node_list/node_list_view_model.dart
```diff
@@ -41,6 +41,12 @@ abstract class NodeListViewModelBase with Store {
       S.current.change_current_node(uri) +
       '${uri.endsWith('.onion') || uri.contains('.onion:') ? '\n' + (CakeTor.instance.enabled ? '' : S.current.orbot_running_alert) : ''}';
 
+  @computed
+  bool get enableAutomaticNodeSwitching => settingsStore.enableAutomaticNodeSwitching;
+
+  void setEnableAutomaticNodeSwitching(bool value) =>
```
OmarHatem28, line 48: shouldn't this be annotated with @action?

lib/view_model/node_list/pow_node_list_view_model.dart
```diff
@@ -41,6 +41,12 @@ abstract class PowNodeListViewModelBase with Store {
       S.current.change_current_node(uri) +
       '${uri.endsWith('.onion') || uri.contains('.onion:') ? '\n' + (CakeTor.instance.enabled ? '' : S.current.orbot_running_alert) : ''}';
 
+  @computed
+  bool get enableAutomaticNodeSwitching => settingsStore.enableAutomaticNodeSwitching;
+
+  void setEnableAutomaticNodeSwitching(bool value) =>
```
OmarHatem28, line 48: same

lib/entities/default_settings_migration.dart
OmarHatem28, line 1: `electrum.cakewallet.com` node should also be trusted, it's not in the yml file, not sure why, but it gets added in here
add it to the .yml file and make sure it's not added twice for new and existing users and that it is trusted for both as well

lib/entities/default_settings_migration.dart
```diff
@@ -510,7 +510,10 @@ Future<void> defaultSettingsMigration(
             providerName: "SwapTrade",
             enabled: true,
           );
-			    break;
+          break;
+        case 50:
+          await migrateExistingNodesToUseAutoSwitching(nodes: nodes);
```
OmarHatem28: this can be async, no need to be awaited
```suggestion
          migrateExistingNodesToUseAutoSwitching(nodes: nodes);
```

OmarHatem28 (changes requested):

OmarHatem28 (approved):

PR #2366, opened 2025-07-09
fix: handle case when feeFormated doesn't have currency at the end

MrCyjaneK:
# Description

fix: handle case when feeFormated doesn't have currency at the end

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (changes requested):
I would prefer to do it right instead of adding if condition.
add a `fee` attribute to the main `PendingTransaction` mixin and override it in each class that uses this mixin, with the actual fee itself not `feeFormatted`

lib/view_model/send/send_view_model.dart
```diff
@@ -183,7 +183,7 @@ abstract class SendViewModelBase extends WalletChangeListenerViewModel with Stor
         final currency = pendingTransactionFeeCurrency(walletType);
         final fiat = calculateFiatAmount(
             price: _fiatConversationStore.prices[currency]!,
-            cryptoAmount: pendingTransaction!.feeFormatted.substring(0, pendingTransaction!.feeFormatted.indexOf(" ")),
+            cryptoAmount: pendingTransaction!.feeFormattedValue,
```
OmarHatem28, line 186: should be used in deuro view model as well

OmarHatem28 (approved):

PR #2373, opened 2025-07-10
Cw 1118 silent payment fixes

rafael-xmr:


lib/core/address_validator.dart
```diff
@@ -281,7 +282,8 @@ class AddressValidator extends TextValidator {
     }
   }
 
-  static String get silentPaymentAddressPattern => SilentPaymentAddress.regex.pattern;
+  static String get silentPaymentAddressPatternMainnet => RegExp(r'^sp1[0-9a-zA-Z]{113}').pattern;
+  static String get silentPaymentAddressPatternTestnet => RegExp(r'^tsp1[0-9a-zA-Z]{113}').pattern;
```
OmarHatem28:
```suggestion
  static String get silentPaymentAddressPatternTestnet => RegExp(r'^(tsp|sprt)1[0-9a-zA-Z]{113}').pattern;
```

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -402,11 +402,15 @@ abstract class ElectrumWalletBase
             } else {
               // else: First time seeing this TX after scanning
               tx.unspents!.forEach(_updateSilentAddressRecord);
+              await save();
```
OmarHatem28: too many save
```suggestion
```

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -402,11 +402,15 @@ abstract class ElectrumWalletBase
...
 
               // Add new TX record
               transactionHistory.addMany(message);
+              await save();
```
OmarHatem28: same
```suggestion
```

lib/core/address_validator.dart
OmarHatem28, line 1: the decision of the regex should be based on the network, cuz if the user created a btc wallet with testnet, he should be able to send to test addresses.

Plus, this should be handled for other address types, Segwit, Taproot, SP, etc...

OmarHatem28 (changes requested):

PR #2385, opened 2025-07-13
CW-1129 fix: frozen coins displaying incorrect amounts in monero

MrCyjaneK:
# Description

fix: frozen coins displaying incorrect amounts in monero

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_monero/lib/monero_wallet.dart
+++ b/cw_monero/lib/monero_wallet.dart
@@ -841,10 +841,12 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
   }
 
   void _askForUpdateBalance() {
-    final unlockedBalance = _getUnlockedBalance();
+    final _ub = _getUnlockedBalance();
+    final _fb =_getFrozenBalance();
+    final unlockedBalance = _ub - _fb;
     final fullBalance = monero_wallet.getFullBalance(
-      accountIndex: walletAddresses.account!.id);
-    final frozenBalance = _getFrozenBalance();
+      accountIndex: walletAddresses.account!.id) - _fb;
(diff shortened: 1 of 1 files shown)


cw_monero/lib/monero_wallet.dart
```diff
@@ -841,10 +841,12 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
...
 
   void _askForUpdateBalance() {
-    final unlockedBalance = _getUnlockedBalance();
+    final _ub = _getUnlockedBalance();
+    final _fb =_getFrozenBalance();
+    final unlockedBalance = _ub - _fb;
     final fullBalance = monero_wallet.getFullBalance(
-      accountIndex: walletAddresses.account!.id);
-    final frozenBalance = _getFrozenBalance();
+      accountIndex: walletAddresses.account!.id) - _fb;
```
OmarHatem28, line 848: shouldn't the fullBalance, stay as fullBalance? i.e including the frozen coins? or are we using them differently in our UI layer?

MrCyjaneK: I'm not sure, it is counter intuitive but it's just monero handling things on it's side and we doing out part in the view model afaik. This change resulted in correct behavior in testing

OmarHatem28: ok I see now, it's because we don't care about the full balance, we pass to the supper class the full balance as the `additional` balance.

ok, I believe that choice was made when we wanted to shrink down the size of the balance card a bit, so all good then

OmarHatem28 (approved):

PR #2390, opened 2025-07-16
CW 1109 fix: incorrect restore height sometimes being set on monero

MrCyjaneK:

# Description

fix: incorrect restore height sometimes being set on monero

# Pull Request - Checklist  

- [ ] Format code


lib/exchange/provider/chainflip_exchange_provider.dart
OmarHatem28, line 1: I believe this is not the intended change, right? as it was modified recently by @CumpsD

cw_monero/lib/monero_wallet.dart
```diff
@@ -804,28 +804,39 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
...
     monero_wallet.setRecoveringFromSeed(isRecovery: true);
-    monero_wallet.setRefreshFromBlockHeight(height: height);
     setupBackgroundSync(password, currentWallet!);
   }
 
   int _getHeightDistance(DateTime date) {
     final distance =
-        DateTime.now().millisecondsSinceEpoch - date.millisecondsSinceEpoch;
+        DateTime.now().difference(date).inSeconds;
     final daysTmp = (distance / 86400).round();
```
OmarHatem28, line 836: Minor:
```
final daysTmp = DateTime.now().difference(date).inDays;
```

cw_monero/lib/monero_wallet.dart
```diff
@@ -804,28 +804,39 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
...
+        throw Exception("height is <= 0");
+      }
+      monero_wallet.setRefreshFromBlockHeight(height: height);
+    } catch (_) {
+      if (tryNum <= 3) {
+        printV("Failed to set height from date, retrying... $tryNum");
+        unawaited(() async {
+          await Future.delayed(Duration(seconds: 10));
+          _setHeightFromDate(tryNum: tryNum + 1);
+        }());
```
OmarHatem28, line 825: should we return after this `unawaited`?
else it will excute 
```
monero_wallet.setRecoveringFromSeed(isRecovery: true);
setupBackgroundSync(password, currentWallet!);
```
for the other trials as well,
or are they safe functions to get called frequently?

even better i would prefer separating the function that can fail and retried, from the other function calls

lib/exchange/provider/chainflip_exchange_provider.dart
CumpsD: the `as double`s should probably be ok.

the `List<Map<String, dynamic>>` I tried as well, but it crashed out when I tested so I had to go `List<dynamic>`

But if you test `List<Map<String, dynamic>>` and it works, that would be more correct, but please do test it :)

MrCyjaneK: not needed but couldn't get the code to compile with that, I can revert

cw_monero/lib/monero_wallet.dart
MrCyjaneK: we can continue, this function is safe to call multiple times, we can separate that if you prefer, but setupBackgroundSync also needs to be called - so only setRecoveringFromSeed would get removed

lib/exchange/provider/chainflip_exchange_provider.dart
CumpsD: Let me ping @konstantinullrich to see what he has to say :)

OmarHatem28 (approved):

PR #2405, opened 2025-07-23
CW-1078: Electrum Wallets not reconencting correctly if the OS closes the connection

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/core/node_switching_service.dart
+++ b/lib/core/node_switching_service.dart
@@ -15,12 +15,14 @@ class NodeSwitchingService {
 
   static const int _healthCheckIntervalSeconds = 30;
 
-  // The number of times we want to reset the overall used trusted nodes list. 
+  // The number of times we want to reset the overall used trusted nodes list.
   // We don't want an infinite loop if all trusted nodes are down.
   static const int _usedTrustedNodeResetCount = 2;
 
   // State to manage the reset count
   int _resetCount = 0;
+
+  String walletName = '';
 
   Timer? _healthCheckTimer;
 
@@ -70,6 +72,10 @@ class NodeSwitchingService {
   /// Switch to the next available trusted node
   Future<void> _switchToNextTrustedNode() async {
     _isSwitching = true;
+
+    if (walletName.isNotEmpty && (walletName != appStore.wallet!.name)) _resetCount = 0;
+
+    walletName = appStore.wallet!.name;
 
     try {
       final walletType = appStore.wallet!.type;


cw_bitcoin/lib/electrum.dart
```diff
@@ -446,7 +453,17 @@ class ElectrumClient {
 
   Future<dynamic> call(
       {required String method, List<Object> params = const [], Function(int)? idCallback}) async {
-    if (socket == null) {
+    if (!_isConnected || socket == null) {
+      SocketHealthLogger().logHealthCheck(
+        walletType: WalletType.bitcoin,//This is a placeholder
```
MrCyjaneK: maybe accept a null value instead to not confuse future debugging?

cw_core/lib/utils/socket_health_logger.dart
```diff
@@ -0,0 +1,154 @@
...
+class SocketHealthLogger {
+  static final SocketHealthLogger _instance = SocketHealthLogger._internal();
+  factory SocketHealthLogger() => _instance;
+  SocketHealthLogger._internal();
+
+  static const String _logFileName = 'socket_health_logs.json';
+  static const int _maxLogs = 1000;
+
+  Future<File> get _logFile async {
+    final directory = await getApplicationDocumentsDirectory();
```
MrCyjaneK: use getAppDir instead

cw_core/lib/utils/socket_health_logger.dart
MrCyjaneK, line 1: This should only be stored to disk on debug (if we really need to persist network health logs across app restarts - which I believe we shouldn't)

I've did a quick benchmark of this class (isolated from cake), and `Add Logs to File: 10000 ops in 31294m (3.129ms/op, 320 ops/sec)` takes on average 3.1ms - which is extremely close to 8.3ms that is the moment when we start dropping frames, so we shouldn't write them to files ideally at all as each time we use this class to addLog we are very likely to drop one frame, and calling it multiple times will affect user experience by a lot.

Maybe let's make this class a singleton and store everything in a `static List<>`?

cw_core/lib/utils/socket_health_logger.dart
```diff
@@ -0,0 +1,154 @@
...
+      walletType: WalletType.values.firstWhere((e) => e.name == json['walletType'] as String),
+      walletName: json['walletName'] as String,
+      isHealthy: json['isHealthy'] as bool,
+      error: json['error'] as String?,
+      syncStatus: json['syncStatus'] as String?,
+      wasReconnected: json['wasReconnected'] as bool,
+      trigger: json['trigger'] as String,
+    );
+  }
+}
```
MrCyjaneK: this is an unversioned struct (which is a non-issue if it is RAM-only class), if we decide to store it to disk we probably should add a version and be much more forgiving in term of types, for example:

`as String` -> `?.toString()`
`as String?` -> `?.toString() ?? ""`
`as bool` -> `?.toString == "true"`
`WalletType.values.firstWhere((e) => e.name == json['walletType'] as String)` -> I don't have an oneliner but switch statement would be much more efficient here - or even better - just store index of the enum

cw_core/lib/utils/socket_health_logger.dart
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  Future<void> addLog(SocketHealthLogEntry entry) async {
+    try {
+      printV('Adding log entry: ${entry.toLogString()}');
+      final logs = await getLogs();
+      logs.add(entry);
+
+      if (logs.length > _maxLogs) {
```
MrCyjaneK: `if (logs.length > _maxLogs*2) {`

removeRange is rather expensive operation, we shouldn't do it too often.

cw_core/lib/utils/socket_health_logger.dart
```diff
@@ -0,0 +1,154 @@
...
+  factory SocketHealthLogger() => _instance;
+  SocketHealthLogger._internal();
+
+  static const String _logFileName = 'socket_health_logs.json';
+  static const int _maxLogs = 1000;
+
+  Future<File> get _logFile async {
+    final directory = await getApplicationDocumentsDirectory();
+    return File('${directory.path}/$_logFileName');
+  }
```
MrCyjaneK:
```suggestion
  File? _file;
  Future<File> get _logFile async {
    if (_file != null) return file;
    final directory = await getApplicationDocumentsDirectory();
    return File('${directory.path}/$_logFileName');
  }
```

MrCyjaneK (changes requested):

cw_bitcoin/lib/electrum.dart
Blazebrain: Yeah, adjusted it to reflect that

cw_core/lib/utils/socket_health_logger.dart
Blazebrain: Migrated to a static list for the logs

cw_core/lib/utils/socket_health_logger.dart
Blazebrain: Migrated the logs to a static list

cw_core/lib/utils/socket_health_logger.dart
Blazebrain: Yep, makes sense

Removed this though, as we no more need it

cw_core/lib/utils/socket_health_logger.dart
Blazebrain: Alright, modified that

cw_core/lib/utils/socket_health_logger.dart
Blazebrain: Migrated to a static list for the logs

MrCyjaneK (approved):

PR #2416, opened 2025-07-29
CW-1139-Migrate-CakePay-mobile-to-use-the-new-backend-APIs

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/cake_pay/src/models/cake_pay_card.dart
+++ b/lib/cake_pay/src/models/cake_pay_card.dart
@@ -10,12 +10,11 @@ class CakePayCard {
   final String? cardImageUrl;
   final String? country;
   final FiatCurrency fiatCurrency;
-  final List<String> denominationsUsd;
-  final List<String> denominations;
   final String? minValueUsd;
   final String? maxValueUsd;
   final String? minValue;
   final String? maxValue;
+  final List<Denomination> denominationItems;
 
   CakePayCard({
     required this.id,
@@ -27,27 +26,41 @@ class CakePayCard {
     this.cardImageUrl,
     this.country,
     required this.fiatCurrency,
-    required this.denominationsUsd,
-    required this.denominations,
     this.minValueUsd,
     this.maxValueUsd,
     this.minValue,
     this.maxValue,
-  });
+    List<Denomination>? denominationItems,
+  }) : denominationItems = denominationItems ?? const [];
 
   factory CakePayCard.fromJson(Map<String, dynamic> json) {
-
     final name = stripHtmlIfNeeded(json['name'] as String? ?? '');
     final description = stripHtmlIfNeeded(json['description'] as String? ?? '');
     final termsAndConditions = stripHtmlIfNeeded(json['terms_and_conditions'] as String? ?? '');
     final howToUse = stripHtmlIfNeeded(json['how_to_use'] as String? ?? '');
-
     final fiatCurrency = FiatCurrency.deserialize(raw: json['currency_code'] as String? ?? '');
 
-    final List<String> denominationsUsd =
-        (json['denominations_usd'] as List?)?.map((e) => e.toString()).toList() ?? [];
-    final List<String> denominations =
-        (json['denominations'] as List?)?.map((e) => e.toString()).toList() ?? [];
+    List<Denomination> parsedDenomination = [];
+    final rawDenomination = json['denominations'];
+
+    if (rawDenomination is List) {
+      if (rawDenomination.isNotEmpty && rawDenomination.first is Map) {
+        parsedDenomination = rawDenomination
+            .whereType<Map<dynamic, dynamic>>()
+            .map<Denomination>((e) => Denomination.fromJson(Map<String, dynamic>.from(e)))
+            .toList();
+      } else {
+        parsedDenomination = rawDenomination
+            .map((e) => Denomination(value: Denomination._toDouble(e) ?? 0))
+            .toList();
+      }
+    } else if (rawDenomination is Map) {
+      final desc = rawDenomination['description'];
+      if (desc is List) {
+        parsedDenomination =
+            desc.map((e) => Denomination(value: Denomination._toDouble(e) ?? 0)).toList();
+      }
+    }
 
     return CakePayCard(
       id: json['id'] as int? ?? 0,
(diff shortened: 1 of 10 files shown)


lib/cake_pay/src/models/cake_pay_vendor.dart
```diff
@@ -17,8 +19,9 @@ class CakePayVendor {
     this.card,
   });
 
-  factory CakePayVendor.fromJson(Map<String, dynamic> json, String country) {
+  factory CakePayVendor.fromJson(Map<String, dynamic> json, String countryCode) {
     final name = stripHtmlIfNeeded(json['name'] as String);
+    final country = Country.fromCountryCode(countryCode)?.fullName ?? countryCode;
```
OmarHatem28: the vendor already has a country, you can get it from the response instead of passing it.

and more importantly, the `unavailable` boolean was removed and replaced by `available` for better readability and easily understandable, so please change it and make sure it's handled correctly if it has any negation `!` somewhere
(note: if `available` is null, it should default to false)

lib/view_model/cake_pay/cake_pay_cards_list_view_model.dart
```diff
@@ -220,15 +207,16 @@ abstract class CakePayCardsListViewModelBase with Store {
...
           search: searchString,
           giftCards: displayGiftCards,
-          prepaidCards: displayPrepaidCards,
-          custom: displayCustomValueCards,
-          onDemand: displayDenominationsCards);
+          prepaidCards: displayPrepaidCards);
 
-      cakePayVendors = CakePayVendorList = newVendors;
+      cakePayVendors = newVendors.where((vendor) {
+        if (vendor.card == null) return false;
```
OmarHatem28, line 217: also you can check `if (vendor.available != true) return false;` after the available variable change mentioned above

lib/cake_pay/src/models/cake_pay_card.dart
```diff
@@ -27,27 +26,41 @@ class CakePayCard {
...
+        parsedDenomination = rawDenomination
+            .map((e) => Denomination(value: Denomination._toDouble(e) ?? 0))
+            .toList();
+      }
+    } else if (rawDenomination is Map) {
+      final desc = rawDenomination['description'];
+      if (desc is List) {
+        parsedDenomination =
+            desc.map((e) => Denomination(value: Denomination._toDouble(e) ?? 0)).toList();
+      }
```
OmarHatem28: with the new backend the denominations is always a List of objects, why are we adding all these checks? let's simplify it please, since we made the new backend to simplify things not to complicate it

lib/cake_pay/src/models/cake_pay_card.dart
```diff
@@ -59,16 +72,75 @@ class CakePayCard {
...
+    if (v == null) return null;
+    if (v is num) return v.toDouble();
+    return double.tryParse(v.toString());
+  }
+
+  factory Denomination.fromJson(Map<String, dynamic> json) {
+    return Denomination(
+      value: _toDouble(json['value']) ??
+          _toDouble(json['amount']) ??
+          _toDouble(json['denomination']) ??
```
OmarHatem28: no need, they don't exist in the new backend
```suggestion
```

lib/cake_pay/src/models/cake_pay_card.dart
```diff
@@ -59,16 +72,75 @@ class CakePayCard {
...
+          0,
+      cardId: json['card_id'] is int ? json['card_id'] as int : int.tryParse('${json['card_id']}'),
+      usdValue:
+          _toDouble(json['usd_value']) ?? _toDouble(json['value_usd']) ?? _toDouble(json['usd']),
```
OmarHatem28:
```suggestion
          _toDouble(json['usd_value']),
```

lib/cake_pay/src/models/cake_pay_card.dart
```diff
@@ -59,16 +72,75 @@ class CakePayCard {
...
+  }
+
   static String stripHtmlIfNeeded(String text) {
     return text.replaceAll(RegExp(r'<[^>]*>|&[^;]+;'), ' ');
   }
 }
+
+class Denomination {
+  final double value;
+  final int? cardId;
```
OmarHatem28, line 91: this should be required, and this is the one that you should send when creating an order if it's there, if not (in the case of a price range card i.e card that doesn't have denominations) then you send the card id in the card object

OmarHatem28 (changes requested):
I don't see changes to the order creation API and its response, do we not use the response of the order creation? or is everything compatible with the new backend?

lib/cake_pay/src/models/cake_pay_card.dart
serhii-bor: removed unnecessary checks

lib/cake_pay/src/models/cake_pay_card.dart
serhii-bor: done

lib/cake_pay/src/models/cake_pay_vendor.dart
serhii-bor: done

lib/view_model/cake_pay/cake_pay_cards_list_view_model.dart
serhii-bor: done

lib/view_model/cake_pay/cake_pay_buy_card_view_model.dart
```diff
@@ -44,10 +49,13 @@ abstract class CakePayBuyCardViewModelBase with Store {
...
   bool confirmsVoidedRefund = false;
   bool confirmsTermsAgreed = false;
+  (String, int?) selectedDenomination = ('', null);
 
   String simulatedResponse = '';
 
-  bool get isDenominationSelected => card.denominations.isNotEmpty;
+  bool get isDenominationSelected =>
+      card.denominationItems.isNotEmpty &&
+      card.denominationItems.any((item) => item.value == amount && item.cardId == card.id);
```
OmarHatem28: cardId can be different, because the denominations might come from different cards (ex. if we have 2 Amazon cards, 1 from reloadly, and 1 from phaze, we will show 1 Amazon vendor and the denominations would have reloadly's card id and phaze's card id)
```suggestion
      card.denominationItems.any((item) => item.value == amount);
```

lib/cake_pay/src/models/cake_pay_card.dart
```diff
@@ -63,22 +61,79 @@ class CakePayCard {
...
-      denominations: denominations,
       minValueUsd: json['min_value_usd'] as String?,
       maxValueUsd: json['max_value_usd'] as String?,
       minValue: minValue,
       maxValue: json['max_value'] as String?,
+      denominationItems: denominations,
     );
   }
 
+  List<Denomination> get denominationItemsWithUniqueValue {
```
OmarHatem28: I believe we don't need this function as well, the backend already send us the denominations as unique and sorted by value.

please double check that if you want, through multiple requests, and if so, let's remove this so we don't add un-necessary logic to the client side while its handled by the backend

serhii-bor: I definitely can see cards with duplicated denomination values in the response.

OmarHatem28 (approved):

PR #2418, opened 2025-07-29
sp improvements

rafael-xmr:


lib/view_model/wallet_keys_view_model.dart
```diff
@@ -161,20 +162,20 @@ abstract class WalletKeysViewModelBase with Store {
         ]);
         break;
       case WalletType.bitcoin:
+        keys = bitcoin!.getSilentPaymentKeys(_appStore.wallet!);
```
OmarHatem28: this should be added after the normal keys below, preferably with different names as we enabled normal keys to show for bitcoin as well

lib/src/screens/wallet_keys/wallet_keys_page.dart
```diff
@@ -145,7 +161,12 @@ class _WalletKeysPageBodyState extends State<WalletKeysPageBody>
                 if (showKeyTab)
                   Padding(
                     padding: const EdgeInsets.only(left: 22, right: 22),
-                    child: _buildKeysTab(context),
+                    child: _buildKeysTab(context, widget.walletKeysViewModel.items.sublist(0, 4)),
```
OmarHatem28: `sublist` will throw an exception if the items list is less than 4
```suggestion
                    child: _buildKeysTab(context, showSilentPaymentsTab ? widget.walletKeysViewModel.items.sublist(0, 4) : widget.walletKeysViewModel.items),
```

PR #2429, opened 2025-08-02
Fix Spanish translations of search field text

jorgesumle:
I also did a recursive search with `grep -R search_category` (string from line 729) and it can't be found, so I guess it could be removed.


OmarHatem28 (approved):
it is being used in a few places actually.
Thanks for the contribution

PR #2437, opened 2025-08-06
sp improv

rafael-xmr:


lib/bitcoin/cw_bitcoin.dart
```diff
@@ -802,18 +802,22 @@ class CWBitcoin extends Bitcoin {
...
-    final bitcoinSPAddrs = wallet.walletAddresses as BitcoinWalletAddresses;
-    final bitcoinSPAddr = bitcoinSPAddrs.silentAddresses.first;
-    return bitcoinSPAddr.address;
+    final addresses = <String>[];
+    final labels = <String>[];
+    bitcoinTx.unspents!.forEach((unspent) {
+      addresses.add(bitcoinWallet.walletAddresses.silentAddresses
+          .firstWhere((address) => address.silentPaymentTweak == unspent.silentPaymentLabel)
+          .address);
+    });
```
OmarHatem28: wrap this with a try and catch, as this will be probably called with also normal transactions, so I believe `firstWhere` would throw an exception at that time

also rename the function to `getSPTransactionAddresses` since it only relates to SP

OmarHatem28 (changes requested):

lib/bitcoin/cw_bitcoin.dart
OmarHatem28:
```suggestion
    try {
          bitcoinTx.unspents!.forEach((unspent) {
            addresses.add(bitcoinWallet.walletAddresses.silentAddresses
                .firstWhere((address) => address.silentPaymentTweak == unspent.silentPaymentLabel)
                .address);
          });
     } catch (e) {}
```

PR #2450, opened 2025-08-12
feat: Pay Anything from Any Wallet

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/view_model/send/send_view_model.dart
```diff
@@ -73,6 +75,8 @@ abstract class SendViewModelBase extends WalletChangeListenerViewModel with Stor
   }
 
   UnspentCoinsListViewModel unspentCoinsListViewModel;
+  PaymentViewModel paymentViewModel;
+  WalletSwitcherViewModel walletSwitcherViewModel;
```
OmarHatem28: why are they here? I don't see them used

lib/core/universal_address_detector.dart
```diff
@@ -0,0 +1,391 @@
...
+        input.length > 10 &&
+        !input.contains('_');
+  }
+
+  /// Detect address from raw address patterns
+  static AddressDetectionResult _detectFromRawAddress(String input) {
+    final cleanInput = input.trim();
+
+    // Define detection patterns for each currency (ordered by specificity - most specific first)
+    final detectionPatterns = [
```
OmarHatem28, line 100: why not use the patterns function in `address_validator.dart`

lib/core/universal_address_detector.dart
```diff
@@ -0,0 +1,280 @@
...
+
+  /// Detects address from URI format (e.g., bitcoin:address?amount=0.001)
+  static AddressDetectionResult _detectFromUri(String input) {
+    try {
+      final uri = Uri.parse(input);
+
+      final paymentRequest = PaymentRequest.fromUri(uri);
+
+      // Determine currency from scheme
+      final currency = _schemeToCurrency[uri.scheme.toLowerCase()];
```
OmarHatem28: why limit it to only the `_schemeToCurrency` map? why not use `CryptoCurrency.fromString`?

Blazebrain: True, that's more comprehensive, I'll move to it and update the scheme with the relevant ones

lib/view_model/send/send_view_model.dart
Blazebrain: Was initially using SendViewModel for the flow before abstracting to a PaymentViewModel, didn't bring it out after that. I'll modify to inject it directly through SendPage constructor

lib/core/universal_address_detector.dart
Blazebrain: The pattern methods in `address_validator.dart` are too broad for some of the checks. 

They are also built on the fact that the cryptocurrency is known ahead, and we can then target that.
 
That's not the case for the detection flow, we have no other information asides the address itself, so we have to range our pattern checks from the most specific to the broader patterns, so we have an higher level of accuracy. 

Some of the test cases for the various address types fail when I route them through the AddressValidator, and we have some misfires also.

PR #2453, opened 2025-08-14
feat: use better embedded Tor

MrCyjaneK:
# Description

This was tested by me on:
- android
- macos

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_core/lib/utils/tor/torch.dart
+++ b/cw_core/lib/utils/tor/torch.dart
@@ -1,18 +1,32 @@
+import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
+import 'dart:isolate';
 
 import 'package:cw_core/utils/print_verbose.dart';
 import 'package:cw_core/utils/tor/abstract.dart';
 import 'package:path_provider/path_provider.dart';
 import 'package:path/path.dart' as p;
-import 'package:tor_binary/tor_binary_platform_interface.dart';
+import 'package:torch_dart/abstract_tor.dart';
 
-class CakeTorAndroid implements CakeTorInstance {
-  @override
-  bool get bootstrapped => _proc != null;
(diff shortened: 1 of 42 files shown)


cw_monero/lib/monero_wallet.dart
```diff
@@ -229,12 +229,12 @@ abstract class MoneroWalletBase extends WalletBase<MoneroBalance,
   @override
   Future<void> connectToNode({required Node node}) async {
     String socksProxy = node.socksProxyAddress ?? '';
-    printV("bootstrapped: ${CakeTor.instance.bootstrapped}");
-    printV("     enabled: ${CakeTor.instance.enabled}");
-    printV("        port: ${CakeTor.instance.port}");
-    printV("     started: ${CakeTor.instance.started}");
-    if (CakeTor.instance.enabled) {
-      socksProxy = "127.0.0.1:${CakeTor.instance.port}";
+    printV("bootstrapped: ${CakeTor.instance!.bootstrapped}");
```
OmarHatem28, line 232: I would prefer to make it `?` instead of `!` just in case, since initializing tor can fail for whatever reason.
and also do it for other places please

cw_core/lib/utils/tor/torch.dart
```diff
@@ -23,63 +37,54 @@ class CakeTorAndroid implements CakeTorInstance {
...
+  bool started = false;
+
+  static bool isTorRunning = false;
 
   @override
   Future<void> stop() async {
-    _proc?.kill();
-    await _proc?.exitCode;
-    _proc = null;
+    started = false;
```
OmarHatem28, line 46: should we set this also to false?
```suggestion
    started = false;
    isTorRunning = false;
```

PR #2455, opened 2025-08-14
Cw 1161 add user added tokens to swap currencies

Serhii-Borodenko:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -720,12 +738,26 @@ abstract class ExchangeViewModelBase extends WalletChangeListenerViewModel with
...
+        final hasStatic = provider.pairList.any((pair) => pair.from == from && pair.to == to);
+        if (hasStatic) return true;
+
+        // Allow all currencies for LetsExchange that are ETH or SOL based
+        if (provider is LetsExchangeExchangeProvider) {
+          return _isEthOrSolNetwork(from) || _isEthOrSolNetwork(to);
+        }
+
+        return false;
+      }).toList();
```
OmarHatem28: what if we remove this function? as for other tokens/exchanges we will have to manually update their supported currencies, but if we remove this, it will just call their API, and if they don't support that currency then it will just fail, let's give this a try

lib/solana/cw_solana.dart
```diff
@@ -97,7 +97,7 @@ class CWSolana extends Solana {
       mintAddress: contractAddress,
       decimal: token.decimals,
       mint: token.name.toUpperCase(),
-      enabled: token.enabled,
+      enabled: true,
```
OmarHatem28: nope, this function is also used to change the availability of the token, so like this the user will never be able to disable a token

lib/tron/cw_tron.dart
```diff
@@ -82,7 +82,7 @@ class CWTron extends Tron {
       symbol: token.title,
       contractAddress: contractAddress,
       decimal: token.decimals,
-      enabled: token.enabled,
+      enabled: true,
```
OmarHatem28: same

OmarHatem28:
and also in this last commit, please move the SPL tokens and TronTokens to cw_core @Serhii-Borodenko

lib/exchange/provider/trocador_exchange_provider.dart
```diff
@@ -274,10 +276,13 @@ class TrocadorExchangeProvider extends ExchangeProvider {
...
 
+      final from = responseJSON['ticker_from'] as String;
+      final to = responseJSON['ticker_to'] as String;
+
       return Trade(
         id: id,
-        from: CryptoCurrency.fromString(responseJSON['ticker_from'] as String),
-        to: CryptoCurrency.fromString(responseJSON['ticker_to'] as String),
+        from: CryptoCurrency.fromString(from),
+        to: CryptoCurrency.fromString(to),
```
OmarHatem28: this should take into consideration if the from/to are not in our CryptoCurrency list, thus this will throw an exception

lib/exchange/provider/xoswap_exchange_provider.dart
```diff
@@ -76,22 +76,22 @@ class XOSwapExchangeProvider extends ExchangeProvider {
       if (response.statusCode != 200) {
         throw Exception('Failed to fetch assets for ${currency.title} on ${currency.tag}');
       }
-      final assets = json.decode(response.body) as List<dynamic>;
+      final assets = (json.decode(response.body) as List).cast<Map<String, dynamic>>();
```
OmarHatem28: how are we casting a list to a map?

lib/exchange/provider/trocador_exchange_provider.dart
serhii-bor: fixed

lib/exchange/provider/xoswap_exchange_provider.dart
serhii-bor: refactored

OmarHatem28 (approved):

PR #2466, opened 2025-08-19
fix: fix wallet group errors for BIP39 wallets

### LINKED ISSUE #2413: Wallet groups falsely reporting corrupted seed during new wallet wizard
**Describe the bug**
It's impossible to use wallet group feature to open multiple wallets using shared BIP39, probably Linux specific issue?

**To Reproduce**
1. Restore (or create, doesn't matter) XMR wallet named "Test", choose BIP39.
2. Complete setup/restoration process.
3. Navigating to security and backup, show seed/keys outputs valid seeds.
4. Attempting to Wallets>Create New Wallet>BTC>Choose Wallet Group>Test results in prompt.
```
Corrupted seeds
The files for this wallet are corrupted and are unable to be opened. Please view the seed phrase, save it, and restore the wallet.
If the value is empty, then the seed was unable to be correctly recovered.
```
5. Pressing show seed (even though we can literally confirm it working through "Security and Backup") results in another prompt.
```
Corrupted seeds
Null check operator used on a null value
Corrupted wallets seeds (if retrievable, empty otherwise):
Failed to fetch Test seeds: Null
check operator used on a null value
Failed to fetch Test seeds: Null
check operator used on a null value
```
6. Ignoring this and progressing as app wants, which means creating new wallet instead (say BTC), newly created wallet (group) with different seed is created and that group has exactly the same corruption problem.
7. Purging both `$CAKE_WALLET_DIR` and `~/.local/share/com.example.cake_wallet` and redoing previous steps results in the same issue.

**Expected behavior**
BIP39 created wallet can be reused to open a wallet for each blockchain.

**Platform:**
 - OS: Artix Linux
 - Cake Wallet Version: 5.1.2-1 (AUR package, uses official binaries)

**Additional context**
Run command used (didn't matter for this, required by still broken #2079)
```shell
CAKE_WALLET_DIR="$XDG_CONFIG_HOME/cake_wallet" LD_PRELOAD=/opt/cake-wallet-bin/lib/monero_libwallet2_api_c.so LD_LIBRARY_PATH=/opt/cake-wallet-bin/lib cake_wallet --db-sync-mode safe:sync:250000000bytes
```

ronantakizawa:
Fixes #2413

# Description

Fixed a critical null pointer exception that caused false "Corrupted seeds" dialogs when using BIP39 Monero
  wallets in wallet groups.

**Problem:** When users created a BIP39 Monero wallet and then tried to create additional wallets in the same
wallet group, the app would crash with "Null check operator used on a null value" and display a misleading
"Corrupted seeds" error dialog, even though the wallet was functioning normally.

**Root Cause:** The getSeed() and getSeedLegacy() functions in cw_monero/lib/api/wallet.dart assumed the global
currentWallet variable was always non-null, but during wallet group operations, currentWallet could be
temporarily null while loading wallet metadata.

Solution: Added defensive null checks at the beginning of both functions to return empty strings gracefully
instead of crashing when currentWallet is null.

Impact:
- Resolves false "Corrupted seeds" dialogs for BIP39 wallet groups
- Maintains all existing functionality for normal wallet operations
- Affects all platforms (Linux, macOS, Windows, Android, iOS)
- Minimal change with zero risk of regression

# Pull Request - Checklist  

- [x] Format code


OmarHatem28:
Hi @ronantakizawa, thank you for opening this PR. However, this is not a fix for the issue, as we fetch the seed of the selected wallet group to create the wallet you selected with the same seed.

So if we applied your fix, it would return an empty string, which will not create the wallet you want to create.

Instead we want to fix the main reason for the error, because at this point in time `currentWallet` should not be null, so we want to fix that.
But neither me nor our testing team were able to reproduce this issue.

Can you share more info on how you reproduced it? And is it always reproducible or sometimes only?

nova2moon:
did git clone, then applied this fix (cherry-pick), yet still when trying to create a new wallet for any coin and selecting group wallet, getting these errors:

`flutter: _onRelayClientError ReownCoreError(code: 3000, message: Project ID is missing, data: null)
flutter: [INFO] cake_wallet/core/wallet_loading_service.dart#null:null WalletLoadingService: Null check operator used on a null value
flutter: [INFO] cake_wallet/core/wallet_loading_service.dart#null:null WalletLoadingService: Null check operator used on a null value
`
and claim that seeds are "corrupted", even though they verify/show just fine in security/backup.  Maybe similar fix is needed in other areas of the wallet/code based on above output?

nova2moon:
did git clone, then applied this fix, yet still

`flutter: _onRelayClientError ReownCoreError(code: 3000, message: Project ID is missing, data: null)
flutter: [INFO] cake_wallet/core/wallet_loading_service.dart#null:null WalletLoadingService: Null check operator used on a null value
flutter: [INFO] cake_wallet/core/wallet_loading_service.dart#null:null WalletLoadingService: Null check operator used on a null value
`

kathleenmotley49-maker (approved):

PR #2489, opened 2025-08-28
feat: Swap flow for Pay Anything.

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -360,6 +367,45 @@ abstract class ExchangeViewModelBase extends WalletChangeListenerViewModel with
...
+    var amount = '0.00';
+    try {
+      if (receiveAmount.isNotEmpty) {
+        if (fiatConversionStore.prices[receiveCurrency] == null) {
+          log('Price is null for ${receiveCurrency.title}, fetching price now');
+        }
+        amount = calculateFiatAmount(
+          price: fiatConversionStore.prices[receiveCurrency]!,
+          cryptoAmount: receiveAmount,
+        );
```
OmarHatem28: this should be in the if condition

lib/src/widgets/bottom_sheet/swap_confirmation_bottom_sheet.dart
```diff
@@ -0,0 +1,533 @@
...
+        },
+      ),
+    );
+  }
+
+  void _setUpReactions(
+    BuildContext context,
+    ExchangeViewModel exchangeViewModel,
+    PaymentFlowResult paymentFlowResult,
+  ) async {
```
OmarHatem28, line 299: you should add a boolean at the start of this function so it doesn't trigger it again if the `build` function gets triggered again, or even better, call this function in `initState` instead

lib/src/widgets/bottom_sheet/swap_details_bottom_sheet.dart
```diff
@@ -0,0 +1,461 @@
...
+
+        if (state is ExecutedSuccessfullyState) {
+          WidgetsBinding.instance.addPostFrameCallback((_) {
+            if (context.mounted) {
+              showModalBottomSheet<void>(
+                context: context,
+                isDismissible: false,
+                isScrollControlled: true,
+                builder: (BuildContext bottomSheetContext) {
+                  return ConfirmSendingBottomSheet(
```
OmarHatem28: like this, the user will slide to confirm swap, then slide to confirm sending, which is not a good user experience, I believe the first slide should do the sending as well, or make the confirm swap a button not a slider.
please check with @tuxpizza on that

lib/view_model/exchange/exchange_view_model.dart
Blazebrain: No it should not actually 

That's because the rate for the receive currency does not currently exist in the fiatConversionStore at that point, so we cannot calculate the fiat value for receiveAmount

I should however do an early return here, so it returns an empty string instead

lib/src/widgets/bottom_sheet/swap_confirmation_bottom_sheet.dart
Blazebrain: Alright

lib/src/widgets/bottom_sheet/swap_details_bottom_sheet.dart
Blazebrain: Yes, according to the current design, it would be two slides for the flow, we could probably switch that to a regular button, since the confirm sending is always going to have a slide to confirm on its own end. I'll have the chat with tux on it

lib/view_model/exchange/exchange_view_model.dart
OmarHatem28: oh I thought it was `!= null` but yeah, should return

kathleenmotley49-maker (approved):

PR #2500, opened 2025-09-05
CW-1189: Improve Automatic Node Switching

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/core/node_switching_service.dart
+++ b/lib/core/node_switching_service.dart
@@ -15,12 +15,12 @@ class NodeSwitchingService {
 
   static const int _healthCheckIntervalSeconds = 30;
 
-  // The number of times we want to reset the overall used trusted nodes list.
-  // We don't want an infinite loop if all trusted nodes are down.
-  static const int _usedTrustedNodeResetCount = 1;
+  // Maximum number of node attempts before giving up
+  // This prevents endless switching for users with slow network connections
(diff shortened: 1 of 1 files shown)


lib/core/node_switching_service.dart
```diff
@@ -58,8 +70,27 @@ class NodeSwitchingService {
...
     try {
-      final isHealthy = await appStore.wallet!.checkNodeHealth();
+      // Add timeout to prevent hanging on slow network connections
+      final isHealthy = await appStore.wallet!.checkNodeHealth()
+          .timeout(
+            Duration(seconds: 15),
+            onTimeout: () {
+              printV('Health check timed out, considering node unhealthy');
+              return false;
+            },
```
OmarHatem28, line 92: you do realize that this will not apply an actual timeout, right?

OmarHatem28 (changes requested):
also you added more places that might have a hidden edge case that will require more investigation, like the issue Ramo had.
Please simplify the checks, fix the timeout comment, and test it

lib/core/node_switching_service.dart
Blazebrain: Yeah, we're not actually looking to do an actual timeout for the underlying operation, we just want to stop "waiting" for it if it extends past the accepted timeframe

PR #2523, opened 2025-09-15
fix: initial 16KB pages support

MrCyjaneK:
# Description

all .so dependencies other than reown walletkit (`libuniffi_yttrium.so`) which is a prebuild are 16KB now

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/.github/workflows/pr_test_build_android.yml
+++ b/.github/workflows/pr_test_build_android.yml
@@ -230,6 +230,13 @@ jobs:
             popd
           popd
 
+      - name: Fetch reown dependencies
+        run: |
+          set -x -e
+          wget https://static.mrcyjanek.net/lfs/cake-yttrium/x86_64.so -O android/app/src/main/jniLibs/x86_64/libuniffi_yttrium.so
+          wget https://static.mrcyjanek.net/lfs/cake-yttrium/arm64-v8a.so -O android/app/src/main/jniLibs/arm64-v8a/libuniffi_yttrium.so
+          wget https://static.mrcyjanek.net/lfs/cake-yttrium/armeabi-v7a.so -O android/app/src/main/jniLibs/armeabi-v7a/libuniffi_yttrium.so
+
       - name: Execute Build and Setup Commands
         run: |
           pushd scripts/android
--- a/pubspec_base.yaml
+++ b/pubspec_base.yaml
@@ -42,7 +42,7 @@ dependencies:
   flutter_spinkit: ^5.1.0
   uni_links:
     git:
-      url: https://github.com/MrCyjaneK/uni_links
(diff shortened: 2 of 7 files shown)


pubspec_base.yaml
```diff
@@ -176,6 +176,11 @@ dependency_overrides:
       url: https://github.com/cake-tech/web3dart.git
       ref: cake
   flutter_secure_storage_platform_interface: 1.0.2
+  flutter_secure_storage:
+    git:
+      url: https://github.com/cake-tech/flutter_secure_storage.git
+      path: flutter_secure_storage
+      ref: ca897a08677edb443b366352dd7412735e098e7b
```
OmarHatem28, line 183: This is added through configure.dart
I remember we were adding it there because it didn't support Linux, but now I do see it, but anyway, we are not using it for Linux, so I would say keep it that way until we get rid of it

```suggestion
```

.github/workflows/pr_test_build_android.yml
```diff
@@ -230,6 +230,13 @@ jobs:
             popd
           popd
 
+      - name: Fetch reown dependencies
+        run: |
+          set -x -e
+          wget https://static.mrcyjanek.net/lfs/cake-yttrium/x86_64.so -O android/app/src/main/jniLibs/x86_64/libuniffi_yttrium.so
+          wget https://static.mrcyjanek.net/lfs/cake-yttrium/arm64-v8a.so -O android/app/src/main/jniLibs/arm64-v8a/libuniffi_yttrium.so
+          wget https://static.mrcyjanek.net/lfs/cake-yttrium/armeabi-v7a.so -O android/app/src/main/jniLibs/armeabi-v7a/libuniffi_yttrium.so
```
OmarHatem28, line 238: can we please host those and others on a Cake Wallet server instead?

in a separate PR of course, but let's move all to a Cake Wallet server (send me a message to set it up) and let's also make packages forks through Cake Labs github org

android/app/src/main/java/com/cakewallet/haven/Application.java
OmarHatem28, line 1: this file can be removed, no longer needed

OmarHatem28 (approved):

.github/workflows/pr_test_build_android.yml
MrCyjaneK: dm sent. There's no fork involved.

pubspec_base.yaml
MrCyjaneK: we have to add it to dependency_overrides (and configure.dart is not handling that currently), would you prefer me to keep it here as it is or change the way configure.dart works?

PR #2529, opened 2025-09-22
Pay Anything Enhancements

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/core/universal_address_detector.dart
+++ b/lib/core/universal_address_detector.dart
@@ -106,13 +106,25 @@ class UniversalAddressDetector {
 
       // Bitcoin Bech32
       _DetectionPattern(
-        pattern: RegExp(r'^bc1[a-km-zA-HJ-NP-Z1-9]{25,39}$'),
+        pattern: RegExp(r'^bc1[qpzry9x8gf2tvdw0s3jn54khce6mua7l]{25,62}$'),
+        currency: CryptoCurrency.btc,
+      ),
+
+      // Bitcoin Silent Payment (mainnet)
+      _DetectionPattern(
+        pattern: RegExp(r'^sp1[qpzry9x8gf2tvdw0s3jn54khce6mua7l]{113}$'),
+        currency: CryptoCurrency.btc,
+      ),
+
+      // Bitcoin Silent Payment (testnet)
+      _DetectionPattern(
+        pattern: RegExp(r'^(tsp|sprt)1[qpzry9x8gf2tvdw0s3jn54khce6mua7l]{113}$'),
         currency: CryptoCurrency.btc,
       ),
 
       // Litecoin Bech32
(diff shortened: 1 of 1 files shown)


lib/core/universal_address_detector.dart
OmarHatem28, line 1: why not use the predefined regex we have [here](https://github.com/cake-tech/cake_wallet/blob/main/lib/core/address_validator.dart#L327)

and for ltc, also why not use the regex in this file

lib/core/payment_uris.dart
```diff
@@ -0,0 +1,334 @@
...
+    final (isContract, targetAddress) = _getTargetAddress(uri.path);
+    final chainId = _getChainID(uri.path);
+
+    final address = isContract ? uri.queryParameters["address"] ?? '' : targetAddress;
+    final amount = isContract ? uri.queryParameters["uint256"] : uri.queryParameters["value"];
+
+    var formatedAmount = "";
+
+    if (amount != null) {
+      formatedAmount = formatFixed(BigInt.parse(amount), 18);
```
OmarHatem28: I don't see the `e18` handled here

lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -538,19 +542,22 @@ abstract class ExchangeViewModelBase extends WalletChangeListenerViewModel with
...
     for (int i = 0; i < result.length; i++) {
       if (result[i] != 0) {
         /// add this provider as its valid for this trade
-        try {
-          _sortedAvailableProviders[result[i]] = _providers[i];
-        } catch (e) {
-          // will throw "Concurrent modification during iteration" error if modified at the same
-          // time [createTrade] is called, as this is not a normal map, but a sorted map
-        }
+        newSortedProviders[result[i]] = _providers[i];
```
OmarHatem28: I know this is a new instance, but just in case so that if 2 calls were made
```suggestion
        try {
           newSortedProviders[result[i]] = _providers[i];
        } catch (e) {
          // will throw "Concurrent modification during iteration" error if modified at the same
          // time [createTrade] is called, as this is not a normal map, but a sorted map
        }
```

lib/utils/token_utilities.dart
```diff
@@ -0,0 +1,226 @@
...
+      }
+    }
+    return unique;
+  }
+
+  /// Finds a token by address across wallets depending on [walletType]
+  /// - EVM chains: match by contractAddress
+  /// - Solana: match by mintAddress
+  /// - Tron: match by contractAddress
+  static Future<dynamic> findTokenByAddress({
```
OmarHatem28:
```suggestion
  static Future<CryptoCurrency?> findTokenByAddress({
```

PR #2565, opened 2025-10-06
LCW-1128-Add-support-for-Trezor

konstantinullrich:
# Description

Add trezor support for BTC, LTC, ETH, POL using Trezor Connect

# Pull Request - Checklist  

- [ ] Format code


cw_bitcoin/lib/hardware/litecoin_ledger_service.dart
```diff
@@ -42,4 +45,29 @@ class LitecoinLedgerService extends HardwareWalletService {
...
+  Future<String> signLitecoinTransaction({
+    required List<BitcoinBaseOutput> outputs,
+    required List<LedgerTransaction> inputs,
+    required Map<String, PublicKeyWithDerivationPath> publicKeys,
+  }) {
+    String? changePath;
+    for (final output in outputs) {
+      final maybeChangePath = publicKeys[(output as BitcoinOutput).address.pubKeyHash()];
+      if (maybeChangePath != null) changePath ??= maybeChangePath.derivationPath;
+    }
```
OmarHatem28, line 74: can you add documentation in the code for this please

cw_bitcoin/lib/hardware/trezor_service.dart
```diff
@@ -0,0 +1,187 @@
...
+    final psbt = PsbtV2()..deserialize(base64Decode(transaction));
+
+    final inputs = <TrezorTxInput>[];
+    final inputCount = psbt.getGlobalInputCount();
+    for (var i = 0; i < inputCount; i++) {
+      final inputTxRaw = psbt.getInputNonWitnessUtxo(i);
+      final inputTx = BtcTransaction.fromRaw(hex.encode(inputTxRaw!));
+      final inputOutputIndex = psbt.getInputOutputIndex(i);
+
+      final publicKeys = psbt.inputMaps[i].keys.where((e) => e.startsWith("06"));
```
OmarHatem28, line 56: what's `06`? and also this part can use a little bit of code documentation

lib/main.dart
```diff
@@ -375,6 +376,7 @@ class AppState extends State<App> with SingleTickerProviderStateMixin {
           linkViewModel: linkViewModel,
           tradeMonitor: tradeMonitor,
           nodeSwitchingService: nodeSwitchingService,
+          trezorConnect: getIt<TrezorConnect>(),
```
OmarHatem28, line 379: why keep the object live for every instance of the app, just to check if a deep link contains a static string `callbackBackUri`

just check that static string in `root.dart` and if it indeed is targeted for Trezor, then initialize this

cw_bitcoin/lib/hardware/trezor_service.dart
konstantinullrich: `06` is the psbt key for a public key. So here we are selecting all public keys of that input

lib/main.dart
konstantinullrich: I didn't want to have a `getIt` call inside `root.dart`

PR #2596, opened 2025-10-17
Improve Automatic Node Switching

Blazebrain:
- Added connectivity checks before node health verification to ensure network availability.
- Introduced maximum node switching attempts and cooldown periods to prevent excessive retries.

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/core/node_switching_service.dart
```diff
@@ -15,12 +16,16 @@ class NodeSwitchingService {
...
-  // The number of times we want to reset the overall used trusted nodes list.
-  // We don't want an infinite loop if all trusted nodes are down.
-  static const int _usedTrustedNodeResetCount = 1;
+  // Maximum number of node switching attempts per session
+  static const int _maxNodeSwitchingAttempts = 5;
 
-  // State to manage the reset count
-  int _resetCount = 0;
+  // Cooldown period between node switching attempts (in seconds)
+  static const int _nodeSwitchingCooldownSeconds = 60;
```
OmarHatem28: 15 seconds is reasonable enough, since we have the timeout of node connection to 10 seconds anyway
```suggestion
  static const int _nodeSwitchingCooldownSeconds = 15;
```

lib/core/node_switching_service.dart
```diff
@@ -58,11 +63,45 @@ class NodeSwitchingService {
...
       final isHealthy = await appStore.wallet!.checkNodeHealth();
 
       if (!isHealthy) {
+        printV('Node health check failed. Attempting to switch to next trusted node.');
         await _switchToNextTrustedNode();
+      } else {
+        // Reset switching attempts on successful health check
+        _switchingAttempts = 0;
+        _hasExhaustedAllNodes = false;
+        printV('Node health check passed. Current node is healthy.');
```
OmarHatem28, line 104: being healthy doesn't mean it connected successfully, I would move that in a `finally` clause after the switching happens maybe

lib/core/node_switching_service.dart
```diff
@@ -108,21 +152,31 @@ class NodeSwitchingService {
...
+        } else {
+          printV('Maximum attempts reached. No more node switching.');
+          _hasExhaustedAllNodes = true;
+          return;
+        }
       }
 
+      // Increment switching attempts counter
+      _switchingAttempts++;
+      _lastSwitchingAttempt = DateTime.now();
```
OmarHatem28: should be in `finally`

lib/core/node_switching_service.dart
```diff
@@ -136,8 +190,24 @@ class NodeSwitchingService {
       printV('Successfully switched to node: ${nextNode.uriRaw}');
     } catch (e) {
       printV('Error switching to next trusted node: $e');
+
+      // Increment attempts even on error to prevent infinite retries
+      _switchingAttempts++;
+      _lastSwitchingAttempt = DateTime.now();
```
OmarHatem28: same

lib/core/node_switching_service.dart
Blazebrain: I reviewed the code again, and I don't think so, we should not move it from the current block.

Here's what I have in mind: The code block checks for the health for the currently connected node, it makes calls like fetching balance, a piece of history etc. A successful response from this `checkNodeHealth` call means that the current node is  connected and healthy, as it would have failed otherwise.

If this node health check connection is confirmed to be healthy, we can safely reset our switching state, because we're not switching neither have we exhausted the node tries.

I think it should remain here

lib/core/node_switching_service.dart
Blazebrain: Alright, adjusting to 15 seconds

lib/core/node_switching_service.dart
Blazebrain: Yeah, combined both into the finally block

lib/core/node_switching_service.dart
Blazebrain: Yeah, combined both into the finally block

lib/core/node_switching_service.dart
OmarHatem28: oh right, I forgot that the checkNodeHealth function does that, thought it just pings the node to see if it's online or not, ok all good then, thanks

OmarHatem28 (approved):

PR #2600, opened 2025-10-22
CW-1208: Pay Anything EVM enahancements

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/src/screens/send/widgets/send_card.dart
+++ b/lib/src/screens/send/widgets/send_card.dart
@@ -446,6 +446,8 @@ class SendCardState extends State<SendCard> with AutomaticKeepAliveClientMixin<S
                       final address =
                           output.isParsedAddress ? output.extractedAddress : output.address;
 
+                      if (address.contains('@')) return;
+
                       await _handlePaymentFlow(
                         address,
                         PaymentRequest(


lib/utils/token_utilities.dart
```diff
@@ -220,10 +221,77 @@ class TokenUtilities {
...
+    final baseCurrency = walletTypeToCryptoCurrency(network);
+    final allTokens = <CryptoCurrency>[];
+    final addedAddresses = <String>{};
+
+    allTokens.add(baseCurrency);
+
+    for (final currency in CryptoCurrency.all) {
+      final matches = baseCurrency == CryptoCurrency.eth
+          ? baseCurrency.title == currency.tag
+          : currency.tag == baseCurrency.tag;
```
OmarHatem28: what about sol and tron?
just make it like this
```suggestion
      final matches = baseCurrency.title == currency.tag || currency.tag == baseCurrency.tag;
```

lib/utils/token_utilities.dart
```diff
@@ -220,10 +221,77 @@ class TokenUtilities {
...
+  static Future<List<CryptoCurrency>> _getUserTokensForNetwork(
+    CryptoCurrency baseCurrency,
+    Box<WalletInfo> walletInfoSource,
+  ) async {
+    final tokens = await TokenUtilities.loadAllUniqueEvmTokens(walletInfoSource);
+
+    return tokens.where((token) {
+      if (baseCurrency == CryptoCurrency.eth) return token.tag == baseCurrency.title;
+
+      return token.tag?.toLowerCase() == baseCurrency.tag?.toLowerCase();
```
OmarHatem28: same, will miss solana and tron, and will probably match any currencies that has no tags together, although I think this function is only called with EVM, SOL, and TRX so it's safe, but double check it please, or safer, modify this

Blazebrain: The function is safe as its doesn't cover SOL and TRX.

The tokens loaded are EVM tokens, and among these, the exception we have is ETH as it doesn't have a tag(and I've handled it). All the other L2s have tags for their respective networks.

`final tokens = await TokenUtilities.loadAllUniqueEvmTokens(walletInfoSource);`

lib/utils/token_utilities.dart
Blazebrain: Same here too. 

This function covers the EVM currencies and it's tokens, and `Cryptocurrency.eth` is the only exception, which was handled.

I'll add a change to make it generic(though not needed).

lib/utils/token_utilities.dart
```diff
@@ -216,10 +217,76 @@ class TokenUtilities {
...
+      // Match by tag for POL/BASE, match by title==tag for ETH
+      final matches = (baseCurrency.tag == null && baseCurrency.title == currency.tag) ||
+          (baseCurrency.tag != null &&
+              currency.tag?.toLowerCase() == baseCurrency.tag?.toLowerCase());
+
+      if (matches && _shouldAddToken(allTokens, currency, addedAddresses)) {
+        allTokens.add(currency);
+        if (currency is Erc20Token) {
+          addedAddresses.add(currency.contractAddress.toLowerCase());
+        }
```
OmarHatem28: currency will never be `ERC20Token` you are looping on `CryptoCurrency.all` which doesn't have any ERC20Token, it would still work since addedAddresses will always be empty, but just stating that this is useless

lib/utils/token_utilities.dart
```diff
@@ -216,10 +217,76 @@ class TokenUtilities {
...
+    }
+
+    // Add user tokens that don't already exist
+    final userTokens = await _getUserTokensForNetwork(baseCurrency);
+    for (final token in userTokens) {
+      if (_shouldAddToken(allTokens, token, addedAddresses)) {
+        allTokens.add(token);
+        if (token is Erc20Token) {
+          addedAddresses.add(token.contractAddress.toLowerCase());
+        }
```
OmarHatem28, line 249: here it make sense because you are looping on user's token

lib/utils/token_utilities.dart
```diff
@@ -216,10 +217,76 @@ class TokenUtilities {
...
+      }
+      return true;
+    }
+
+    return !existingTokens.any((existing) => _matchesCurrency(existing, token));
+  }
+
+  static bool _matchesCurrency(CryptoCurrency a, CryptoCurrency b) {
+    return a.title.toUpperCase() == b.title.toUpperCase() &&
+        (a.tag?.toUpperCase() == b.tag?.toUpperCase() || (a.tag == null && b.tag == null));
```
OmarHatem28: already included in the first part
```suggestion
        a.tag?.toUpperCase() == b.tag?.toUpperCase();
```

lib/src/widgets/bottom_sheet/swap_confirmation_bottom_sheet.dart
```diff
@@ -137,7 +137,9 @@ class SwapConfirmationContentState extends State<SwapConfirmationContent> {
 
   @override
   Widget build(BuildContext context) {
-    final detectedCurrency = walletTypeToCryptoCurrency(widget.paymentFlowResult.walletType!);
+    final detectedCurrency = widget.paymentFlowResult.type == PaymentFlowType.evmNetworkSelection
+        ? widget.paymentFlowResult.addressDetectionResult!.detectedCurrency!
+        : walletTypeToCryptoCurrency(widget.paymentFlowResult.walletType!);
```
OmarHatem28: I would prefer that `addressDetectionResult!.detectedCurrency` would already return the currency, and handle this logic inside of it, instead of having this condition in a couple of places

lib/utils/token_utilities.dart
Blazebrain: Yeah that makes sense

We don't need the check there. I'll take it out

lib/utils/token_utilities.dart
Blazebrain: Yeah

lib/src/widgets/bottom_sheet/swap_confirmation_bottom_sheet.dart
Blazebrain: Alright, pushed an update for it

OmarHatem28 (approved):

PR #2620, opened 2025-10-30
CW-1238-Solana-Fixes

Blazebrain:
- Optimize SPL token transactions and balance fetching
- Auto Populate all fields when fetching SPL tokens
- Gracefully handle errors when sending Solana tokens

### DIFF
--- a/lib/view_model/send/send_view_model.dart
+++ b/lib/view_model/send/send_view_model.dart
@@ -1022,7 +1022,7 @@ abstract class SendViewModelBase extends WalletChangeListenerViewModel with Stor
       }
 
       if (error is CreateAssociatedTokenAccountException) {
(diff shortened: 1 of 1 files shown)


cw_solana/lib/solana_wallet.dart
```diff
@@ -315,27 +315,41 @@ abstract class SolanaWalletBase
...
+
+      for (final list in results) {
+        if (list.isNotEmpty) allResults.addAll(list);
       }
     }
 
-    // await _addTransactionsToTransactionHistory(splTokenTransactions);
+    if (allResults.isNotEmpty) {
+      await _addTransactionsToTransactionHistory(allResults);
+    }
```
OmarHatem28: I remember we moved this inside the loop before because the transactions took to long to show up to the user, so let's do it after every batch at least

Blazebrain: Alright, I've moved it to update per batch

PR #2630, opened 2025-11-03
CW1272 tails, linux generic fixes

MrCyjaneK:
# Description

# Pull Request - Checklist  

- [ ] Format code


cw_core/lib/db/sqlite.dart
```diff
@@ -1,9 +1,14 @@
 
-import 'package:sqflite/sqflite.dart';
+import 'dart:io';
+
+import 'package:sqflite_common_ffi/sqflite_ffi.dart';
 
 late Database db; 
 
 Future<void> initDb({String? pathOverride}) async {
+  if (Platform.isLinux) {
```
OmarHatem28: this should also include macos and windows.
also, we forgot to test backward compatability on desktop platforms, so please run a quick test on them, by having the production version and then building latest main on top of it

MrCyjaneK: @konstantinullrich did you get a chance to test windows? Mac is working just fine

konstantinullrich: Will Test today

konstantinullrich: @MrCyjaneK 
````Bad state: databaseFactory not initialized
databaseFactory is only initialized when using sqflite. When using `sqflite_common_ffi`
You must call `databaseFactory = databaseFactoryFfi;` before using global openDatabase API```

Comment (konstantinullrich) [reply cw_core/lib/db/sqlite.dart]:
```suggestion
  if (Platform.isLinux || Platform.isWindows) { 
```
That fixes the error

PR #2633, opened 2025-11-04
refactor-receive-vm

konstantinullrich:
# Description

Refactor Receive logic to make View Models less complex

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/cw_bitcoin/lib/bitcoin_wallet_addresses.dart
+++ b/cw_bitcoin/lib/bitcoin_wallet_addresses.dart
@@ -119,7 +119,6 @@ abstract class BitcoinWalletAddressesBase extends ElectrumWalletAddresses with S
       final invoice = await lightningWallet!.getBolt11Invoice(amountSats, "Send to Cake Wallet");
       return LightningPaymentRequest(address: address, amount: amount, bolt11Invoice: invoice);
     }
-    print(amount);
     return getPaymentUri(amount);
   }
 }


cw_bitcoin/lib/bitcoin_wallet_addresses.dart
```diff
@@ -89,4 +92,22 @@ abstract class BitcoinWalletAddressesBase extends ElectrumWalletAddresses with S
...
   }
+
+  @override
+  List<ReceivePageOption> get receivePageOptions {
+    if (isHardwareWallet) {
+      return [
+        ...BitcoinReceivePageOption.allViewOnly,
+        ...ReceivePageOptions.where((element) => element != ReceivePageOption.mainnet)
+      ];
+    }
```
OmarHatem28, line 105: what about cupcake and the other airgapped wallets

cw_core/lib/payment_uris.dart
OmarHatem28, line 1: The `XXXURI` classes are all the same, it definitely can be generalized and shortened by a lot, and since this is a refactoring PR, why not do it here as well

I don't think we check `if (x is MoneroURI)` at any place in the code, so why do we need separate classes.
the only ones that need custom `toString` like bitcoin, can have their separate child class of the main one, but the rest should just use the base class and call it a day

cw_core/lib/wallet_addresses.dart
```diff
@@ -79,4 +83,8 @@ abstract class WalletAddresses {
 
   bool containsAddress(String address) =>
       addressesMap.containsKey(address) || allAddressesMap.containsKey(address);
+
+  List<ReceivePageOption> get receivePageOptions => ReceivePageOptions;
+
+  PaymentURI getPaymentUri(String amount);
```
OmarHatem28: same here, this would be standard for most of the wallets, so preferably just implement it here and whatever coin that needs custom implementation, should override it

cw_evm/lib/evm_chain_wallet_addresses.dart
```diff
@@ -36,4 +39,16 @@ abstract class EVMChainWalletAddressesBase extends WalletAddresses with Store {
...
+
+  @override
+  PaymentURI getPaymentUri(String amount) {
+    switch (chainId) {
+      case 8453:
+        return BaseURI(amount: amount, address: address);
+      case 137:
+        return PolygonURI(amount: amount, address: address);
+      default:
+        return EthereumURI(amount: amount, address: address);
```
OmarHatem28: those can be a single child class like bitcoin and would take an extra chainId param, and here it would be one line of code instead of a switch

cw_solana/lib/solana_wallet_addresses.dart
OmarHatem28, line 1: this and the likes of it should be removed by the changes mentioned above

lib/bitcoin/cw_bitcoin.dart
```diff
@@ -304,30 +304,6 @@ class CWBitcoin extends Bitcoin {
     return bitcoinWallet.walletAddresses.addressPageType == SilentPaymentsAddresType.p2sp;
   }
 
-  @override
-  List<ReceivePageOption> getBitcoinReceivePageOptions(Object wallet) {
-    final bitcoinWallet = wallet as ElectrumWallet;
-    final keys = bitcoinWallet.keys;
-    if (keys.privateKey.isEmpty) {
-      return BitcoinReceivePageOption.allViewOnly;
```
OmarHatem28, line 312: this one is missing in the bitcoin wallet addresses

lib/core/address_validator.dart
```diff
@@ -19,6 +19,7 @@ class AddressValidator extends TextValidator {
                       r'^(lightning:)?(lnbc|lntb|lnbs|lnbcrt)[a-z0-9]+$',
                       caseSensitive: false);
                   if (lightningInvoiceRegex.hasMatch(txt)) return true;
+                  if (txt.contains("@")) return true;
```
OmarHatem28, line 22: technically we don't need that, since it will get resolved as an actual address like we have for any other address alias (birdpay, unstoppable, ens, etc...)

konstantinullrich: For alias pay yes, but for Lightning Adresses they don't get resolved

cw_bitcoin/lib/bitcoin_wallet_addresses.dart
konstantinullrich: Since my HWW refactor they are tagged as Hardware wallets

cw_core/lib/payment_uris.dart
```diff
@@ -0,0 +1,249 @@
...
+  String _formatAmountForERC20(String amount) {
+    try {
+      // Convert decimal amount to BigInt (assuming 18 decimals)
+      final amountDouble = double.parse(amount.replaceAll(",", "."));
+      final amountBigInt = BigInt.from(amountDouble * 1e18);
+      return amountBigInt.toString();
+    } catch (e) {
+      return amount.replaceAll(",", ".");
+    }
+  }
```
OmarHatem28, line 130: minor, but I guess there will be an edge case of precision with this approach

cw_core/lib/payment_uris.dart
```diff
@@ -0,0 +1,249 @@
...
+    final amountParam = isContract ? uri.queryParameters["uint256"] : uri.queryParameters["value"];
+
+    var formatedAmount = "";
+
+    if (amountParam != null) {
+      final normalized = _normalizeToIntegerWei(amountParam);
+      formatedAmount = formatFixed(BigInt.parse(normalized), 18);
+    } else {
+      formatedAmount = uri.queryParameters["amount"] ?? "";
+    }
```
OmarHatem28, line 159: there should be a lot of error handling here in cases of an unexpected uri

PR #2656, opened 2025-11-17
fix: block wrongly parsed addresses

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/lib/view_model/payment/payment_view_model.dart
+++ b/lib/view_model/payment/payment_view_model.dart
@@ -45,7 +45,7 @@ abstract class PaymentViewModelBase with Store {
         return PaymentFlowResult.incompatible('Unable to detect address type');
       }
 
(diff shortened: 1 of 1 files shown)


lib/src/screens/send/widgets/send_card.dart
```diff
@@ -852,6 +852,8 @@ class SendCardState extends State<SendCard> with AutomaticKeepAliveClientMixin<S
 
         final parsedAddress = output.isParsedAddress ? output.extractedAddress : output.address;
 
+        if (parsedAddress.contains('@')) return;
```
OmarHatem28: why not just use that `if` or any address validation mechanism we may have in the future, in `_handlePaymentFlow` instead

OmarHatem28 (approved):

PR #2677, opened 2025-11-25
CW-1250: Refactor L2 Chains Code

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/utils/qr_util.dart
```diff
@@ -2,6 +2,7 @@ import 'package:cw_core/wallet_type.dart';
 
 String getQrImage(WalletType type) {
   switch (type) {
+    case WalletType.evm:
```
OmarHatem28:
```suggestion
    // TODO: in the refactoring, it would be `wallet.qrImage` so in case of evm, let's display the QR based on the current selected chain id
    case WalletType.evm:
```

lib/entities/node_list.dart
```diff
@@ -19,6 +19,7 @@ Future<List<Node>> loadDefaultNodes(WalletType type) async {
     case WalletType.haven:
       path = 'assets/haven_node_list.yml';
       break;
+    case WalletType.evm:
```
OmarHatem28:
```suggestion
    // TODO: (refactoring) each wallet would have its path, so `wallet.nodePath` would be decided based on chain id in Evm wallet
    case WalletType.evm:
```

lib/view_model/wallet_keys_view_model.dart
```diff
@@ -14,25 +14,44 @@ import 'package:flutter/foundation.dart';
...
+        return null;
+      }, (_) {
+        final wallet = _appStore.wallet;
+        if (wallet != null && isEVMCompatibleChain(wallet.type)) {
+          final currentChain = evm!.getCurrentChain(wallet);
+          _title =
+              '${currentChain?.name ?? walletTypeToString(wallet.type)} ${S.current.wallet_keys}';
+        }
+      });
+    }
```
OmarHatem28: why do we need a reaction here? this view model is registered as a factory anyway, and there is no way a user can switch wallet while at this stage.
(except on desktop, which is already broken and doesn't update the seeds until you go out of this page and open it again, which we shouldn't care about, since we will make a new design for desktop anyway)

cw_evm/lib/evm_chain_wallet.dart
```diff
@@ -141,42 +179,207 @@ abstract class EVMChainWalletBase
...
+    if (selectedChainId == chainId) return;
+
+    final oldNativeCurrency = currency;
+
+    _client.stop();
+
+    final tokensToRemove = balance.keys.whereType<Erc20Token>().toList();
+    for (final token in tokensToRemove) {
+      balance.remove(token);
+    }
```
OmarHatem28: just clear the whole thing, even if the oldNativeCurrency is the same, it will be re-fetched anyway when you call `startSync`

cw_evm/lib/evm_chain_wallet.dart
```diff
@@ -141,42 +179,207 @@ abstract class EVMChainWalletBase
...
+        // Check if erc20TokensBox is already initialized (migration already done)
+        final _ = erc20TokensBox;
+        // If we get here, migration is done, use normal chain-specific logic
+        // Fall through to normal chain-specific box logic below
+      } catch (_) {
+        // erc20TokensBox not initialized yet, run migration
+        await _initEthereumErc20TokensBox();
+        return;
+      }
+    }
```
OmarHatem28, line 256: what's that?!

lib/view_model/contact_list/contact_list_view_model.dart
```diff
@@ -44,7 +47,11 @@ abstract class ContactListViewModelBase with Store {
             walletContacts.add(WalletContact(
               address.address,
               name,
-              walletTypeToCryptoCurrency(info.type),
+              walletTypeToCryptoCurrency(
+                info.type,
+                chainId: info.type == WalletType.evm ? evm!.getChainIdByWalletType(info.type) : null,
+              ),
+              walletType: info.type,
```
OmarHatem28: why are you defaulting to ethereum chain id?

lib/view_model/dashboard/balance_view_model.dart
```diff
@@ -66,6 +66,16 @@ abstract class BalanceViewModelBase with Store {
...
+
+    reaction((_) {
+      if (isEVMCompatibleChain(wallet.type)) {
+        final selectedChainId = evm!.getSelectedChainId(wallet);
+        return '${wallet.currency.title}_${selectedChainId}';
+      }
+      return null;
+    }, (_) {
+      wallet.balance;
+    });
```
OmarHatem28: what's the point of this?

lib/view_model/send/fees_view_model.dart
OmarHatem28, line 1: why is arbitrum missing from here?

lib/view_model/anon_invoice_page_view_model.dart
```diff
@@ -182,8 +182,8 @@ abstract class AnonInvoicePageViewModelBase with Store {
 
   @action
   void reset() {
-    selectedCurrency = walletTypeToCryptoCurrency(_wallet.type);
-    cryptoCurrency = walletTypeToCryptoCurrency(_wallet.type);
+    selectedCurrency = walletTypeToCryptoCurrency(_wallet.type, chainId: _wallet.chainId);
+    cryptoCurrency = walletTypeToCryptoCurrency(_wallet.type, chainId: _wallet.chainId);
```
OmarHatem28: sometimes you are using `wallet.currency` and someothers you are using `walletTypeToCryptoCurrency`, let's unify one across all places, so if we will make any modifications it will be in one place

cw_core/lib/erc20_token.dart
```diff
@@ -74,9 +74,7 @@ class Erc20Token extends CryptoCurrency with HiveObjectMixin {
   static const arbitrumBoxName = 'ArbitrumErc20Tokens';
 
   @override
-  bool operator ==(other) =>
-      (other is Erc20Token && other.contractAddress == contractAddress) ||
-      (other is CryptoCurrency && other.title == title);
+  bool operator ==(Object other) => other is Erc20Token && other.contractAddress == contractAddress;
```
OmarHatem28, line 77: why is this change?

lib/src/widgets/bottom_sheet/swap_confirmation_bottom_sheet.dart
```diff
@@ -166,16 +167,20 @@ class SwapConfirmationContentState extends State<SwapConfirmationContent> {
...
                     ),
-                    Positioned(
-                      bottom: -4,
-                      right: -4,
-                      child: CakeImageWidget(
-                        imageUrl: walletTypeToCryptoCurrency(widget.paymentFlowResult.walletType!)
-                            .iconPath!,
-                        width: 16,
-                        height: 16,
+                    if (isEVMCompatibleChain(widget.paymentFlowResult.walletType!)) ...[
```
OmarHatem28, line 170: why now only show this if it's evm?

cw_evm/lib/clients/evm_chain_client.dart
```diff
@@ -11,26 +11,92 @@ import 'package:cw_evm/evm_chain_transaction_priority.dart';
...
-  int get chainId;
+  //! Can be overridden by child classes
+
+  int get chainId => _chainId;
 
   Future<List<EVMChainTransactionModel>> fetchTransactions(String address,
-      {String? contractAddress});
+      {String? contractAddress}) async {
+    try {
+      final response = await client.get(Uri.https("api.etherscan.io", "/v2/api", {
```
OmarHatem28, line 42:
```suggestion
      /// when adding new chains, make sure they are supported by the same api through https://docs.etherscan.io/supported-chains
      final response = await client.get(Uri.https("api.etherscan.io", "/v2/api", {
```

lib/src/screens/connect_device/select_hardware_wallet_account_page.dart
```diff
@@ -149,8 +149,10 @@ class _SelectHardwareWalletAccountFormState extends State<SelectHardwareWalletAc
                         padding: EdgeInsets.only(top: 10),
                         child: SelectButton(
                           image: Image.asset(
-                            walletTypeToCryptoCurrency(_walletHardwareRestoreVM.type).iconPath ??
-                                '',
+                            walletTypeToCryptoCurrency(
+                              _walletHardwareRestoreVM.type,
+                              chainId: _walletHardwareRestoreVM.type == WalletType.evm ? 1 : null,
```
OmarHatem28: why default to eth?

lib/evm/cw_evm.dart
```diff
@@ -0,0 +1,492 @@
...
+  @override
+  String getAddress(WalletBase wallet) => (wallet as EVMChainWallet).walletAddresses.address;
+
+  @override
+  String getPrivateKey(WalletBase wallet) {
+    final privateKeyHolder = (wallet as EVMChainWallet).evmChainPrivateKey;
+    if (privateKeyHolder is EthPrivateKey) {
+      return bytesToHex(privateKeyHolder.privateKey);
+    }
+    return "";
```
OmarHatem28, line 79: is the private key only available on eth wallets?

lib/evm/cw_evm.dart
```diff
@@ -0,0 +1,492 @@
...
+    if (service is EVMChainLedgerService) {
+      ((wallet as EVMChainWallet).evmChainPrivateKey as EvmLedgerCredentials).setLedgerConnection(
+          service.ledgerConnection, (await wallet.walletInfo.getDerivationInfo()).derivationPath);
+    } else if (service is EVMChainBitboxService) {
+      ((wallet as EVMChainWallet).evmChainPrivateKey as EvmBitboxCredentials)
+          .setBitbox(service.manager, (await wallet.walletInfo.getDerivationInfo()).derivationPath);
+    } else if (service is EVMChainTrezorService) {
+      ((wallet as EVMChainWallet).evmChainPrivateKey as EvmTrezorCredentials).setTrezorConnect(
+          service.connect, (await wallet.walletInfo.getDerivationInfo()).derivationPath);
+    }
```
OmarHatem28: can be simplified and more readable

lib/src/screens/send/widgets/send_card.dart
```diff
@@ -245,10 +256,14 @@ class SendCardState extends State<SendCard> with AutomaticKeepAliveClientMixin<S
           paymentViewModel: paymentViewModel,
           paymentRequest: paymentRequest,
           onNext: (PaymentFlowResult newResult) {
-            if (newResult.addressDetectionResult!.detectedWalletType ==
-                paymentViewModel.currentWalletType) {
+            final selectedChainId = newResult.chainId;
+            final isCompatible = selectedChainId == evm!.getSelectedChainId(sendViewModel.wallet);
+
+            if (isCompatible) {
```
OmarHatem28, line 262: why you removed wallet type compatibility and only added chainId compatibility, shouldn't both be here?

lib/src/screens/send/widgets/send_card.dart
```diff
@@ -286,9 +301,27 @@ class SendCardState extends State<SendCard> with AutomaticKeepAliveClientMixin<S
...
-          .setSelectedCryptoCurrency(result.addressDetectionResult!.detectedCurrency!.title);
+
+      final detectedCurrency = result.addressDetectionResult!.detectedCurrency;
+      if (detectedCurrency != null) {
+        if (isEVMCompatibleChain(sendViewModel.wallet.type) &&
+            detectedCurrency == sendViewModel.wallet.currency) {
+          sendViewModel.selectedCryptoCurrency = sendViewModel.wallet.currency;
+        } else {
+          sendViewModel.setSelectedCryptoCurrency(detectedCurrency.title);
+        }
```
OmarHatem28: what's the difference? just remove the if/else and call `sendViewModel.setSelectedCryptoCurrency(detectedCurrency.title);`

lib/src/screens/wallet_connect/services/chain_service/eth/evm_chain_id.dart
```diff
@@ -9,29 +11,28 @@ enum EVMChainId {
...
+    return evm!.getCaip2ByChainId(chainId);
+  }
+
+  int? _getChainIdForEnum(EVMChainId id) {
+    return switch (id) {
+      EVMChainId.ethereum => 1,
+      EVMChainId.polygon => 137,
+      EVMChainId.base => 8453,
+      EVMChainId.arbitrum => 42161,
+      _ => 1,
```
OmarHatem28: would prefer if we handle all enum cases instead of having a default case, so that when we add any other chain, this will give a syntax error until we properly add it here

lib/src/screens/wallet_connect/services/walletkit_service.dart
```diff
@@ -562,9 +578,11 @@ abstract class WalletKitServiceBase with Store {
     final filteredPairings = allPairings.where(
       (pairing) {
         bool isInCurrentTopics = currentTopicsForWallet.contains(pairing.topic);
-        bool isActive = pairing.active;
+        // bool isActive = pairing.active;
+        // bool hasSession = sessions.any((session) => session.pairingTopic == pairing.topic);
 
-        return isInCurrentTopics && isActive;
+        // return isInCurrentTopics && isActive;
+        return isInCurrentTopics;
```
OmarHatem28, line 585: why are those no longer needed?

lib/src/widgets/bottom_sheet/payment_confirmation_bottom_sheet.dart
```diff
@@ -102,6 +112,13 @@ class _PaymentConfirmationContent extends StatelessWidget {
         final isMwebOrSpAddress =
             _isMwebOrSpAddress(paymentFlowResult.addressDetectionResult?.address ?? '');
 
+        final isEVMChainMismatch = paymentFlowResult.type == PaymentFlowType.evmNetworkSelection &&
+            paymentFlowResult.wallet != null &&
+            isEVMCompatibleChain(paymentViewModel.currentWalletType);
```
OmarHatem28: the naming is confusing

lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -185,6 +186,19 @@ abstract class ExchangeViewModelBase extends WalletChangeListenerViewModel with
...
+        return wallet.currency;
+      }
+      return null;
+    }, (_) async {
+      // When chain changes, update currencies and refresh token lists
+      await Future.delayed(const Duration(milliseconds: 100));
+      receiveCurrency = wallet.currency;
+      depositCurrency = wallet.currency;
+      _injectUserEthTokensIntoCurrencyLists();
+    });
```
OmarHatem28: can the user change the chain at this stage?
maybe for pay anything, but if so, wouldn't it make sense to make this reaction listen on wallet.currency always, not just when it's an evm compatible chain?

OmarHatem28 (changes requested):

lib/view_model/send/fees_view_model.dart
Blazebrain: Arbitrum does not support priority fees, it processes based on a first in first out (FIFO) basis

cw_core/lib/erc20_token.dart
Blazebrain: It affected the indexOf operator, which makes use of the "==" operator for comparison.

Based on the previous override, which adds `|| (other is CryptoCurrency && other.title == title);`, it means that any cryptocurrency that had the smae title compared the same, regardless of the chain/tag.

That's what caused one of the bugs Jack reported earlier, that USDC(Base) was returned, even though he selected USDC(Arb).

So indexOf stops at the first USDC(which was base), because it satisfies its check.

cw_evm/lib/clients/evm_chain_client.dart
Blazebrain: Alright, added the comment

cw_evm/lib/evm_chain_wallet.dart
Blazebrain: We could do that

Was looking at the UX, didn't want to instantly remove all, so the user still sees something.

But since we have a loader, doesn't matter still. 

I'll clear it.

lib/evm/cw_evm.dart
Blazebrain: No, EthPrivateKey doesn't refer to only eth wallets only.

It's the generic class used by web3dart to refer to the credentials holder(priv key)

lib/evm/cw_evm.dart
Blazebrain: Okay

lib/src/screens/connect_device/select_hardware_wallet_account_page.dart
Blazebrain: At the moment, the ETH icon is what we use as the default for EVM chain, since we don't still have a new icon.

Defaulting for eth here returns the eth icon image for EVM wallets

lib/src/screens/send/widgets/send_card.dart
Blazebrain: No it shouln't. All wallets have a chainId, both new and old wallets. Which means we can make the direct check on the chainId of the wallets

Also, if we still use the types, we'll have a conflict of the old types and the new type.

An old wallet, having the type WalletType.eth would clash with a wallet of type WalletType.evm(with a chainId of 1), even though both of them are for the ETh chain, the flow would repeat the whole process all over again.

lib/src/screens/send/widgets/send_card.dart
Blazebrain: Yeah, I think I initially added some other code into the if block earlier, and removed later on in the implementation.

But it can be called directly.

lib/src/screens/wallet_connect/services/chain_service/eth/evm_chain_id.dart
Blazebrain: Okay

lib/src/widgets/bottom_sheet/payment_confirmation_bottom_sheet.dart
Blazebrain: Adjusted it

lib/src/widgets/bottom_sheet/swap_confirmation_bottom_sheet.dart
Blazebrain: It's used to show the network for the detected chain. 

So instead of just show USDC image alone, we show USDC image with the base network image shown at the bottom right, so the user knows its USDC on Base network.

Works like a badge. 

Other non-evm networks don't need this.

lib/view_model/anon_invoice_page_view_model.dart
Blazebrain: Yeah, should be wallet.currency across board. I already added the same parameters in WalletBase

lib/view_model/dashboard/balance_view_model.dart
Blazebrain: Was using it to update the balance earlier. But its now redundant

lib/view_model/exchange/exchange_view_model.dart
Blazebrain: Yeah, updated to list to wallet currency

cw_evm/lib/evm_chain_wallet.dart
Blazebrain: Adjusted the comment to reflect the flow

It handles the backward-compatibility migration code for old WalletType.ethereum wallets. It was a case in the old EthWallet class, the `movePreviousErc20BoxConfigsToNewBox()` method.

The code checks if the old global box exists by accessing erc20TokensBox; if it throws (not initialized), it runs _initEthereumErc20TokensBox() to migrate tokens from the global box to the wallet-specific box, then returns.

If the box already exists (migration done), it falls through to the normal chain-specific box initialization logic below.

lib/view_model/wallet_keys_view_model.dart
Blazebrain: Yeah, the user can't switch chains here. The title updates here too, but I'll move it to another entry

lib/src/screens/wallet_connect/services/walletkit_service.dart
Blazebrain: Mainly because we no longer need them

For the isActive check, the pairing can be valid but inactive, and we want to be able to either disconnect properly or reconnect. The previous check filters out those types of pairing completely so we don't have the option to take action on them.

Same for the hasSession check, we have another way to retrieve sessions when needed, and a pairing should not depend on that.

lib/src/widgets/bottom_sheet/swap_confirmation_bottom_sheet.dart
OmarHatem28: for me it seems like this is used to display the currency image, so now you are only showing it in case of an EVM wallet.

the if condition was not there, so this is not a new added badge, this hides an existing image in case of it's not evm, and I don't think we had the badge previously

lib/exchange/provider/chainflip_exchange_provider.dart
```diff
@@ -29,7 +29,7 @@ class ChainflipExchangeProvider extends ExchangeProvider {
     CryptoCurrency.sol,
     CryptoCurrency.usdcsol,
     CryptoCurrency.arbEth,
-    // TODO: Add CryptoCurrency.usdcarb
+    CryptoCurrency.usdcArb,
```
OmarHatem28, line 32: @Serhii-Borodenko this is supported now right?

lib/src/screens/send/widgets/send_card.dart
OmarHatem28: ok, just noticed that this function itself is specific to EVM, I thought it's generic

lib/src/widgets/bottom_sheet/swap_confirmation_bottom_sheet.dart
Blazebrain: No it's not, I think you mixed up the widgets a bit.

The parent widget is a Stack, which houses two children. There's an image widget that always displays the main detected currency image, then there is another image widget, gated with the if check, that displays the network of the detected wallet type. 

The first image would always display, regardless of walletType, the second one would only display for EVM wallets.

cw_evm/lib/evm_chain_wallet_service.dart
```diff
@@ -1,56 +1,358 @@
...
+      throw Exception('Chain config not found for wallet type: ${walletInfo.type}');
+    }
+    final initialChainId = chainConfig.chainId;
+
+    final client = EVMChainClientFactory.createClient(initialChainId);
+
+    final wallet = _createWalletInstance(
+      walletType: walletInfo.type,
+      walletInfo: walletInfo,
+      derivationInfo: await walletInfo.getDerivationInfo(),
```
OmarHatem28: at this point the sqlite table doesn't have the derivation info right? how would this work?
shouldn't this be something we get from the chainConfig?
or default it to `m/44'/60'/` since this is the derivation path for evm wallets?

scripts/android/pubspec_gen.sh
OmarHatem28, line 1: missing for macos and linux

lib/view_model/settings/other_settings_view_model.dart
```diff
@@ -67,7 +69,7 @@ abstract class OtherSettingsViewModelBase with Store {
 
   @computed
   bool get displayTransactionPriority =>
-      !(changeRepresentativeEnabled || [WalletType.solana, WalletType.tron].contains(_wallet.type));
+      !(changeRepresentativeEnabled || [WalletType.solana, WalletType.tron].contains(_wallet.type) || chainId == 42161);
```
OmarHatem28:
```suggestion
      !(changeRepresentativeEnabled || [WalletType.solana, WalletType.tron, WalletType.arbitrum].contains(_wallet.type));
```

cw_evm/lib/evm_chain_wallet_service.dart
Blazebrain: Ohh, that's true. It will default to the unknown derivation type and no derivation type at that point.

I've added logic to set the default derivation path for EVM wallets if it's missing, so even if the table doesn't have it, we can still have the default to use for creating the wallet instance

OmarHatem28 (approved):
Missing Arbitrum scheme in AndroidManifestBase and info.plist and adding this scheme to restore from qr flow

Blazebrain:
> Missing Arbitrum scheme in AndroidManifestBase and info.plist and adding this scheme to restore from qr flow

They are all added, or are you referring to something else?

cw_evm/lib/utils/evm_chain_formatter.dart
```diff
@@ -0,0 +1,45 @@
...
+      final whole = parts[0].isEmpty ? '0' : parts[0];
+      final fraction = parts.length > 1 ? parts[1] : '';
+
+      final fractionPadded = fraction
+          .padRight(decimals, '0')
+          .substring(0, decimals > fraction.length ? decimals : fraction.length);
+
+      final wholeBigInt = BigInt.parse(whole);
+      final fractionBigInt = BigInt.parse(fractionPadded);
+      final multiplier = BigInt.from(10).pow(decimals);
```
OmarHatem28:
```suggestion
      String cleanAmount = amount.replaceAll(',', '.');

      bool isNegative = cleanAmount.startsWith('-');
      if (isNegative) {
        cleanAmount = cleanAmount.substring(1);
      }

      final parts = cleanAmount.split('.');
      String whole = parts[0].isEmpty ? '0' : parts[0];
      String fraction = parts.length > 1 ? parts[1] : '';

      // 3. Strict Truncation Logic
      // If fraction is longer than decimals, strictly cut it off.
      // If shorter, pad it with zeros.
      if (fraction.length > decimals) {
        fraction = fraction.substring(0, decimals);
      } else {
        fraction = fraction.padRight(decimals, '0');
      }

      final wholeBigInt = BigInt.parse(whole);
      final fractionBigInt = BigInt.parse(fraction);
      final multiplier = BigInt.from(10).pow(decimals);
```

OmarHatem28 (approved):

PR #2689, opened 2025-11-28
fix: fixes missing model generation that causes build failures

karlbuys:
# Description

The omission of arbitrum in ./model_generator.sh was causing errors when building the app from scratch

### DIFF
--- a/model_generator.sh
+++ b/model_generator.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -x -e
 
-for cwcoin in cw_{arbitrum,core,evm,monero,bitcoin,nano,bitcoin_cash,solana,tron,wownero,zano,decred,dogecoin,base}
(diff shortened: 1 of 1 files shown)


model_generator.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -x -e
 
-for cwcoin in cw_{core,evm,monero,bitcoin,nano,bitcoin_cash,solana,tron,wownero,zano,decred,dogecoin,base}
+for cwcoin in cw_{arbitrum,core,evm,monero,bitcoin,nano,bitcoin_cash,solana,tron,wownero,zano,decred,dogecoin,base}
```
MrCyjaneK: arbitrum should be added later on, as it depends on cw_core (and it's generated output)
```suggestion
for cwcoin in cw_{core,evm,monero,bitcoin,nano,bitcoin_cash,solana,tron,wownero,zano,decred,dogecoin,base,arbitrum}
```

PR #2711, opened 2025-12-08
Bloc-based Viewmodel for card customizer

malik1004x:

# Description

Card customizer Viewmodel based on Bloc.

# Pull Request - Checklist  

- [x] Format code


cw_core/lib/crypto_currency.dart
OmarHatem28, line 1: **minor**
would rather revert the formatting here since nothing changed and all would be a single line currency

cw_core/lib/wallet_info.dart
```diff
@@ -311,6 +326,82 @@ class DerivationInfo {
   }
 }
 
+class BalanceCardStyleSettings {
```
OmarHatem28: can you extract it please, I don't see it being here is optimal, if we want to scale it or so it will get bigger and bigger in a not necessarily related file

lib/new-ui/new_dashboard.dart
```diff
@@ -32,7 +34,11 @@ class _NewDashboardState extends State<NewDashboard> {
     return Scaffold(
       body: Stack(
         children: [
-         widget.dashboardPageWidgets[_selectedPage],
+          MultiBlocProvider(providers: [
+            BlocProvider<CardCustomizerBloc>(
+              create: (_) => CardCustomizerBloc(widget.dashboardViewModel.wallet),
```
OmarHatem28: preferably create it in di.dart

lib/new-ui/viewmodels/card_customizer/card_customizer_state.dart
```diff
@@ -0,0 +1,42 @@
...
+      return availableDesigns[selectedDesignIndex].withGradient(selectedColor);
+    } else {
+      return availableDesigns[selectedDesignIndex];
+    }
+  }
+
+  Gradient get selectedColor => CardDesign.allGradients[selectedColorIndex];
+}
+
+final class CardCustomizerInitial extends CardCustomizerState {
```
OmarHatem28, line 37: (Good practices) since you are re-using this state with different params, you can have a copyWith constructor and just pass the changed variable (ex. in `_onColorSelected` you can just do `emit(state.copyWith(colorIndex: event.newColorIndex));`

lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -330,6 +332,49 @@ abstract class DashboardViewModelBase with Store {
 
   bool _isTransactionDisposerCallbackRunning = false;
 
+
+  Future<void> loadCardDesigns() async {
```
OmarHatem28, line 337: would probably inject the bloc here and call `loadCardDesigns` from it directly, so that when we refactor this dashboardViewModel to bloc, we know that it will have a dependency on the other bloc

PR #2721, opened 2025-12-11
fix: backup-compatible db location on all platforms

malik1004x:

# Description

Moves the SQLite database to getAppDir() for all platforms. This ensures the DB gets included in backups and no data is lost.

# Pull Request - Checklist  

- [x] Format code


MrCyjaneK (changes requested):
instead of await copy use copySync, and equivalent of other IO operations. Sync functions are faster and we halt app startup until this is done.

OmarHatem28 (approved):

PR #2755, opened 2025-12-19
Node hive -> sqlite migration

malik1004x:

# Description

Migrate the Node class to be stored in the SQLite DB instead of Hive.

Also adds logic that ensures app-default nodes stored in the database always match the ones declared in .yml files. Default node data can now be changed simply by editing the files, no migration is required.

# Pull Request - Checklist  

- [x] Format code


assets/arbitrum_node_list.yml
```diff
@@ -1,12 +1,15 @@
 -
   uri: arbitrum.nownodes.io
-  is_default: true
+  isDefault: true
```
OmarHatem28: PublicNode should be the default node, please make sure it is and make the necessary migration

assets/base_node_list.yml
```diff
@@ -1,15 +1,19 @@
 -
   uri: base.nownodes.io
+  isDefault: true
```
OmarHatem28, line 3: same, but this one is already merged in latest dev, so just pull and update it please

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -791,9 +791,10 @@ abstract class ElectrumWalletBase
 
       if (server.toLowerCase().contains('electrs')) {
         node!.isElectrs = true;
-        if (node!.isInBox) {
+        // TODO figure out why condition was needed
+        // if (node!.isInBox) {
```
OmarHatem28, line 795: it was throwing random exceptions that this object is not in a box for some users, so we just wrapped it with this instead of debugging Hive boxes

cw_core/lib/db/sqlite.dart
```diff
@@ -41,7 +41,7 @@ Future<void> initDb({String? pathOverride}) async {
     }
   }
   await db?.close();
-  db = await openDatabase(dbFile.path, version: 6,
+  db = await openDatabase(dbFile.path, version: 7,
```
OmarHatem28: fix the conflict with dev, and update this to 8

lib/di.dart
```diff
@@ -1126,22 +1118,22 @@ Future<void> setup({
 
   getIt.registerFactory(() => AddressListPage(getIt.get<WalletAddressListViewModel>()));
 
+  getIt.registerLazySingleton(() {
+    final appStore = getIt.get<AppStore>();
+    return NodeListViewModel(appStore);
+  });
```
OmarHatem28, line 1124: why singleton?

lib/di.dart
```diff
@@ -1126,22 +1118,22 @@ Future<void> setup({
...
-    (showHidden, _) => NewAddressesPage(
+        (showHidden, _) => NewAddressesPage(
       showHidden: showHidden,
       addressListViewModel: getIt<WalletAddressListViewModel>(),
       dashboardViewModel: getIt<DashboardViewModel>(),
     ),
   );
 
-  getIt.registerFactory(() {
+  getIt.registerLazySingleton(() {
```
OmarHatem28, line 1134: same

lib/entities/default_settings_migration.dart
```diff
@@ -196,46 +188,43 @@ Future<void> defaultSettingsMigration(
...
-          await addWalletNodeList(nodes: nodes, type: WalletType.haven);
+          await addWalletNodeList(type: WalletType.haven);
           await _changeDefaultNode(
-            nodes: nodes,
             sharedPreferences: sharedPreferences,
             type: WalletType.haven,
             currentNodePreferenceKey: PreferencesKey.currentHavenNodeIdKey,
           );
-          await checkCurrentNodes(nodes, powNodes, sharedPreferences);
+          // await checkCurrentNodes(sharedPreferences);
```
OmarHatem28, line 215: Why is every call commented? at least call it once in the last version

cw_core/lib/node_legacy.dart
```diff
@@ -0,0 +1,600 @@
...
+    return _keyIndex;
+  }
+
+  WalletType get type => deserializeFromInt(typeRaw);
+
+  set type(WalletType type) => typeRaw = serializeToInt(type);
+
+  dynamic _keyIndex;
+
+  Future<bool> requestNode() async {
```
OmarHatem28: Are all of these functions no longer used? If so, why don't you remove them?

lib/entities/default_settings_migration.dart
```diff
@@ -937,16 +904,19 @@ Future<void> replaceNodesMigration({required Box<Node> nodes}) async {
...
 }
 
-Node? getDefaultNode({required Box<Node> nodes, required WalletType type}) {
+Future<Node?> getDefaultNode({required WalletType type}) async {
+  final nodes = await Node.getAll();
   final defaultUri = _getDefaultNodeUri(type);
-  return nodes.values.firstWhereOrNull((Node node) => node.uriRaw == defaultUri) ??
-      nodes.values.firstWhereOrNull((node) => node.type == type);
+  return nodes.firstWhereOrNull((Node node) => node.uriRaw == defaultUri) ??
+      nodes.firstWhereOrNull((node) => node.type == type);
```
OmarHatem28: why not use SQL where to do that?

lib/entities/node_check.dart
```diff
@@ -0,0 +1,69 @@
...
+  final preferenceKey = isPow ? powNodePreferenceKeys[type] : nodePreferenceKeys[type];
+
+  if(preferenceKey == null) {
+    return;
+  }
+
+  final currentId = await sharedPreferences.getInt(preferenceKey);
+
+  final nodes = isPow ? await Node.getAllForWalletTypePow(type) : await Node.getAllForWalletType(type);
+  final currentNode = nodes.firstWhereOrNull((item)=>item.id == currentId);
```
OmarHatem28, line 44: In the migration, I don't see us migrating the ID, then how are you expecting that It will find a value here?

cw_core/lib/node.dart
```diff
@@ -50,62 +79,144 @@ class Node extends HiveObject with Keyable {
...
-        isEnabledForAutoSwitching = map['isEnabledForAutoSwitching'] as bool? ?? false;
+        isPow = (map["isPow"] != null && map['isPow'] != 0) as bool? ?? false,
+        useSSL = (map['useSSL'] != 0) as bool?,
+        typeRaw = (map["typeRaw"] ?? 0) as int,
+        trusted = (map['trusted'] != 0) as bool? ?? false,
+        socksProxyAddress = map['socksProxyAddress'] as String?,
+        isEnabledForAutoSwitching = (map['isEnabledForAutoSwitching'] != 0) as bool? ?? false,
+        isOfficial = (map['isOfficial'] != 0) as bool? ?? false,
+        isBuiltin = (map['isBuiltin'] != 0) as bool? ?? false,
+        isDefault = (map['isDefault'] != 0) as bool? ?? false;
```
OmarHatem28, line 103: so, `map['x'] != 0` and all of those are booleans, and the toMap saves them as boolean, not int, so if they are false, this condition will evaluate true

lib/entities/default_settings_migration.dart
malik1004x: `checkCurrentNodes` is now being called on app startup (`main.dart`) rather than inside migrations. This way, if the database somehow is messed up, it will always fix itself before attempting a connection.

lib/entities/default_settings_migration.dart
malik1004x: I do have a where-based function to do this, just don't use it here. Good catch though, this doesn't work properly as it relies on hardcoded strings rather than YAML `isDefault`. I'll just switch over the uses to the correct function and remove this one.

lib/di.dart
malik1004x: It's a fix for the node list not refreshing – it relies on being updated by `NodeCreateOrEditViewModel`, which previously had a bug where it updated only its own, separate instance. I actually restructured this approach as part of the new Node UI PR, so this fix can be considered temporary.

cw_core/lib/node.dart
malik1004x: From the [SQLite documentation](https://sqlite.org/datatype3.html#boolean_datatype):

> SQLite does not have a separate Boolean storage class. Instead, Boolean values are stored as integers 0 (false) and 1 (true).

This has been verified and has correct behavior. Even if you explicitly save `false` to the database and then get it back, it'll be returned as `0`.

lib/entities/default_settings_migration.dart
```diff
@@ -678,19 +646,21 @@ Future<void> _changeDefaultNode({
...
+
   final currentNodeId = sharedPreferences.getInt(currentNodePreferenceKey);
   final bool shouldReplace;
   if (currentNodeId == null) {
     shouldReplace = true;
   } else {
-    final currentNode = nodes.values.firstWhere((node) => node.key == currentNodeId);
-    shouldReplace = oldUri?.any((e) => currentNode.uriRaw.contains(e)) ?? true;
+    final currentNode = nodes.firstWhereOrNull((node) => node.id == currentNodeId);
+    shouldReplace = oldUri?.any((e) => currentNode?.uriRaw.contains(e) ?? true) ?? true;
```
OmarHatem28:
```suggestion
    shouldReplace = currentNode == null || oldUri?.any((e) => currentNode!.uriRaw.contains(e)) ?? true;
```

OmarHatem28 (approved):

OmarHatem28:
continuing in #3258

PR #2825, opened 2026-01-21
CW-1157-Use-correct-derivation-paths-for-other-address-types

serhii-bor:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_bitcoin/lib/bitcoin_address_record.dart
```diff
@@ -76,21 +79,38 @@ class BitcoinAddressRecord extends BaseBitcoinAddressRecord {
     }
 }
 
+  static bool _legacyDefaultForType(BitcoinAddressType type) {
+
+    // Some address types (p2wpkh, p2wsh) were historically derived from the same account/path as our new standard
+    // but using legacy formats. For these, we default to legacy = false.
+    if (type == SegwitAddresType.p2wpkh || type == SegwitAddresType.p2wsh) return false;
```
OmarHatem28, line 86: shouldn't it be only p2wpkh?
p2wsh should have another derivation path, right?

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -318,13 +419,13 @@ abstract class ElectrumWalletBase
...
-      wif = WifEncoder.encode(hd.privateKey.raw, netVer: network.wifNetVer);
+      wif = WifEncoder.encode(mainHd.privateKey.raw, netVer: network.wifNetVer);
     } catch (_) {}
     try {
-      privateKey = hd.privateKey.toHex();
+      privateKey = mainHd.privateKey.toHex();
     } catch (_) {}
     try {
-      publicKey = hd.publicKey.toHex();
+      publicKey = mainHd.publicKey.toHex();
```
OmarHatem28: shouldn't those use `_hdFor` not directly `mainHd`?
if I understand correctly, `mainHd` shouldn't be used directly anymore, and instead we should always call `_hdFor`

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -2939,6 +3055,15 @@ abstract class ElectrumWalletBase
...
     }
   }
+
+  Bip32Slip10Secp256k1 _hdFor({required BaseBitcoinAddressRecord record}) {
+    final addrType = record.type;
+    return  record.isLegacyDerivation
+        ? (record.isHidden ? walletAddresses.legacySideHd : walletAddresses.legacyMainHd)
+        : (record.isHidden
+        ? (sideHdByType[addrType] ?? sideHd)
+        : (mainHdByType[addrType] ?? mainHd));
```
OmarHatem28: ternary operators are used for simple if conditions, here it would be better to use normal if/else conditions for better readability

cw_bitcoin/lib/electrum_wallet_addresses.dart
```diff
@@ -162,10 +166,23 @@ abstract class ElectrumWalletAddressesBase extends WalletAddresses with Store {
...
+    final typeMatchingAddresses = <BitcoinAddressRecord>[
+      ...typeMatchingAddressesAll.where((a) => !a.isLegacyDerivation),
+      ...typeMatchingAddressesAll.where((a) => a.isLegacyDerivation),
+    ];
+
+    final typeMatchingReceiveAddressesAll =
+        typeMatchingAddressesAll.where((addr) => !addr.isUsed).toList();
+    final typeMatchingReceiveAddresses = <BitcoinAddressRecord>[
+      ...typeMatchingReceiveAddressesAll.where((a) => !a.isLegacyDerivation),
+      ...typeMatchingReceiveAddressesAll.where((a) => a.isLegacyDerivation),
```
OmarHatem28, line 203: legacy addresses should be treated as hidden and should have hidden set to true.
so technically, they should be already excluded from `typeMatchingAddressesAll`
so that if the correct addresses with the correct derivation path are finished, it doesn't get from the old ones, but it should generate new correct ones.

**Edit:**
I see now that isHidden is for change addresses, which is not very good, is it also used when the user manually hide an address? would that affect addresses getting different derivation path set to them because isHidden is true?

cw_bitcoin/lib/electrum_wallet_addresses.dart
```diff
@@ -607,66 +645,96 @@ abstract class ElectrumWalletAddressesBase extends WalletAddresses with Store {
...
-      {BitcoinAddressType type = SegwitAddresType.p2wpkh}) async {
-    var countOfReceiveAddresses = 0;
-    var countOfHiddenAddresses = 0;
+    Future<void> _generateInitialAddresses(
+        {BitcoinAddressType type = SegwitAddresType.p2wpkh,
+          bool isLegacyDerivation = false }) async {
+
+      // Legacy derivation produces the same addresses as standard for these types.
+      // Don't generate a legacy set to avoid duplicates.
+      if (isLegacyDerivation && (type == SegwitAddresType.p2wpkh || type == SegwitAddresType.p2wsh)) {
```
OmarHatem28, line 712: same here, isn't p2wsh different than p2wpkh

cw_bitcoin/lib/bitcoin_address_record.dart
serhii-bor: no, not really. In our implementation, P2WSH is just a wrapper around P2WPKH, so it uses the same derivation path, but the script construction is different (hash160 vs sha256), the address remains the same.

P2WSH is usually used for multisig wallets, but our current implementation of P2WSH isn’t fully ready for that. If we want to support multisig later, we should probably add it as a separate wallet type, not just another address type. So for now we can safely leave it as it is

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: yes, mainHd represents the legacy account. It would be better to use mainHdByType[SegwitAddresType.p2wpkh] here instead. The resulting keys remain the same, so this change just makes the code align with the new derivation structure.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: done

cw_bitcoin/lib/electrum_wallet_addresses.dart
serhii-bor: I don’t want to mark those addresses as hidden here because isHidden is used for change addresses. There is a separate UI option for hiding addresses, which is a completely different feature.

cw_bitcoin/lib/electrum_wallet_addresses.dart
serhii-bor: the address remains the same

OmarHatem28 (approved):

PR #2827, opened 2026-01-22
CW-1228: Automatically detect wallet tokens for EVM chains

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_evm/lib/evm_chain_wallet.dart
```diff
@@ -481,6 +508,49 @@ abstract class EVMChainWalletBase
...
+
+      final existingAddresses =
+          evmChainErc20TokensBox.values.map((t) => t.contractAddress.toLowerCase()).toSet();
+
+      final List<Erc20Token> discoveredTokens = [];
+
+      for (final token in walletTokens) {
+        final addr = token.contractAddress.toLowerCase();
+
+        if (existingAddresses.contains(addr)) continue;
```
OmarHatem28: but if we have the token but disabled, why not enable it for the user?
i.e if the user has balance in PEPE for example, and we have PEPE as default token, but disabled, then it should be enabled here

OmarHatem28 (approved):

PR #2828, opened 2026-01-22
CW-1260: Enable Swap all for EVM wallets and persist amount when currency changes

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -848,6 +948,62 @@ abstract class ExchangeViewModelBase extends WalletChangeListenerViewModel with
     _setAvailableProviders();
   }
 
+  @action
+  Future<void> _onPairChangeWithAmountPreservation({
```
OmarHatem28: I don't think this is needed, beside that it adds more complexity to a flow that is not needed that much, I think just disabling the reset of the amount field when changing the currencies and triggering the load limits and calculate best rate would be enough

lib/src/screens/exchange/exchange_page.dart
```diff
@@ -636,8 +636,6 @@ class ExchangePage extends BasePage {
     key.currentState!.changeAddress(
         address:
             isCurrentTypeWallet ? exchangeViewModel.wallet.walletAddresses.addressForExchange : '');
-
-    key.currentState!.changeAmount(amount: '');
```
OmarHatem28, line 640: what's the reason for this change, and if it's intended, then please apply it on the new swap page

lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -354,15 +357,17 @@ abstract class ExchangeViewModelBase extends WalletChangeListenerViewModel with
...
+  bool get hasAllAmount {
+    if ([
+      WalletType.monero,
+      WalletType.bitcoin,
+      WalletType.litecoin,
+      WalletType.bitcoinCash,
+      WalletType.dogecoin,
+    ].contains(wallet.type)) return depositCurrency == wallet.currency;
+
+    return isEVMCompatibleChain(wallet.type);
```
OmarHatem28: technically, like this, the user can select any currency he wants from an EVM wallet and he would have ALL enabled, that's why the `depositCurrency == wallet.currency` condition was added.

for EVM, we probably would have to check that depositCurrency.tag == wallet.currency.title or so, but take care of BNB and BSC and if there is other exceptional cases.

and I see you do this checks later and return 0, but let's not show the button in the first place if it's not an evm currency

lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -813,6 +856,63 @@ abstract class ExchangeViewModelBase extends WalletChangeListenerViewModel with
       final amount = await unspentCoinsListViewModel.getSendingBalance(UnspentCoinType.any);
 
       changeDepositAmount(amount: wallet.formatCryptoAmount(amount.toString()));
+    } else if (isEVMCompatibleChain(wallet.type)) {
+      final balanceCurrency = wallet.balance.keys.firstWhereOrNull(
+        (currency) =>
+            currency.title == depositCurrency.title &&
+            (currency.tag == depositCurrency.tag || currency.tag == depositCurrency.title),
```
OmarHatem28, line 1317: here as well, take into consideration BNB and BSC

lib/view_model/exchange/exchange_view_model.dart
```diff
@@ -813,6 +856,63 @@ abstract class ExchangeViewModelBase extends WalletChangeListenerViewModel with
...
+
+      if (!isNative) changeDepositAmount(amount: balanceAmount);
+
+      try {
+        final priority = _settingsStore.getPriority(wallet.type, chainId: wallet.chainId);
+
+        if (priority == null) {
+          changeDepositAmount(amount: balanceAmount);
+          return;
+        }
```
OmarHatem28: having a null priority doesn't mean it doesn't have fee, Arbitrum doesn't have priority fees, but still has fees
```suggestion
```

lib/view_model/exchange/exchange_view_model.dart
Blazebrain: Alright, I've removed the function and simplified it

lib/src/screens/exchange/exchange_page.dart
Blazebrain: It keeps the amount when switching currencies so the opposite amount can be recalculated with the new rate

I've added it to the new swap page

lib/view_model/exchange/exchange_view_model.dart
Blazebrain: Okayy

I've updated it so EVM only shows ALL when the deposit currency is native or an ERC20 in the wallet balance

lib/view_model/exchange/exchange_view_model.dart
Blazebrain: Fixed it. We now pass null to updateEstimatedFeesParams(priority) when priority is null, for cases like Arbitrum

That way the wallet uses base fee with priorityFee = 0, so the fees are still calculated correctly.

lib/view_model/exchange/exchange_view_model.dart
OmarHatem28: still not what I mean, right now you are doing exactly the same as what's in the `_onPairChange()` so why not just remove the lines that resets the amounts in this function
if it's a fixed rate, clear the deposit field only, else clear the receive only
and just force a re-calculation of best rate

Blazebrain: Are you referring to the `hasAmount` or another function?

None of those happen in it

lib/view_model/exchange/exchange_view_model.dart
OmarHatem28, line 1: @konstantinullrich please merge this into your Money PR and make sure the changes here are also migrated properly to the new Money class

OmarHatem28 (approved):

PR #2845, opened 2026-01-29
Lightning Username setup flow for new UI

malik1004x:

# Description

Allows creating and changing the Lightning username in the new UI settings and while creating a new Bitcoin wallet. Includes Bloc-based Viewmodel for the associated page.

# Pull Request - Checklist  

- [x] Format code


lib/src/screens/seed/seed_verification/seed_verification_success_view.dart
```diff
@@ -62,7 +65,11 @@ class SeedVerificationSuccessView extends StatelessWidget {
           PrimaryButton(
             key: ValueKey('wallet_seed_page_open_wallet_button_key'),
             onPressed: () {
-              Navigator.of(context).popUntil((route) => route.isFirst);
+              if (walletType == "Bitcoin") {
```
OmarHatem28: again, string comparison, why have string comparisons when you have enums!!

lib/new-ui/viewmodels/lightning_username/lightning_username_event.dart
```diff
@@ -0,0 +1,20 @@
...
+  UsernameChanged(this.newUsername);
+}
+
+final class UsernameSaveRequested extends LightningUsernameEvent {}
+
+final class UsernameSaveError extends LightningUsernameEvent {
+  final String error;
+
+  UsernameSaveError(this.error);
+}
```
OmarHatem28: that's a state not an event, and is not used.
also when using blocs, try to name events as actions, like `ChangeUserName` `GenerateRandomUsername` etc...

OmarHatem28:
for the import failure in the CI, we shouldn't use `cw_bitcoin` packages directly, that's why we have the proxy layers, check how it is used in other places and fix it please

lib/new-ui/viewmodels/lightning_username/lightning_username_bloc.dart
```diff
@@ -0,0 +1,88 @@
...
+final Map<String, String> errorPatterns = {"name already taken": S.current.username_not_available};
+
+const usernameRegex = r"^[a-z0-9_.-]+$";
+const usernameSuffix = "@cake.cash";
+
+class LightningUsernameBloc extends Bloc<LightningUsernameEvent, LightningUsernameState> {
+  final WalletBase _wallet;
+
+  LightningUsernameBloc(this._wallet) : super(LightningUsernameInitial("")) {
+    on<UsernameChanged>((event, emit) async {
```
OmarHatem28: **Enhancement**
Events are actions, so better call them as such, and extract them as functions
```suggestion
    on<ChangeUsername>((event, emit) async => _changeUsername());
    on<SaveUsername>((event, emit) async => _saveUsername());
    etc....
```

lib/new-ui/viewmodels/lightning_username/lightning_username_bloc.dart
```diff
@@ -0,0 +1,88 @@
...
+        bitcoin!.setLightningUsername(_wallet, state.username);
+        emit(LightningUsernameSaved(state.username));
+      } catch (e) {
+        for (final pattern in errorPatterns.keys) {
+          if (e.toString().contains(pattern)) {
+            emit(LightningUsernameError(state.username, UsernameError(errorPatterns[pattern]!)));
+            return;
+          }
+        }
+        emit(LightningUsernameError(state.username, UsernameError(S.current.unknown_error)));
```
OmarHatem28: I would rather show the error so if we get a report we know where to look
```suggestion
        emit(LightningUsernameError(state.username, UsernameError("${S.current.unknown_error}\n${e.toString()}")));
```

lib/new-ui/pages/lightning_username_page.dart
```diff
@@ -0,0 +1,317 @@
...
+import 'package:cw_core/generate_name.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:flutter_svg/flutter_svg.dart';
+
+class LightningUsernamePage extends StatefulWidget {
+  LightningUsernamePage({super.key, required this.isSetup});
+
+  final bool isSetup;
+  final bool isLightMode = !(getIt.get<ThemeStore>().currentTheme.isDark);
```
OmarHatem28: same thing with di

lib/new-ui/pages/lightning_username_page.dart
```diff
@@ -0,0 +1,317 @@
...
+  @override
+  void initState() {
+    super.initState();
+    _editMode = widget.isSetup;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return BlocProvider(
+      create: (context) => getIt.get<LightningUsernameBloc>(),
```
OmarHatem28: **Enhancement:**
it would be cleaner to pass it as a dependency, that's why we have the di.dart (dependency injection.dart) instead of just defining it there and using it in multiple place
I know that we don't have much testing anyway, but still cleaner, and when we add testing for this area we don't have many changes needed

lib/new-ui/pages/lightning_username_page.dart
```diff
@@ -0,0 +1,317 @@
...
+              Navigator.of(context).pop();
+            }
+          }
+          if (state is LightningUsernameInitial) {
+            if (!widget.isSetup) {
+              _controller.text = state.username;
+            }
+          }
+        },
+        child: BlocBuilder<LightningUsernameBloc, LightningUsernameState>(
```
OmarHatem28: **Minor:**
you can use BlocConsumer, it has both listener and builder attributes, so no need for nested children

lib/src/screens/seed/seed_verification/seed_verification_success_view.dart
malik1004x: This is how the wallet type was passed into this flow originally. Decided to do it like this rather than refactoring legacy code, but I can change it.

lib/new-ui/viewmodels/lightning_username/lightning_username_bloc.dart
malik1004x: Pragmatically, that's the best solution. But these errors are 40+ lines long, and the UI is made for a single line of text. This is why i tried matching them with string lookup rather than direct comparison in the first place.

lib/src/screens/seed/seed_verification/seed_verification_success_view.dart
OmarHatem28: ok yeah, please change it, it's weird it's passed as a string, we don't have that anywhere else

lib/new-ui/viewmodels/lightning_username/lightning_username_bloc.dart
OmarHatem28: can you send me on Slack an example of an error they send

lib/new-ui/pages/lightning_username_page.dart
```diff
@@ -0,0 +1,315 @@
...
+  @override
+  void initState() {
+    super.initState();
+    _editMode = widget.isSetup;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return BlocProvider(
+      create: (context) => getIt.get<LightningUsernameBloc>(),
```
OmarHatem28: you add it, use it 😅 

```suggestion
      create: (context) => widget.lightningUsernameBloc,
```

OmarHatem28 (approved):

PR #2859, opened 2026-02-04
Account Customizer for new UI Monero

malik1004x:

# Description

Allows adding, reordering, renaming and customizing accounts for Monero wallets in the new UI. Also includes a "compact" mode for the cards on the dashboard, activated when there's over 3 accounts.

# Pull Request - Checklist  

- [x] Format code


lib/new-ui/pages/account_customizer.dart
```diff
@@ -0,0 +1,388 @@
...
+  Widget build(BuildContext context) {
+    if (_items.isEmpty) return SizedBox.shrink();
+
+    return PopScope(
+      canPop: false,
+      onPopInvokedWithResult: (didPop, result) {
+        saveCardOrder().then((value) {
+          widget.dashboardViewModel.loadCardDesigns();
+          Navigator.of(context).popUntil((route) => route.isFirst);
+        });
```
OmarHatem28: do we show any loading state to the user or he will think the app is lagging?

lib/new-ui/pages/account_customizer.dart
```diff
@@ -0,0 +1,456 @@
...
+
+  }
+
+  @override
+  void dispose() {
+    saveCardOrder().then((value) => widget.dashboardViewModel.loadCardDesigns());
+    super.dispose();
+  }
+
+  Future<void> loadCards() async {
```
OmarHatem28: I see no reason for it to be async
```suggestion
  void loadCards() {
```

lib/new-ui/widgets/coins_page/cards/cards_view.dart
```diff
@@ -27,36 +27,53 @@ class CardsView extends StatefulWidget {
...
 class _CardsViewState extends State<CardsView> {
-  int? _selectedIndex = 0;
+  late int _selectedIndex;
+
+  @override
+  void initState() {
+    super.initState();
+    setState(() {
+      _selectedIndex = widget.dashboardViewModel.cardOrder.length - 1;
+    });
```
OmarHatem28: setState in initState?

OmarHatem28 (changes requested):

lib/new-ui/pages/account_customizer.dart
malik1004x: This happens so quick there's no lag, even in debug mode on an emulator.

lib/new-ui/pages/account_customizer.dart
malik1004x: This was async because this function used to do more, but now you are right.

OmarHatem28 (approved):

PR #2860, opened 2026-02-04
New UI fixes

malik1004x:

# Description

Fixes many minor issues with new UI.

# Pull Request - Checklist  

- [x] Format code


lib/new-ui/widgets/coins_page/cards/cards_view.dart
```diff
@@ -98,13 +96,15 @@ class _CardsViewState extends State<CardsView> {
             final CardDesign cardDesign;
             if (widget.dashboardViewModel.cardDesigns.isEmpty)
               cardDesign = CardDesign.genericDefault;
+            else if(widget.lightningMode)
+              cardDesign = widget.dashboardViewModel.cardDesigns[index + 1];
```
OmarHatem28, line 102: this won't throw an out of range exception?
can't check the whole flow right now, so just asking

lib/new-ui/widgets/coins_page/cards/cards_view.dart
```diff
@@ -162,7 +162,9 @@ class _CardsViewState extends State<CardsView> {
         final parentWidth = MediaQuery.of(context).size.width;
         final children = <Widget>[];
 
-        int numCards = widget.dashboardViewModel.cardDesigns.length;
+    int numCards = widget.dashboardViewModel.wallet.type == WalletType.bitcoin
+        ? 1
+        : widget.dashboardViewModel.cardDesigns.length;
```
OmarHatem28, line 170: why specifically bitcoin is 1? aren't all wallets except monero should be 1?

lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -397,16 +398,37 @@ abstract class DashboardViewModelBase with Store {
...
+            .where((e) => e.accountIndex == index)
             .firstOrNull;
 
-        cardDesigns.add(CardDesign.fromStyleSettings(setting, wallet.currency));
+        late final CryptoCurrency curr;
+        if(wallet.type == WalletType.bitcoin && i == 1) {
+          curr = CryptoCurrency.btcln;
+        } else {
+          curr = wallet.currency;
+        }
```
OmarHatem28, line 428: can you explain the logic here in some comments or better code readability

lib/new-ui/pages/coin_control_page.dart
```diff
@@ -194,7 +195,8 @@ class CoinControlListSection extends StatelessWidget {
       itemBuilder: (_, int index) {
         return Observer(builder: (_) {
           final item = items[index];
-          final fiatAmount = unspentCoinsListViewModel.fiatAmounts[item.hash] ?? '';
+          final fiatAmount = unspentCoinsListViewModel.fiatAmounts[item.amount] ?? '';
```
OmarHatem28, line 198:
```suggestion
          final fiatAmount = unspentCoinsListViewModel.fiatAmounts[item.hash] ?? '';
```

lib/new-ui/widgets/coins_page/cards/cards_view.dart
malik1004x: This is just so it doesn't try to load the Lightning card design as another account. For all other wallets except Bitcoin and Monero, `widget.dashboardViewModel.cardDesigns.length == 1`.

lib/new-ui/pages/coin_control_page.dart
malik1004x: This doesn't work. ViewModel doesn't properly hash these.

lib/view_model/dashboard/dashboard_view_model.dart
malik1004x: This flow was designed with the "one account - one card" rule in mind. Bitcoin now requires 2 cards (regular BTC and Lightning). I use accountIndex to differentiate between the two. Normally, for single-account wallets accountIndex is always -1, so for Lightning I used index 0. The index selection at the top chooses the actual account number for monero, 0 for Lightning (which is the second card) and -1 for anything else.

The Currency selection has a special case now because Lightning is the only card design that exists within another currency's wallet.

lib/new-ui/widgets/coins_page/cards/cards_view.dart
malik1004x: It won't. Bitcoin only draws one card at a time, so index is equal to 0 here. `cardDesigns` for Bitcoin with Lightning holds the regular Bitcoin design at index 0 and Lightning at index 1.

PR #2919, opened 2026-02-18
Fix async handling in Lightning transaction history

konstantinullrich:
# Description

The `fetchTransactions` method was updated to resolve async handling issues related to Lightning transaction history. This ensures smoother and more reliable transaction data retrieval.

# Pull Request - Checklist  

- [x] Format code


OmarHatem28 (approved):
I would also add .onError so if anything goes wrong with lightning wallet function, it doesn't affect normal flow

PR #2920, opened 2026-02-19
New design small fixes

tuxpizza:


lib/new-ui/widgets/coins_page/top_bar_widget/top_bar.dart
```diff
@@ -25,7 +25,7 @@ class TopBar extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return Padding(
-      padding: const EdgeInsets.only(left: 18.0, right: 18.0, top: 10.0),
+      padding: const EdgeInsets.only(left: 18.0, right: 18.0, top: 10.0, bottom: 10.0),
```
malik1004x:
```suggestion
      padding: const EdgeInsets.symmetric(horizontal: 18, vertical: 10),
```

lib/new-ui/widgets/confirm_swiper.dart
```diff
@@ -65,7 +65,10 @@ class _ConfirmSwiperState extends State<ConfirmSwiper> {
                   color: Theme.of(context).colorScheme.surfaceContainerLowest,
                   borderRadius: BorderRadius.circular(99999),
                 ),
-                child: Center(child: FlowingText(text: widget.swiperText, opacity: 1 - (drag / (areaWidth - pillSize)))),
+                child: Padding(
+                  padding: EdgeInsets.only(left: (((areaWidth / 2) - (widget.swiperText.length / 2) + 0.4 * (pillSize / 2))) - MediaQuery.of(context).size.width * 0.23 - pillSize),
```
malik1004x, line 69: what happened here? why would you add padding to a centered widget? and why does the padding look like this?

tuxpizza: This is for optical alignment of the "Swipe to send" text since the swipe circle makes it look off center. This allows it to be more optically aligned for all languages (not just english).

PR #2943, opened 2026-02-22
New design beta fixes

malik1004x:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/android/app/src/main/AndroidManifestBase.xml
+++ b/android/app/src/main/AndroidManifestBase.xml
@@ -117,6 +117,8 @@
                 <data android:scheme="bsc_wallet" />
                 <data android:scheme="zcash" />
                 <data android:scheme="zcash-wallet" />
+                <data android:scheme="lightning" />
+                <data android:scheme="lightning-wallet" />
             </intent-filter>
             <!-- nano-gpt link scheme -->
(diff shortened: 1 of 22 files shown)


android/app/src/main/AndroidManifestBase.xml
OmarHatem28, line 1: I don't think this would be enough, please check what the scheme of the link Czarek used and add the scheme here and in iOS as well

lib/new-ui/pages/swap_page.dart
```diff
@@ -246,6 +248,7 @@ class _NewSwapPageState extends State<NewSwapPage> {
           //             ExchangeProviderDescription.chainflip)
           //     ? Navigator.of(context).pushReplacementNamed(Routes.exchangeTrade)
           //     : Navigator.of(context).pushReplacementNamed(Routes.exchangeConfirm);
+          widget.exchangeViewModel.tradeStarted = true;
```
OmarHatem28, line 251: shouldn't we reset it to false if the process is cancelled by the user or failed halfway or so?

malik1004x: Once the confirmation modal is open you cannot closing without exiting the Swap module, so this will never happen.

PR #2968, opened 2026-02-25
fix: Desktop wallet bugs

Blazebrain:
- repeated wallets post migration
- newly created wallets not showing up
- extra large components on desktop, needed maxwidth

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/main.dart
```diff
@@ -281,6 +281,7 @@ Future<void> initializeAppConfigs({bool loadWallet = true}) async {
     CakeHive.registerAdapter(TronTokenAdapter());
   }
   await performHiveMigration();
+  await WalletInfo.removeDuplicates();
```
MrCyjaneK: What is causing these duplicates to appear in the first place?

Blazebrain: Possibly a migration issue. This particular fix is for a report triggered when an old desktop build was updated to the latest build (5.8.0 as at then), causing wallets to appear multiple times in the desktop wallet list, pushing at it likely being a migration issue. CW-1407 explains in details.

MrCyjaneK: Consider looking at #2969 to make it a one-time fix instead of running that every time app starts

MrCyjaneK: If migration is part of the issue it will point us to a place calling the INSERT instead of UPDATE, fixing the real bug instead of just applying a hotfix (that would break things like address book and addresses.

OmarHatem28 (changes requested):
I merged Czarek's PR, it handles the duplicate wallets with a better approach, so please remove the workaround for it.
everything else looks good

lib/view_model/wallet_list/wallet_list_view_model.dart
```diff
@@ -88,14 +87,16 @@ abstract class WalletListViewModelBase with Store {
 
   bool get ascending => _appStore.settingsStore.walletListAscending;
 
-  
-  bool isUpdating = false;
+  bool _isUpdating = false;
+  bool _needsUpdate = false;
+
   @action
   Future<void> updateList() async {
```
MrCyjaneK, line 98: Maybe use mutex here instead of recursively calling itself? Something may await it in future/now, and this flow is confusing when we consider function signature.

Blazebrain: Yeah, that's a better approach

I've switched it to a Future based mutex, so each caller would wait for the previous update call to finish, and then runs, removing the recursion.

PR #2983, opened 2026-02-27
feat: Auto fetch wallet tokens in solana wallets

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_solana/lib/solana_wallet.dart
```diff
@@ -631,6 +631,65 @@ abstract class SolanaWalletBase
...
+
+      for (final moralisToken in walletTokens) {
+        final mint = moralisToken.mint;
+
+        final existingToken = existingMints[mint];
+        if (existingToken != null) {
+          if (defaultMints.contains(mint) && !existingToken.enabled) {
+            existingToken.enabled = true;
+            await existingToken.save();
+            await addSPLToken(existingToken);
```
OmarHatem28, line 660: you add it here, and then you also add it in cw_solana, kinda repetitive but ok

OmarHatem28 (approved):

cw_solana/lib/solana_wallet.dart
Blazebrain: It's not repetitive, they do different things

The `addSPLToken` here only runs for existing tokens that are disabled, we never return it as part of the newly discovered tokens

The `addSPLToken` in cw_solana is for the newly discovered tokens that are returned from the previous function

So its not repetitive actually, they handle different scenarios

PR #2986, opened 2026-02-28
Fix minor issues in German translation

apertamono:
I checked the language quality for NL, DE, FR and ES. The German translation is pretty good and doesn't need to be reviewed further in my opinion, as a non-native speaker and professional translator. I fixed a few spelling and word order issues, an unusual language name in two genders, informal pronouns and inconsistent translations of the term Seed.


apertamono:
One thing I missed is that `"your": "Dein",` also features an informal pronoun. Which is the wrong gender for `Adresse`. See #2987 for context. Not a word that should be translated separately.

OmarHatem28 (approved):

PR #2987, opened 2026-02-28
Fix minor issue in French translation

apertamono:
I checked the language quality for NL, DE, FR and ES. The French translation is very good. The only thing I'm correcting here is a single use of the informal possessive pronoun, while formal pronouns are used in all other strings.


apertamono:
Actually, it would be nice to know the context in which the "your" string is used. But 'Votre' (masculinine/feminine singular) covers more options than 'Ton' (masculine singular).

apertamono:
I guess this is used in address_page.dart?

`'${S.current.your} ${addressListViewModel.walletTypeName} ${S.current.address}'`

That's a very English word order, by the way. In such cases, it would be preferable to translate a complete sentence, with placeholders for variables like the wallet type name.

OmarHatem28 (approved):

PR #2991, opened 2026-03-01
Fix minor issue in Spanish translation

apertamono:
There used to be a number of serious issues in the Spanish translation, but these have been fixed already. The only remaining issue I found is that "Yes, I have" meaning "Yes, I've written down the seed" is translated as "Yes, I own."


OmarHatem28 (approved):

PR #3002, opened 2026-03-02
New design post release fixes

malik1004x:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_bitcoin/lib/electrum_balance.dart
```diff
@@ -28,19 +28,21 @@ class ElectrumBalance extends Balance {
 
   int confirmed;
   int unconfirmed;
+  final int _frozen;
+
   @override
-  final int frozen;
+  BigInt get frozen => BigInt.from(_frozen);
```
OmarHatem28, line 34: why specifically is frozen here a big int?

cw_core/lib/wownero_balance.dart
```diff
@@ -4,7 +4,7 @@ import 'package:cw_core/wownero_amount_format.dart';
 class WowneroBalance extends Balance {
   WowneroBalance({required this.fullBalance, required this.unlockedBalance, this.frozenBalance = 0})
       : formattedFrozenBalance = wowneroAmountToString(amount: frozenBalance),
-        super(unlockedBalance, fullBalance - unlockedBalance, frozen: frozenBalance);
+        super(BigInt.zero, BigInt.zero, frozen: BigInt.zero);
```
OmarHatem28, line 7: we dropped support of creating/restoring wownero wallets, but we haven't fully removed it yet, so for existing users it's still working, and it should continue to work

lib/src/screens/settings/security_backup_page.dart
```diff
@@ -148,18 +148,18 @@ class SecurityBackupPage extends BasePage {
...
+                  //       keyValue: "security_backup_page_change_password_button_key",
+                  //       label: S.current.create_backup,
+                  //       onTap: () {
+                  //         _authService.authenticateAction(
+                  //           context,
+                  //           route: Routes.backup,
+                  //           conditionToDetermineIfToUse2FA: _securitySettingsViewModel
+                  //               .shouldRequireTOTP2FAForAllSecurityAndBackupSettings,
+                  //         );
+                  //       }),
```
OmarHatem28, line 162: is this moved to another place?

lib/src/widgets/bottom_sheet/swap_details_bottom_sheet.dart
```diff
@@ -112,6 +112,9 @@ class _SwapDetailsBottomSheetState extends State<SwapDetailsBottomSheet> {
                     buttonAction: () {
                       _showingFailureDialog = false;
                       Navigator.of(popupContext).pop();
+                      if(mounted) {
+                        Navigator.of(context, rootNavigator: true).pop();
+                      }
```
OmarHatem28, line 117: would this pop the swap screen as well if there is a failure?
if so, then we don't need that

lib/new-ui/pages/account_customizer.dart
```diff
@@ -293,6 +238,67 @@ class _AccountCustomizerState extends State<AccountCustomizer> {
...
+          return Material(child: modal);
+        });
+    if (res != null && res is bool && res == true) {
+      await widget.dashboardViewModel.loadCardDesigns();
+      loadCards();
+      await saveCardOrder();
+    }
+  }
+
+  void _openCardCustomizer() {
```
OmarHatem28, line 274: not the best way to implement this, but I want you to take a look at it later

lib/new-ui/widgets/coins_page/assets_history/history_section.dart
```diff
@@ -50,27 +62,28 @@ class HistorySection extends StatelessWidget {
...
-                    asset: asset,
-                ),
-              );
-            } else if (item is TradeListItem) {
-              final trade = item.trade;
+                    return GestureDetector(
+                      onTap: () => Navigator.of(context)
+                          .pushNamed(Routes.transactionDetails, arguments: transaction),
+                      child: HistoryTile(
+                        title: item.formattedTitle + transactionType,
```
OmarHatem28, line 73: why? this way it will not show the `1/10` confirmations status and the others

cw_bitcoin/lib/electrum_balance.dart
malik1004x: `frozen` is a field in the main `Balance` class and we can't override a `BigInt` with an `int`. Changing the actual field to `BigInt` would provide no advantages (since Electrum wallets are unlikely to ever cross int range) while being prone to errors.

lib/src/screens/settings/security_backup_page.dart
malik1004x: Yes, there's a separate Backup tab in Settings now, making this redundant.

lib/src/widgets/bottom_sheet/swap_details_bottom_sheet.dart
malik1004x: This is not in the swap screen, we don't use this popup anymore. This is in the Pay Anything flow which is the only place currently still utilizing the old swap confirmation modal. This code fixes an issue where, after swiping to send on this modal, the modal stayed on-screen and allowed infinitely swiping, suggesting the ability to send a swap transaction multiple times (which was not actually possible).

lib/new-ui/widgets/coins_page/assets_history/history_section.dart
malik1004x: This concatenation was moved to the `formattedTitle` getter inside `TransactionListItem`.

PR #3033, opened 2026-03-06
feat: Hide minimal style balance card icon

Blazebrain:

# Description

- Hide icon on minimal-style balance card by default
- Add toggle in card customizer to show it.

# Pull Request - Checklist  

- [ ] Format code


cw_core/lib/card_display_config.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:cw_core/card_design.dart';
+
+class CardDisplayConfig {
+  final CardDesign design;
+  final bool showIconOnCard;
+
+  const CardDisplayConfig({
+    required this.design,
+    this.showIconOnCard = false,
```
malik1004x, line 9: Why put this in a separate class? It complicates the logic and showing the icon is definitely design-related. This way, we have to change storage of CardDesign and pass an additional parameter on saving style settings.

lib/new-ui/pages/account_customizer.dart
```diff
@@ -91,12 +89,16 @@ class _AccountCustomizerState extends State<AccountCustomizer> {
...
+      if (index == null || index < 0 || index >= accounts.length) {
         // db order broken.
         reset();
         break;
       }
 
+      final cardConfigs = widget.dashboardViewModel.cardConfigs;
+      final design =
+          index < cardConfigs.length ? cardConfigs[index].design : CardDesign.genericDefault;
+      final showIconOnCard = index < cardConfigs.length ? cardConfigs[index].showIconOnCard : false;
```
malik1004x:
```suggestion
      final showIconOnCard = index < cardConfigs.length && cardConfigs[index].showIconOnCard;
```

lib/new-ui/pages/card_customizer.dart
```diff
@@ -220,7 +229,18 @@ class _CardCustomizerState extends State<CardCustomizer> {
                                   ],
                                 ),
                               ),
-                            )
+                            ),
+                            if (state.selectedDesign.backgroundType ==
+                                  CardDesignBackgroundTypes.svgIcon)
+                                SettingsSwitcherCell(
```
malik1004x: We don't use SettingsSwitcherCell anymore, it's from old UI. Please use ListItemCheckbox.

cw_core/lib/card_display_config.dart
Blazebrain: I'd rather we keep the config class. The reason is separation of concerns involved and also making it easy for us to extend this when needed without modifying the actual design definition for the card.

CardDesign class handles the design definitions, things like the gradient, background type, image type etc. 

the showIconOnCard on the other hand is purely a display preference, its not part of the card design itself.

If we decide to add more display preferences later, we can easily add them to the config class without modifying the CardDesign itself.

Also, we don't have to change the storage of CardDesign, we only pass the preference to the settings, as we need to save it somewhere.

Blazebrain:
UI adjusted this page to focus on adding a new design card instead of toggling the icon card. 

Closing this PR in respect to #3054

PR #3034, opened 2026-03-06
homescreen improvements

Blazebrain:
- Increase touch area for card editing icon
- Tapping any card in the background should move it ino the foreground without closing the account tab
- make top left chain icon primary color with full opacity when syncing and 20% when done
- add tor indicator to top bar when enabled

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/new-ui/widgets/coins_page/cards/balance_card.dart
```diff
@@ -261,12 +261,11 @@ class BalanceCard extends StatelessWidget {
               child: GestureDetector(
                 behavior: HitTestBehavior.opaque,
                 onTap: onCustomizeTapped,
-                child: Container(
-                  height: 48, width: 48, // for touch target
-                  child: Padding(
-                    padding: const EdgeInsets.only(bottom: 20, left: 20, top: 12, right: 12),
+                child: SizedBox(
```
malik1004x, line 264: Have you tested if this won't make the three dots icon large? I should note I've already increased the size of this touch target before, unless UI still deemed it too small after 6.0.1.

lib/new-ui/widgets/coins_page/top_bar_widget/sync_bar.dart
```diff
@@ -44,8 +45,20 @@ class SyncBar extends StatelessWidget {
...
+                  children: [
+                    CakeImageWidget(
+                      imageUrl: "assets/new-ui/tor.svg",
+                      width: 20,
+                      height: 20,
+                    ),
+                    if (_showDot()) PulsingDot(),
+                  ],
+                ),
+              if (!dashboardViewModel.isTorEnabled && _showDot()) PulsingDot(),
```
malik1004x: Instantiating the same widget twice like this (in this case PulsingDot) may lead to unnecessary code duplication if we ever need to pass something to the widget. Have you considered just showing the row always and putting if-conditions in the row children?

lib/src/widgets/cake_image_widget.dart
```diff
@@ -39,7 +41,7 @@ class CakeImageWidget extends StatelessWidget {
               imageUrl!,
               height: height,
               width: width,
-              color: color,
+              colorFilter: colorFilter,
```
malik1004x:
```suggestion
              colorFilter: colorFilter ?? (color == null ? null ColorFilter.mode(color, BlendMode.srcIn)),
```

This way if we just want to change the image color (most use cases), we can pass it the same way for SVG and PNG.

lib/new-ui/widgets/coins_page/cards/balance_card.dart
Blazebrain: Yeah, I've tested it. The icon doesn't get large. It was still left unchecked on the notion doc, so I assume UI wanted some more adjustments

lib/src/widgets/cake_image_widget.dart
Blazebrain: Yeah, that makes sense. Pushing an update with it

OmarHatem28 (approved):

PR #3046, opened 2026-03-10
fix: Minor swap fixes

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/exchange/provider/changenow_exchange_provider.dart
```diff
@@ -116,9 +115,12 @@ class ChangeNowExchangeProvider extends ExchangeProvider {
       final uri = Uri.https(apiAuthority, estimatedAmountPath, params);
       final response = await ProxyWrapper().get(clearnetUri: uri, headers: headers);
 
+      if (response.statusCode != 200) return 0.0;
```
OmarHatem28: I don't wanna risk it if they for any reason return another success status code, although 200 is the correct one, but this is a third party after all.
if it failed to parse, it will throw an exception, get caught by the try/catch and return 0 already
```suggestion
```

lib/exchange/provider/trocador_exchange_provider.dart
```diff
@@ -135,12 +134,21 @@ class TrocadorExchangeProvider extends ExchangeProvider {
       final uri = await _getUri(newRatePath, params);
       final response = await ProxyWrapper().get(clearnetUri: uri, headers: {'API-Key': apiKey});
 
+      if (response.statusCode != 200) return 0.0;
```
OmarHatem28: same
```suggestion
```

lib/exchange/provider/swaptrade_exchange_provider.dart
```diff
@@ -62,19 +61,23 @@ class SwapTradeExchangeProvider extends ExchangeProvider {
...
+      final normalized = _normalizeCurrency(from);
+      final coin = coinsInfo.cast<Map<String, dynamic>>().firstWhere(
+            (c) => (c['id']?.toString().toUpperCase() ?? '') == normalized,
+        orElse: () => <String, dynamic>{},
       );
 
-      if (coin == null) throw Exception('Coin not found: ${_normalizeCurrency(from)}');
+      if (coin.isEmpty) {
+        // Currency not supported by SwapTrade (e.g. USDC, DOGE).
+        return Limits(min: double.maxFinite, max: 0.0);
```
OmarHatem28: since we are refactoring this a bit and improving it, instead of providing some unreasonable values for failed limits, I would prefer that we return null on fetching limits issues, and in the exchange view model, we only check the limits that are actually returned

lib/exchange/provider/changenow_exchange_provider.dart
```diff
@@ -81,18 +80,18 @@ class ChangeNowExchangeProvider extends ExchangeProvider {
       throw Exception('Unexpected http status: ${response.statusCode}');
 
     final responseJSON = json.decode(response.body) as Map<String, dynamic>;
-    return Limits(
-        min: responseJSON['minAmount'] as double?, max: responseJSON['maxAmount'] as double?);
+    final min = double.tryParse(responseJSON['minAmount']?.toString() ?? '');
+    final max = double.tryParse(responseJSON['maxAmount']?.toString() ?? '');
+    return Limits(min: min, max: max == 0 ? null : max);
```
OmarHatem28: null means that it doesn't have a maximum limit, which is incorrect if the max here is 0

lib/exchange/provider/trocador_exchange_provider.dart
```diff
@@ -104,10 +104,9 @@ class TrocadorExchangeProvider extends ExchangeProvider {
 
     final coinJson = responseJSON.first as Map<String, dynamic>;
 
-    return Limits(
-      min: coinJson['minimum'] as double?,
-      max: coinJson['maximum'] as double?,
-    );
+    final min = double.tryParse(coinJson['minimum']?.toString() ?? '');
+    final max = double.tryParse(coinJson['maximum']?.toString() ?? '');
+    return Limits(min: min, max: max == 0 ? null : max);
```
OmarHatem28: same

OmarHatem28 (changes requested):

lib/exchange/provider/changenow_exchange_provider.dart
Blazebrain: Yeah, that's true. Better to be safe. I've removed the status code check

lib/exchange/provider/swaptrade_exchange_provider.dart
OmarHatem28: this is not handled yet, what I mean is that instead of returning Limits, we return null.
so the function signature should allow null
i.e `Future<Limits?>` and in case of errors we should return null, and in the exchange view model filter out the ones that return null
and if max is 0 then that should be considered a failure state as well, so return null instead of Limits

OmarHatem28 (approved):

PR #3048, opened 2026-03-10
Fix QR scanner to accept fountain coding for BC-UR codes

hectorchu:

# Description

Basically, the current BC-UR decoder assumes it is receiving a fixed set of frames. Each frame starts with `ur:psbt/m-n/` where n is the total number of frames and m is the sequence number of the frame. The current assumption is that the encoder produces the n frames and then shows them in sequence, cycling back to the first frame when it reaches the end of the sequence. The receiver must receive all n frames to decode the message.

The disadvantage of this process as described is that if the scanner misses a frame in the sequence, it needs to wait for the animation to cycle back to that frame. If the number of frames is large, this could take a while, maybe several seconds. Even if only one frame is missed, you still need to wait for the animation to reach the end and loop back to the beginning, until the frame is shown again.

But there is a better way. The designers of the UR codes knew about this, so they designed the code to be a "fountain coding". What this means is that, instead of only producing the fixed set of n frames, it keeps producing frames which are XOR combinations of the original fragments. There is encoded redundancy but that also allows you to work out missing frames in the main sequence, using the XOR frames. You may need to receive a number of frames greater than n to work out all the n frames. But the advantage is that you don't need to wait for the animation to cycle back to the start. Each successfully received frame in the fountain code allows you to make progress with a high degree of probability. Much faster than waiting for the animation to cycle back round.

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/lib/entities/qr_scanner.dart
+++ b/lib/entities/qr_scanner.dart
@@ -121,10 +121,10 @@ class _BarcodeScannerSimpleState extends State<BarcodeScannerSimple> {
             onDetect: _handleBarcode,
             controller: ctrl,
           ),
-          if (decoder.processedPartsCount() != 0)
+          if (decoder.expectedPartCount() != null)
             Center(
               child: Text(
-                "${decoder.processedPartsCount()}/${decoder.expectedPartCount()}",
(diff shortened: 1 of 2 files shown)


MrCyjaneK (approved):
lgtm

PR #3054, opened 2026-03-11
feat Add gradient-only card design and icon style picker

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_core/lib/balance_card_style_settings.dart
```diff
@@ -39,19 +45,35 @@ class BalanceCardStyleSettings {
       gradientIndex: json["gradientIndex"] as int,
       useSpecialDesign: json["useSpecialDesign"] == 1,
       backgroundImagePath: json["backgroundImagePath"] as String? ?? "",
+      iconStyleIndex: json["iconStyleIndex"] as int? ?? 0,
+      isGradientOnly: json["isGradientOnly"] == 1,
```
malik1004x, line 49: Why are `isGradientOnly` and `iconStyleIndex` separate values? This complicates the database and deserialization. We could just use -1 as gradient-only.

cw_core/lib/balance_card_style_settings.dart
```diff
@@ -39,19 +45,35 @@ class BalanceCardStyleSettings {
...
       cardOrder: json["cardOrder"] as int? ?? -1,
     );
   }
 
   static BalanceCardStyleSettings fromCardDesign(
-      int walletInfoId, int accountIndex, int cardOrder, CardDesign design) {
+      int walletInfoId,
+      int accountIndex,
+      int cardOrder,
+      CardDesign design, {
```
malik1004x: At this point I'd prefer changing all these to named parameters. I should have done that in the first place anyway.

cw_core/lib/card_design.dart
```diff
@@ -337,6 +341,179 @@ class CardDesign {
...
+    final balanceCardIconPrefix = "assets/new-ui/balance_card_icons";
+    final chainIconPrefix = "assets/new-ui/card_icons/chain_icons";
+    final ogIconPrefix = "assets/new-ui/card_icons/og_icons";
+    final outlineIconPrefix = "assets/new-ui/card_icons/outline_icons";
+    final symbolIconPrefix = "assets/new-ui/card_icons/symbol_icons";
+    final genericCakeIcon = "$balanceCardIconPrefix/cake-card-icon.svg";
+    switch (currency) {
+      case CryptoCurrency.arbEth:
+        return [
+          "$symbolIconPrefix/arb-symbol.svg",
```
malik1004x: Sure we can't generate these paths instead of hardcoding them for each currency? Example: `"$symbolIconPrefix/$currency-symbol.svg"`.

cw_core/lib/card_design.dart
```diff
@@ -403,31 +580,86 @@ class CardDesign {
     CardDesign.gradientBlack: CardColorCombination.black,
   };
 
+  static CardDesign _applyIconStyleIndex(
```
malik1004x: You made it static then passed a `CardDesign` object to it, why? I think it would be better to have it as an instance method, similar to `withGradient`.

lib/new-ui/pages/card_customizer.dart
```diff
@@ -46,192 +47,316 @@ class _CardCustomizerState extends State<CardCustomizer> {
...
+      return false;
+    }
+
+    final design = state.availableDesigns[state.selectedDesignIndex];
+    return design.backgroundType == CardDesignBackgroundTypes.svgIcon &&
+        state.availableIconPaths.isNotEmpty;
+  }
+
+  CardDesign _cardStylePreviewDesign(CardCustomizerState state, int index) {
+    if (index >= state.availableDesigns.length) {
```
malik1004x: How could this case ever happen?

lib/new-ui/pages/card_customizer.dart
```diff
@@ -46,192 +47,316 @@ class _CardCustomizerState extends State<CardCustomizer> {
...
+      return state.availableDesigns.isNotEmpty
+          ? state.availableDesigns.first.withGradient(state.selectedColor)
+          : CardDesign.genericDefault;
+    }
+    var design = state.availableDesigns[index].withGradient(state.selectedColor);
+    if (design.backgroundType == CardDesignBackgroundTypes.svgIcon &&
+        state.availableIconPaths.isNotEmpty) {
+      final clamped = state.selectedIconIndex.clamp(0, state.availableIconPaths.length - 1);
```
malik1004x: Again, how could this happen? If there were errors with indexing outside the array, the logic in the Bloc that populates the list is likely wrong. From what I've seen before – the list of icon designs is hardcoded for each currency, so again I'm not sure how it could suddenly shrink.

lib/new-ui/widgets/coins_page/cards/balance_card.dart
```diff
@@ -295,6 +285,29 @@ class BalanceCard extends StatelessWidget {
     );
   }
 
+  Widget cornerSvgIcon(CardDesign design, double iconWidth) {
```
malik1004x: Please don't make functions that return `Widget`. They mess up performance and code readability. Declare a class instead, could be in the same file no problem.

lib/new-ui/widgets/coins_page/cards/balance_card.dart
```diff
@@ -295,6 +285,29 @@ class BalanceCard extends StatelessWidget {
...
+    final path = design.imagePath;
+    final isColoredAsset = path.contains('chain_icons') ||
+        path.contains('og_icons');
```
malik1004x: Potentially very problematic in case we have to change the paths.

lib/new-ui/widgets/coins_page/cards/balance_card.dart
```diff
@@ -295,6 +285,29 @@ class BalanceCard extends StatelessWidget {
...
+      width: iconWidth,
+      colorFilter: isColoredAsset
+          ? null
+          : ColorFilter.mode(
+              design.colors.backgroundImageColor.withValues(alpha: 0.33),
+              BlendMode.dstIn,
+            ),
+    );
+    if (isColoredAsset) {
+      return Opacity(opacity: 0.85, child: image);
```
malik1004x: Instead of detecting the opacity by path, can't you just put the opacity in the SVG? What if we add more icon types?

malik1004x (changes requested):

cw_core/lib/balance_card_style_settings.dart
Blazebrain: We're using separate values because they do different things

We have a plain gradient only card, and we have another card that has icons with different styles. I believe we should keep them separated as they have different concerns, and it doesn't complicate the db

cw_core/lib/balance_card_style_settings.dart
Blazebrain: Yeah, we could have them as named parameters. I'll update

cw_core/lib/card_design.dart
Blazebrain: Yeah, we can do that, but it would require renaming some of the image files and adjusting places they are referenced.

I'll say we do that as a separate refactor, so both the files and every other place its currently used is done.

cw_core/lib/card_design.dart
Blazebrain: Yeah, no reason for it to be static. Moving it to `withIconStyleIndex(int, CryptoCurrency)` next to `withGradient`.

lib/new-ui/pages/card_customizer.dart
Blazebrain: Hmm, thats true, the `selectedDesignIndex` only ever comes from a `ListView` over `availableDesigns`, the guard can't fire. I'll drop it and push an update

lib/new-ui/widgets/coins_page/cards/balance_card.dart
Blazebrain: Yeah, nice catch, was to migrate them into full widget before making the PR ready for review. I'll update it

cw_core/lib/card_design.dart
Blazebrain: In the meantime, I'll simplify it a bit more reducing the manual steps

malik1004x (approved):

PR #3055, opened 2026-03-11
Update icons

tuxpizza:


lib/new-ui/widgets/coins_page/cards/balance_card.dart
```diff
@@ -299,7 +299,7 @@ class BalanceCard extends StatelessWidget {
                   style: TextStyle(color: design.colors.textColor, fontSize: 16),
                 ),
               ),
-              Icon(action.icon, color: design.colors.textColorSecondary),
```
OmarHatem28, line 302: is this correct? cuz now it's not using the action icon and is fixed to arrow forward

tuxpizza: Fixed

OmarHatem28 (approved):

PR #2966, opened 2026-02-25
Settings All Pages (New UI)

tuxpizza:

PR #3065, opened 2026-03-11
New Settings

tuxpizza:
The non-scuffed version of [#2928](https://github.com/cake-tech/cake_wallet/pull/2928) and [#2966](https://github.com/cake-tech/cake_wallet/pull/2966)


lib/new-ui/pages/settings_page.dart
```diff
@@ -66,7 +71,7 @@ class SettingsSectionData {
 
   static SettingsSectionData otherSettings = SettingsSectionData("", "", [
     SettingsListItem("assets/new-ui/settings_row_icons/support.svg", S.current.settings_support, Routes.support),
-    SettingsListItem("assets/new-ui/settings_row_icons/info.svg", S.current.about, Routes.aboutPage),
+    SettingsListItem("assets/new-ui/settings_row_icons/info.svg", "About", Routes.aboutPage),
```
malik1004x: Why are we removing translation here?

lib/new-ui/widgets/receive_page/receive_top_bar.dart
```diff
@@ -85,10 +85,10 @@ class ModalTopBar extends StatelessWidget {
                   duration: const Duration(milliseconds: 200),
                   child: trailingIcon != null
                       ? ModernButton(
-                          key: ValueKey(trailingIcon.hashCode),
-                          size: buttonSize,
-                          onPressed: onTrailingPressed,
-                          icon: trailingIcon!,)
+                    key: ValueKey(trailingIcon.hashCode),
+                    size: buttonSize,
```
malik1004x, line 89: [minor] In the future, please keep the formatting as-is for lines you didn't modify, even if the formatting isn't perfectly aligned with what `dart format`/IntelliJ would output. This makes reviewing easier as it doesn't show those lines as modified.

lib/new-ui/widgets/modal_page_wrapper.dart
```diff
@@ -0,0 +1,87 @@
...
+        resizeToAvoidBottomInset: true,
+        body: Stack(
+          alignment: Alignment.topCenter,
+          children: [
+            Positioned.fill(
+              child: Padding(
+                  padding: EdgeInsets.symmetric(horizontal: horizontalPadding),
+                    child: CustomScrollView(
+                      controller: ModalScrollController.of(context),
+                      physics: ClampingScrollPhysics(),
```
malik1004x, line 42: Are we sure about ClampingScrollPhysics? If we want the app to feel more iOS-like, `BouncingScrollPhysics` would be a better choice.

lib/new-ui/widgets/modal_page_wrapper.dart
```diff
@@ -0,0 +1,87 @@
...
+                  )),
+            if (bottomContent != null)
+              Positioned(
+                bottom: 12,
+                left: horizontalPadding,
+                right: horizontalPadding,
+                child: SafeArea(child: bottomContent!),
+              ),
+            Container(
+              height: (MediaQuery.of(context).padding.top + 84),
```
malik1004x: [minor] Since 84 is an arbitrary number, I'd put it in a variable so maintainers know it's not an estimated size of a particular element.

lib/src/screens/settings/widgets/settings_choices_cell.dart
```diff
@@ -33,51 +37,73 @@ class SettingsChoicesCell extends StatelessWidget {
...
-                        child: Center(
-                          child: Text(
-                            choicesListItem.displayItem.call(e),
-                            style: Theme.of(context).textTheme.bodyMedium?.copyWith(
-                                  color: isSelected
-                                      ? Theme.of(context).colorScheme.onPrimary
-                                      : Theme.of(context).colorScheme.onSurfaceVariant,
-                                  fontWeight: isSelected ? FontWeight.w600 : FontWeight.normal,
-                                ),
+              child: LayoutBuilder(
```
malik1004x, line 51: LayoutBuilder is tasking on the Flutter engine, causing potential slowdowns on rebuilds. Are you positive what you're trying to do here is not achievable without it?

lib/src/screens/settings/widgets/settings_theme_choice.dart
```diff
@@ -135,14 +176,14 @@ class SettingsThemeChoicesCell extends StatelessWidget {
...
+    double baseHeight = (screenHeight * 0.251).clamp(150.0, screenHeight * 0.5);
 
     if (hasAccentColors) {
-      baseHeight += (screenHeight * 0.05).clamp(35.0, 60.0);
+      baseHeight += (screenHeight * 0.6).clamp(35.0, 60.0);
     }
 
     if (theme is BlackTheme) {
-      baseHeight += (screenHeight * 0.08).clamp(48.0, 96.0);
+      baseHeight += (screenHeight * 0.057).clamp(48.0, 96.0);
```
malik1004x: Why would height be different for BlackTheme?

lib/src/screens/settings/connection_sync_page.dart
```diff
@@ -1,119 +1,219 @@
...
+                              keyValue: "disable_automatic_exchange_status_updates",
+                              label: S.current.disable_automatic_exchange_status_updates,
+                              value: _connectionSyncViewModel.disableAutomaticExchangeStatusUpdates,
+                              onChanged: (val) {
+                                _connectionSyncViewModel.setDisableAutomaticExchangeStatusUpdates(val);
+                              }),
+                          if (_connectionSyncViewModel.canUseBlinkProtection)
+                            ListItemToggle(
+                                keyValue: "can_use_blink_protection",
+                                label: S.current.use_blink_protection,
```
malik1004x: Why S.current and not S.of(context)? The latter should be used whenever BuildContext is available as it allows reloading translations without restarting the app.

lib/src/screens/base_page.dart
```diff
@@ -191,24 +193,21 @@ abstract class BasePage extends StatelessWidget {
                   : null,
               // color: Colors.grey[200],
             ),
-            child: Padding(
```
malik1004x, line 194: I trust you've checked this doesn't cause elements to clip off on the still-used old UI pages (e.g. Buy/Sell)?

lib/src/widgets/new_list_row/list_item_selector_widget.dart
```diff
@@ -33,25 +33,25 @@ class ListItemSelectorWidget extends StatelessWidget {
...
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
+            children: [
+              Flexible(child: Text(label, style: textStyle)),
+              Row(
+                children: [
+                  Text(
+                    options[selectedIndex],
+                    style: labelStyle,
+                  ),
+                  SizedBox(width: 8),
```
malik1004x: [minor] you can pass `spacing: 8` to `Row` to avoid adding SizedBoxes.

lib/view_model/settings/connection_sync_view_model.dart
```diff
@@ -0,0 +1,236 @@
...
+
+  @action
+  void setUseBscScan(bool value) {
+    _settingsStore.useBscScan = value;
+    evm!.updateScanProviderUsageState(_wallet, value);
+  }
+
+  @action
+  void setBuiltinTor(bool value, BuildContext context) {
+    if (value) {
```
malik1004x, line 198: Did you copy this `setBuiltinTor` function from somewhere? Asking because if this is new then the logic should be thoroughly tested, but if it was just moved from another VM it should be fine.

tuxpizza: Was moved from another VM

lib/src/screens/settings/widgets/settings_theme_choice.dart
tuxpizza: This is to accommodate the extra settings for the Black theme, but it appears there is an overflow so I'm going to fix it

lib/new-ui/widgets/modal_page_wrapper.dart
tuxpizza: I 100% agree with you here, however BouncingScrollPhysics causes an issue on iOS, where the content gets dragged down separately from the modal sheet itself because the sheet itself is draggable. I think I'll probably switch these modal sheets to being fixed instead of draggable from the entire page, but I will do that in a separate PR, that way the scrolling can be more natural

lib/src/screens/settings/widgets/settings_choices_cell.dart
tuxpizza: To my knowledge yes, I'll have to spend more time later to try and see if I can make it more efficient. 

Since the rebuild only occurs when you use the choices cell (which is only used in a couple places now, mostly display settings) it shouldn't cause a slowdown unless the state of the setting is being changed (to my knowledge)

PR #3071, opened 2026-03-12
Add ability to hide Monero accounts

malik1004x:

# Description

Adds ability to hide Monero accounts from the account customization menu. Main use case is when someone creates a bunch of accounts with a zero balance. Accounts can be restored from a separate screen (you cannot permanently delete them).

# Pull Request - Checklist  

- [x] Format code


OmarHatem28:
this doesn't have a ticket, can you please create one and update the branch and see if it should be in needs testing or ui approval

PR #3090, opened 2026-03-14
Use Silent Payment address for DFX buy flow

TaprootFreak:
## Summary
- Add `addressForPrivateBuy` property to `WalletAddresses` base class (default: `addressForBuy`)
- Override in `BitcoinWalletAddresses` to return Silent Payment (`sp1`) address when available, fallback to P2WPKH
- `signMessage()` uses `b_spend` key when signing for an `sp1` address
- DFX provider uses `addressForPrivateBuy` for Bitcoin buy actions — other providers (MoonPay, Robinhood, Onramper) unaffected
- Hardware wallets: no `silentAddress` available → automatic fallback to P2WPKH

## Context
Companion PR to [DFXswiss/api#3426](https://github.com/DFXswiss/api/pull/3426) which adds `sp1` address recognition and signature verification on the DFX API side. Together they enable privacy-preserving Bitcoin payouts via BIP-352 Silent Payments.

## Changed files
- `cw_core/lib/wallet_addresses.dart` — new `addressForPrivateBuy` getter
- `cw_bitcoin/lib/bitcoin_wallet_addresses.dart` — SP override
- `cw_bitcoin/lib/electrum_wallet.dart` — `signMessage()` SP path
- `lib/buy/dfx/dfx_buy_provider.dart` — use `addressForPrivateBuy` for buy

## Test plan
- [x] Cross-verified SP address derivation: same seed produces same `sp1` address in Cake Wallet and DFX API
- [x] Signature from `b_spend` key verified by DFX API verification logic
- [ ] Manual: buy flow with DFX, verify SP address in auth request
- [ ] CI build passes

### DIFF
--- a/cw_bitcoin/lib/bitcoin_wallet_addresses.dart
+++ b/cw_bitcoin/lib/bitcoin_wallet_addresses.dart
@@ -150,6 +150,12 @@ abstract class BitcoinWalletAddressesBase extends ElectrumWalletAddresses with S
   }
 
   @override
+  String get addressForPrivateBuy {
+    if (silentAddress != null) return silentAddress.toString();
+    return addressForBuy;
+  }
+
+  @override
   String get addressForExchange {
     if (addressPageType == LightningAddressType.p2l) {
       final addresses = receiveAddresses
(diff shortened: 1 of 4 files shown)


OmarHatem28 (commented):
Looks good, thanks for making the necessary changes.

But I would argue about making SP the default, since `silentAddress` will have a value regardless of whether the user enabled SP scanning or not, so technically, all users who have a bitcoin wallet, have their SP address already generated, but they might not be using SP or even know what it is, and having the buy flow use that, would result in users not finding their funds, and reach out to support, and support tell them to scan using SP, and explain what is SP, and so on.

so, I would rather it be a good privacy option for people who knows what they are doing, but not the default.

PR #3098, opened 2026-03-16
fix: Validate node type when loading and selecting nodes to avoid wrong-type nodes

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/store/settings_store.dart
```diff
@@ -1020,7 +1034,13 @@ abstract class SettingsStoreBase with Store {
         final walletTypeForChain = evm!.getWalletTypeByChainId(chainId);
         if (walletTypeForChain != null) {
           final node = nodes[walletTypeForChain];
-          if (node != null) return node;
+          if (node != null) {
+            if (node.type != walletTypeForChain) {
+              throw Exception(
+                  'Node is not valid for $walletTypeForChain (found: ${node.type})');
+            }
```
OmarHatem28: this function is being used in a lot of places, I wouldn't risk and throw an unhandled exception like that

OmarHatem28 (changes requested):
you have added a fallback, but I don't see the main reason for the issue is fixed, weren't you able to find the main cause of the issue?

Blazebrain:
> you have added a fallback, but I don't see the main reason for the issue is fixed, weren't you able to find the main cause of the issue?

 The cause is backup restore. When you restore an old backup onto a newer build, the SharedPreferences node IDs come across but the Hive nodes box gets rebuilt with different key assignments. So currentEthereumNodeIdKey ends up pointing to whatever node sits at that key in the rebuilt box,

OmarHatem28:
closing since the migration is already done, and should be robust enough in #3258

PR #3104, opened 2026-03-17
CW-1193-Add-batch-fetching-of-transactions-to-Electrum

serhii-bor:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_bitcoin/lib/electrum.dart
```diff
@@ -172,8 +172,7 @@ class ElectrumClient {
       unterminatedString += message;
 
       if (isJSONStringCorrect(unterminatedString)) {
-        final response = json.decode(unterminatedString) as Map<String, dynamic>;
-        _handleResponse(response);
```
OmarHatem28, line 176: why was this removed?

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -2508,6 +2541,644 @@ abstract class ElectrumWalletBase
...
+      chunkSize: addressHistoryChunkSize,
+      processChunk: (chunk) => _fetchBatchAddressHistory(chunk, tip, addressHistoryChunkSize),
+    );
+
+    if (addressHistory.isNotEmpty) historiesWithDetails.addAll(addressHistory);
+
+    for (final addressRecord in addressesByType) {
+      final matchedAddresses = addressRecord.isHidden ? hiddenAddresses : receiveAddresses;
+
+      final isUsedAddressUnderGap =
```
OmarHatem28, line 2577: **Enhancement:**
the name is missleading, should be `isUsedAddressAboveGap`

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -2508,6 +2541,644 @@ abstract class ElectrumWalletBase
...
+          type: type,
+        );
+
+        final newLength = walletAddresses.allAddresses.length;
+
+        if (newLength > prevLength) {
+          await fetchTransactionsForAddressTypeBatch(
+              historiesWithDetails,
+              type);
+          return;
```
OmarHatem28, line 2609: so each time new addresses are generated, you loop through the whole list again? why?!!
also, the logic here is very confusing, a for loop inside a for loop and a recursion inside a recursion, this way we will complicate the feature and it will be harder to debug later, we should simplify it.

OmarHatem28 (commented):
I will still continue the review, but thought of putting my comments so far, so you can fix them with the fallback flow

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -2508,6 +2541,644 @@ abstract class ElectrumWalletBase
...
+
+        await walletAddresses.discoverAddressesBatch(
+          matchedAddresses,
+          addressRecord.isHidden,
+              (newAddresses) async {
+            await _fetchBatchAddressHistory(
+              newAddresses,
+              tip,
+              discoveryHistoryChunkSize,
+            );
```
OmarHatem28, line 2582: you already called this earlier in the for loop, why call it again now

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -2508,6 +2541,644 @@ abstract class ElectrumWalletBase
...
+
+          // Litecoin peg-out tagging
+          if (this is LitecoinWallet) {
+            for (final tx2 in transactionHistory.transactions.values) {
+              final heightDiff = ((tx2.height ?? 0) - (tx.height ?? 0)).abs();
+              if (tx2.additionalInfo["isPegOut"] == true &&
+                  tx2.amount == tx.amount &&
+                  heightDiff <= 5) {
+                tx.additionalInfo["fromPegOut"] = true;
+              }
```
OmarHatem28, line 2743: can we later, separate these into a function and just call it instead of duplicating the code.
not just this, there are a few more, but just taking this as an example

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -2508,6 +2541,644 @@ abstract class ElectrumWalletBase
...
+      }
+
+      inputTxIdsByHash[txId] = inputTxIds;
+    }
+
+    return inputTxIdsByHash;
+  }
+
+
+  Future<Map<String, Map<String, dynamic>>> _fetchInputTransactionVerboseBatch(
```
OmarHatem28, line 2973: can easily utilize `_fetchTransactionVerboseBatch` instead of creating 2 duplicate functions for just a couple of lines

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -2719,6 +3401,43 @@ abstract class ElectrumWalletBase
...
+    try {
+      final paramsList = hashes.map((hash) => <Object>[hash]).toList();
+
+      printV('[BATCH_TEST] Start: hashes=${hashes.length}, timeout=${batchTestTimeoutMs}ms');
+
+      await electrumClient.callBatchWithTimeout(
+        method: 'blockchain.scripthash.get_history',
+        paramsList: paramsList,
+        timeout: batchTestTimeoutMs,
+      );
```
OmarHatem28, line 3428: we should handle if the server doesn't throw a timeout exception and just returns an error.
Because in this implementation, it will continue normally

OmarHatem28 (approved):

PR #3106, opened 2026-03-18
feat: migrate Trade data to SQLite with TradeLegacy for backward compatibility

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/exchange/trade_legacy.dart
```diff
@@ -0,0 +1,165 @@
...
+    trade.toRaw = toRaw;
+    trade.stateRaw = stateRaw;
+    await trade.save();
+  }
+
+  static Future<void> migrateAllToSqlite(
+    Box<TradeLegacy> box,
+  ) async {
+    printV('Migrating Trades to SQLite: start');
+    final sw = Stopwatch()..start();
```
OmarHatem28: debugging and forgotten?

lib/exchange/trade.dart
```diff
@@ -55,135 +58,85 @@ class Trade extends HiveObject {
...
 
-  @HiveField(2, defaultValue: -1)
   int fromRaw = -1;
 
   CryptoCurrency? get from => CryptoCurrency.safeDeserialize(raw: fromRaw);
 
-  @HiveField(3, defaultValue: -1)
   int toRaw = -1;
 
   CryptoCurrency? get to => CryptoCurrency.safeDeserialize(raw: toRaw);
```
OmarHatem28: we want to get rid of this, and save the currency itself, not the Raw number.
this was the main issue why we want the migration, is because tokens don't have raw value, so when you try and parse the raw value to currency it will not find it and fail.

so get rid of raw, and safe the currency itself so we don't need to parse anything and just show the saved currency for this trade

lib/exchange/trade.dart
```diff
@@ -55,135 +58,85 @@ class Trade extends HiveObject {
...
     final underscoreIndex = userCurrencyFromRaw!.indexOf('_');
+    if (underscoreIndex < 0) {
+      return CryptoCurrency(
+        title: userCurrencyFromRaw!,
+        tag: null,
+        name: '',
+        raw: -1,
+        decimals: 1,
+      );
+    }
```
OmarHatem28: check with @serhii-bor what was the reason for those getters and variables as well, and remove them as well if it fits

lib/view_model/dashboard/trade_list_item.dart
```diff
@@ -15,15 +16,28 @@ class TradeListItem extends ActionListItem {
...
-  String get tradeFormattedAmount => displayMode == BalanceDisplayMode.hiddenBalance
-      ? "---"
-      : appStore.amountParsingProxy.getDisplayCryptoAmount(trade.amountFormatted(), trade.from!);
+  CryptoCurrency? get _from => trade.fromRaw >= 0 ? trade.from : trade.userCurrencyFrom;
 
-  String get tradeFormattedReceiveAmount => displayMode == BalanceDisplayMode.hiddenBalance
-      ? "---"
-      : appStore.amountParsingProxy
-          .getDisplayCryptoAmount(trade.receiveAmountFormatted(), trade.to!);
+  CryptoCurrency? get _to => trade.toRaw >= 0 ? trade.to : trade.userCurrencyTo;
```
OmarHatem28: things like this would be removed as well, since we would have the currency directly

cw_core/lib/db/sqlite.dart
```diff
@@ -176,10 +179,59 @@ CREATE TABLE BalanceCardStyleSettings (
...
 }
 
+Future<void> _createTradeTable(Database db) async {
+  await db.execute('''
+CREATE TABLE IF NOT EXISTS Trade (
+  tradeId INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+  id TEXT NOT NULL,
+  providerRaw INTEGER NOT NULL DEFAULT 0,
+  fromRaw INTEGER NOT NULL DEFAULT -1,
+  toRaw INTEGER NOT NULL DEFAULT -1,
```
OmarHatem28: those 2 and also `userCurrencyFromRaw` `userCurrencyToRaw` should be removed in this migration and replaced with the currency it self, we can do it as a TEXT with the value being the stringified version of the json map of the currency, but then you need to find a way to distinguish between tokens and currencies, and different tokens (erc, spl, trx, etc...)

OmarHatem28 (changes requested):

lib/exchange/trade.dart
```diff
@@ -49,363 +53,311 @@ class Trade extends HiveObject {
...
 
   ExchangeProviderDescription get provider =>
       ExchangeProviderDescription.deserialize(raw: providerRaw);
 
-  @HiveField(2, defaultValue: -1)
-  int fromRaw = -1;
+  String? fromCurrencyJson;
 
-  CryptoCurrency? get from => CryptoCurrency.safeDeserialize(raw: fromRaw);
+  String? toCurrencyJson;
```
MrCyjaneK:
```
{
  "version": 1,
  "title": "XMR",
  "name": "xmr",
  "decimals": 12,
  "raw": 0,
  "fullName": "Monero",
  "iconPath": "assets/images/crypto/monero.webp",
  "flatIconPath": "assets/new-ui/balance_card_icons/monero.svg"
}
```

Some of the data we store here is very fragile (iconPath / flatIconPath).

lib/exchange/trade.dart
```diff
@@ -49,363 +53,311 @@ class Trade extends HiveObject {
...
+    onChanged.add(null);
+    return rows;
   }
 
-  Map<String, dynamic> toMap() {
+  // ── SQLite serialization ─────────────────────────────
+
+  static bool _hasValidString(String? value) => value != null && value.trim().isNotEmpty;
+
+  void mergeFindTradeByIdResult(Trade updated) {
```
MrCyjaneK, line 167: nit: since sqlite fields accept NULL and things like createdAt / expiredAt are still nullable this function can be heavily simplified to just contain raw `x = updated.x` or `x = updated.x?.trim()` instead of the if conditions.

lib/exchange/trade.dart
```diff
@@ -49,363 +53,311 @@ class Trade extends HiveObject {
...
+      routerData = updated.routerData;
+    }
+    if (_hasValidString(updated.routerValue)) {
+      routerValue = updated.routerValue;
+    }
+    if (_hasValidString(updated.sourceTokenAmountRaw)) {
+      sourceTokenAmountRaw = updated.sourceTokenAmountRaw;
+    }
+
+    internalId = keepInternalId;
```
MrCyjaneK: We are not touching internalId anywhere so `internalId = keepInternalId;` and `final keepInternalId = internalId;` is unnecessary

lib/exchange/trade_currency_snapshot.dart
```diff
@@ -0,0 +1,108 @@
+import 'dart:convert';
+
+import 'package:cw_core/crypto_currency.dart';
+
+class TradeCurrencySnapshot {
```
MrCyjaneK, line 3: I'm really allergic to constant-schema JSON in SQL databases :p

In addition to that a few things:

`flatIconPath` and `chainIconPath` are not guaranteed to be constant as we may move assets in future.

Also I've noticed that in case when the currency is not explicitly supported by us the JSON looks like this:


```
{
  "version": 1,
  "title": "SUN",
  "name": "SUN",
  "decimals": 18,
  "raw": -1,
  "tag": "TRX",
  "fullName": "SUN"
}
```

nit: I'd prefer if we store these fields as raw values in the SQL table (currencyFromName, currencyFromDecimals etc...) to make our future life easier, assuming that in future we will want to lookup all TRX trades we can use SQL instead of manually parsing through every trade JSON object, or doing complex migrations.
Same with updating the fields, `chainIconPath` update, for example during a migration, can be done much easier without querying the entire table and doing json.parse on every row.

lib/new-ui/widgets/coins_page/assets_history/history_trade_tile.dart
```diff
@@ -39,28 +39,30 @@ class HistoryTradeTile extends StatelessWidget {
       width: 50,
       child: Stack(
         children: [
-          CakeImageWidget(imageUrl: _getIconPath(from),
-              width: currencyIconSize, height: currencyIconSize),
+          CakeImageWidget(
+              imageUrl: _getIconPath(from), width: currencyIconSize, height: currencyIconSize),
```
MrCyjaneK, line 43: note: here we may use asset icon that doesn't exist.

lib/view_model/dashboard/trade_list_item.dart
```diff
@@ -15,15 +15,26 @@ class TradeListItem extends ActionListItem {
...
+      return '---';
+    }
+    final to = trade.to;
+    if (to == null) return trade.receiveAmountFormatted();
+    return appStore.amountParsingProxy.getDisplayCryptoAmount(trade.receiveAmountFormatted(), to);
+  }
 
   @override
-  DateTime get date => trade.createdAt!;
+  DateTime get date => trade.createdAt ?? DateTime.now();
```
MrCyjaneK: nit: I'd probably default to `DateTime.fromMillisecondsSinceEpoch(0, isUtc: true);`

lib/new-ui/widgets/coins_page/assets_history/history_trade_tile.dart
```diff
@@ -118,22 +120,22 @@ class HistoryTradeTile extends StatelessWidget {
   }
 
   String _getIconPath(CryptoCurrency currency) {
-    try { // temporarily until we migrate from hive to sqlite and store the full currency object
+    try {
       if (currency.iconPath != null) {
         return currency.iconPath!;
```
MrCyjaneK: here we may use iconPath that is stale from the serialization

cw_core/lib/wallet_info_legacy.dart
```diff
@@ -303,9 +303,13 @@ class WalletInfo extends HiveObject {
...
-    for (final wi in list) {
-      await wi.migrateToSqlite();
-      await wi.delete();
+    for (final trade in list) {
+      try {
+        await trade.migrateToSqlite();
+        await trade.delete();
+      } catch (e) {
+        printV('Error migrating trade ${trade.id}: $e');
+      }
```
OmarHatem28: if this is generic to the box that is passed to it, then make it generic name so it's not confusing
```suggestion
    for (final wi in list) {
      try {
        await wi.migrateToSqlite();
        await wi.delete();
      } catch (e) {
        printV('Error migrating ${box.name} ${wi.id}: $e');
      }
```

lib/exchange/provider/chainflip_exchange_provider.dart
```diff
@@ -296,29 +291,19 @@ class ChainflipExchangeProvider extends ExchangeProvider {
...
       );
 
-      // Find trade and update receiveAmount with the real value received
-      final storedTrade = _getStoredTrade(id);
-      
-      if (storedTrade != null) {
-        storedTrade.$2.receiveAmount = newTrade.receiveAmount;
-        storedTrade.$2.outputTransaction = newTrade.outputTransaction;
-        tradesStore.put(storedTrade.$1, storedTrade.$2);
-      }
```
OmarHatem28, line 320: shouldn't we be updating the database now?

lib/view_model/exchange/exchange_trade_view_model.dart
```diff
@@ -367,7 +350,10 @@ abstract class ExchangeTradeViewModelBase with Store {
 
   static bool _checkIfCanSend(TradesStore tradesStore, WalletBase wallet) {
     final trade = tradesStore.trade!;
-    final tradeFrom = trade.fromRaw >= 0 ? trade.from : trade.userCurrencyFrom;
+    final tradeFrom = trade.from;
+
+    bool _sameCurrency(CryptoCurrency? a, CryptoCurrency? b) =>
+        a != null && b != null && a.titleAndTagEqual(b);
```
OmarHatem28, line 356: we can better override the equality operation `==` in CryptoCurrency

lib/exchange/provider/chainflip_exchange_provider.dart
Blazebrain: We are updating it, but not here. We're using the mergeFindTradeByIdResult function, which is called by the caller after this is done.

The findTradeById functions simply has one responsibility, find the trade and return it to the caller. The caller calls another function to update and persist the data.

cw_core/lib/wallet_info_legacy.dart
Blazebrain: No, its not generic to the box. I've updated it to reflect the box name and id with error.

lib/view_model/exchange/exchange_trade_view_model.dart
Blazebrain: Yeah, we could override `==` on CryptoCurrency, but I wouldn’t recommend we do that yet.

That would make almost every comparison mean “same title + tag” and ignore everything else (raw, name, decimals, etc.). I’m thinking of imported tokens where the label sometimes doesn’t align with the default asset even when it’s actually the same asset. if we override equality globally, we only get that one rule everywhere.

We can have a call on it and possibly standardize it, but for now I’d suggest we keep this kind of check for only places where we need it

cw_core/lib/wallet_info_legacy.dart
OmarHatem28: ok, so this shouldn't have been modified anyway, because this is not a trade object we are migrating, it's the wallet info

OmarHatem28 (approved):

PR #3122, opened 2026-03-21
feat(ios): add iPhone Action Button support for QR scanner

hundehausen:
# Description

Adds support for the iPhone Action Button (iPhone 15 Pro+) to open the QR scanner directly in Cake Wallet. This reduces the steps to scan a payment QR from: unlock → find app → open → tap send → tap scan, down to a single Action Button press.

**⚠️ Note: This PR is untested and should be treated as inspiration/proof-of-concept, not merge-ready code. It has not been manually verified on a real device. Review of logic and integration is welcome before any actual implementation.**

## Changes

- **`ios/Runner/ScanQRIntent.swift`** (new) — `AppIntent` struct that opens the app via `cakewallet://quickaction/scan`
- **`ios/Runner/AppShortcutsProvider.swift`** (new) — Registers the intent as an `AppShortcut`, making it discoverable in Action Button settings, the Shortcuts app, and Spotlight (requires iOS 16.4+)
- **`lib/view_model/link_view_model.dart`** — Adds `scan` case to route and args resolution for the deep link handler
- **`lib/router.dart`** — Passes `autoOpenScanner` arg from deep link into `SendPageParams`
- **`lib/new-ui/pages/send_page.dart`** — Adds `autoOpenScanner` field to `SendPageParams` and `NewSendPage`; triggers a post-frame QR scanner open when set

## Flow

```
Action Button press
  → iOS runs ScanQRIntent.perform()
  → Opens cakewallet://quickaction/scan
  → uni_links delivers URI → auth check (PIN/biometric)
  → LinkViewModel routes to Send page with {autoOpenScanner: true}
  → Post-frame callback opens QR scanner
  → Scanned result populates address/amount fields
```

## Known gaps / TODOs for a real implementation

- The two Swift files need to be added to the Runner Xcode target manually
- `_autoScanQR` currently uses a stub — the scanned URI handling needs to be wired to the same logic as `NewSendAddressInput.onURIScanned`
- Needs testing on a physical device (iOS 16.4+) for Action Button assignment
- Cold-start and background-resume scenarios untested
- Existing quick actions (send/receive from Home Screen) should be regression-tested

# Pull Request - Checklist

- [ ] Format code

### DIFF
--- a/ios/Runner/AppShortcutsProvider.swift
+++ b/ios/Runner/AppShortcutsProvider.swift
@@ -1,0 +1,17 @@
+import AppIntents
+
+@available(iOS 16.4, *)
+struct CakeWalletShortcuts: AppShortcutsProvider {
+    static var appShortcuts: [AppShortcut] {
+        AppShortcut(
+            intent: ScanQRIntent(),
+            phrases: [
+                "Scan QR with \(.applicationName)",
+                "Pay with \(.applicationName)",
+                "Scan payment with \(.applicationName)"
+            ],
+            shortTitle: "Scan QR Code",
+            systemImageName: "qrcode.viewfinder"
+        )
+    }
+}
--- a/ios/Runner/ScanQRIntent.swift
+++ b/ios/Runner/ScanQRIntent.swift
@@ -1,0 +1,14 @@
+import AppIntents
+import UIKit
+
+@available(iOS 16.0, *)
+struct ScanQRIntent: AppIntent {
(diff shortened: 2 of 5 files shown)


MrCyjaneK:
You can already scan any QR code with your camera app and it will open cake if it's compatible. I don't think it is necessary to create a worse copy of existing feature present in OS in the app

If you want the please open an issue with feature request so we can discuss it, as PRs should be used as implementations (and should be tested before opening) and not just discussion.

PR #3134, opened 2026-03-26
Node Labels Migration

tuxpizza:

# Description

Adds labels to the default nodes, and adds a settings migration for existing installations.

Adds `isDefault` bool to each node to more easily check the default node.

Adds `isOfficial` bool to each node to know which one is an official Cake node.

Adds `isIncluded` bool to each node to tag all included app nodes.

# Pull Request - Checklist  

- [x] Format code


lib/entities/default_settings_migration.dart
```diff
@@ -614,6 +610,7 @@ Future<void> defaultSettingsMigration(
           break;
         case 63:
           await _addXaut0TokenToExistingSolanaWallets();
+          await updateAllDefaultNodes(nodes: nodes);
```
malik1004x, line 615: Why is this the same migration step? If you didn't also add the above call (`_addXaut0...`) in this same PR, please add `case 64:` and bump the migration version in `main.dart` (Ctrl+F `63` and you'll find it).

lib/entities/default_settings_migration.dart
```diff
@@ -671,25 +670,41 @@ Future<void> _changeDefaultNode({
     shouldReplace = oldUri?.any((e) => currentNode.uriRaw.contains(e)) ?? true;
   }
 
+  Node? oldDefaultNode = nodes.values.firstWhereOrNull(
```
malik1004x, line 675: Are you modifying logic for an existing migration? This could potentially break something, especially if you're relying on this migration to run for existing installs as well (which it won't if the user keeps their app up-to-date). If you're just doing changes to still have it compile, that's fine, but generally don't modify old migration logic.

lib/entities/default_settings_migration.dart
```diff
@@ -1457,6 +1472,23 @@ Future<void> migrateExistingNodesToUseAutoSwitching(
   }
 }
 
+Future<void> updateAllDefaultNodes(
+    {required Box<Node> nodes}) async {
+
+      final defaultNodes = await loadAllDefaultNodes();
+
+      for (var node in [...nodes.values.toList()]) {
```
malik1004x: [minor] Why the list joining and not just a `.toList()` call?

lib/src/screens/settings/manage_nodes_page.dart
```diff
@@ -100,7 +100,7 @@ class ManageNodesPage extends BasePage {
                       }
                       final nodeListRow = NodeListRow(
                         title: node.uriRaw,
-                        subtitle: node.label ?? '',
+                        subtitle: node.label == null ? null : node.label! + ((node.isDefault ?? false) ? ' (Default)' : ''),
```
malik1004x: [minor] This tenary is a bit confusing, I'd suggest putting this in a variable set with an if-clause rather than cramming it into this object constructor.

lib/entities/default_settings_migration.dart
tuxpizza: Its only for the current (in this version) and future runs to be compatible with the new node variables, previous runs won't matter.

malik1004x (approved):

assets/polygon_node_list.yml
```diff
@@ -1,14 +1,14 @@
...
   uri: polygon-bor-rpc.publicnode.com
   useSSL: true
   isDefault: true
   isEnabledForAutoSwitching: true
+  label: "polygon-bor-rpc.publicnode.com"
 -
-  uri: polygon.llamarpc.com
-  useSSL: true
+  uri: polygon-rpc.com
+  label: "Ankr"
```
OmarHatem28, line 9:
```suggestion
  label: "polyhon RPC"
```

cw_core/lib/node.dart
```diff
@@ -106,6 +119,15 @@ class Node extends HiveObject with Keyable {
...
   String? label;
 
+  @HiveField(13, defaultValue: false)
+  bool? isDefault;
+
+  @HiveField(14, defaultValue: false)
+  bool? isOfficial;
+
+  @HiveField(15, defaultValue: false)
+  bool? isIncluded;
```
OmarHatem28, line 129: whats this for?

lib/entities/default_settings_migration.dart
```diff
@@ -615,6 +611,9 @@ Future<void> defaultSettingsMigration(
         case 63:
           await _addXaut0TokenToExistingSolanaWallets();
           break;
+        case 64:
+          await updateAllDefaultNodes(nodes: nodes);
```
OmarHatem28, line 615: this will need to be 65 after you update the branch

PR #3158, opened 2026-04-08
fix: get height by date for monero

MrCyjaneK:
# Description

# Pull Request - Checklist  

- [ ] Format code


cw_core/lib/get_height_by_date_xmr.dart
```diff
@@ -0,0 +1,423 @@
+DateTime _t(int t) => DateTime.fromMicrosecondsSinceEpoch(t * 1000);
```
OmarHatem28: should be milliseconds not micro.
Plus, I don't think we need this, you are just using it and then converting it back to milliseconds and saving it as an int, so there is no need for the DateTime object to be saved in the memory, just pass the int and multiply it by 1000 in the constructor, and this way we can make the constructor const as well
i.e
```
class _Checkpoint {
  final int timeMs;
  final int height;

  const _Checkpoint(int timeSeconds, this.height) : timeMs = timeSeconds * 1000;
}

OmarHatem28 (approved):

PR #3162, opened 2026-04-09
charts

malik1004x:

# Description

Adds the Charts page to the app. Includes model and bloc for interacting with the prices API as well as cache in the local SQLite database.

# Pull Request - Checklist  

- [ ] Format code


lib/entities/default_settings_migration.dart
```diff
@@ -1545,3 +1550,9 @@ Future<void> _addXaut0TokenToExistingSolanaWallets() async {
     printV('Error in XAUT0 migration: $e');
   }
 }
+
+Future<void> createDefaultChartsData() async {
+    await ChartsAsset(asset: CryptoCurrency.btc, isFavorite: true).insert();
+    await ChartsAsset(asset: CryptoCurrency.xmr, isFavorite: false).insert();
+    await ChartsAsset(asset: CryptoCurrency.eth, isFavorite: false).insert();
```
OmarHatem28: shouldn't we be basing that on users' wallets at least?

lib/main.dart
```diff
@@ -322,7 +322,7 @@ Future<void> initializeAppConfigs({bool loadWallet = true}) async {
     payjoinSessionSource: payjoinSessionSource,
     anonpayInvoiceInfo: anonpayInvoiceInfo,
     havenSeedStore: havenSeedStore,
-    initialMigrationVersion: 65,
+    initialMigrationVersion: 66,
```
OmarHatem28: update with dev please

lib/new-ui/widgets/charts_page/asset_grid_header.dart
```diff
@@ -0,0 +1,81 @@
...
+          ],
+        );
+      },
+    );
+  }
+
+  void _showCurrencyPicker(BuildContext context, ChartsBloc bloc) {
+    showPopUp(
+        context: context,
+        builder: (context) => CurrencyPicker(
```
OmarHatem28: probably this is using the old picker, cuz the new one was recently merged, please check if the new one was updated in the same class or there is a new class for it and use it

cw_core/lib/db/sqlite.dart
```diff
@@ -41,7 +41,7 @@ Future<void> initDb({String? pathOverride}) async {
     }
   }
   await db?.close();
-  db = await openDatabase(dbFile.path, version: 7,
+  db = await openDatabase(dbFile.path, version: 8,
```
OmarHatem28: update with dev please

lib/new-ui/model/charts/util/price_change_data.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:cake_wallet/new-ui/model/charts/util/price_change_direction.dart';
+
+class PriceChangeData implements Comparable<PriceChangeData> {
+  final PriceChangeDirection direction;
+  final String amount;
```
konstantinullrich: Why not use Money here?

lib/new-ui/model/charts/price_data.dart
```diff
@@ -0,0 +1,100 @@
...
+
+  switch (type) {
+    case "fiat":
+      return FiatCurrency.deserialize(raw: id);
+    case "crypto":
+      return CryptoCurrency.fromString(id);
+    case "evm":
+      throw UnimplementedError();
+    case "sol":
+      throw UnimplementedError();
```
konstantinullrich, line 22: So no support for ERC20 and Sol tokens yet?

konstantinullrich (approved):

lib/new-ui/model/charts/util/price_change_data.dart
malik1004x: `Money` didn't exist when I wrote this. I would have used it now, I suppose it'd be good to discuss if we should refactor or merge as-is.

lib/new-ui/model/charts/price_data.dart
malik1004x: There is support, but not for user-added ones. I cannot reliably add it as it depends on Hive and required the `Box` to be passed in. Since user-added tokens are quite niche I say we wait until the tokens are moved to SQLite (the PR for that is in-progress anyway). It should be noted that implementing this would be useless without also adding those tokens to the currency picker list – I have code for that in the Swap Refactor PR, but since it'll be a while till that's merged I'd just paste it in here if deemed necessary.

konstantinullrich:
Also where are the tests?

lib/new-ui/model/charts/util/price_change_data.dart
konstantinullrich: I would be in favor of refactor now, since I assume it would be minor.

lib/new-ui/model/charts/price_data.dart
konstantinullrich: Fair point

lib/new-ui/model/charts/price_data.dart
```diff
@@ -0,0 +1,104 @@
...
+        start?.secondsSinceEpoch ?? 0,
+        // dart doesn't have INT_MAX, apparently.
+        end?.secondsSinceEpoch ?? 0x7FFFFFFFFFFFFFFF,
+      ],
+    );
+    return List.generate(json.length, (index) => PriceData.fromJson(json[index]));
+  }
+
+  Future<void> insert() async {
+    await db!.insert(tableName, toJson(), conflictAlgorithm: ConflictAlgorithm.ignore);
```
OmarHatem28: you sure this and the insertMany should be ConflictAlgorithm.ignore not replace? it should be the same anyway, but just in case we had an issue in the previous data and we fixed it on the server, shouldn't it update here

lib/new-ui/model/charts/price_store.dart
```diff
@@ -0,0 +1,133 @@
...
+    Duration interval,
+  ) =>
+      PriceApiClient.getPrices(
+        PriceRequest(beginTime: start, interval: interval, from: from, to: to),
+      );
+}
+
+class PriceStore {
+  static const priceSources = [
+    // TODO InMemoryPriceSource() - easily implementable w this pattern but idk if we need to optimize this that much
```
OmarHatem28, line 60: I would say we don't, the app already have a lot of things running on memory, let's not add more for fancy things

lib/new-ui/pages/charts_page.dart
```diff
@@ -0,0 +1,103 @@
...
+              ),
+              child: CustomScrollView(
+                physics: const BouncingScrollPhysics(parent: AlwaysScrollableScrollPhysics()),
+                slivers: [
+                  SliverPadding(
+                    padding: EdgeInsets.only(top: MediaQuery.of(context).padding.top),
+                    sliver: CupertinoSliverRefreshControl(
+                      refreshTriggerPullDistance: 160,
+                      refreshIndicatorExtent: 90,
+                      onRefresh: () async => context.read<ChartsBloc>().add(PageRefreshed()),
```
OmarHatem28: why not use the existing variable
```suggestion
                      onRefresh: () async => chartsBloc.add(PageRefreshed()),
```

lib/di.dart
```diff
@@ -1500,14 +1514,12 @@ Future<void> setup({
...
   getIt.registerFactory(() => BuyAmountViewModel());
 
-  getIt.registerFactoryParam<BuySellViewModel, BuySellPageMode, void>((mode, _) => BuySellViewModel(
-      mode: mode, getIt.get<AppStore>(), fiatConversionStore: getIt.get<FiatConversionStore>()));
+  getIt.registerFactoryParam<BuySellViewModel, NewBuySellParams, void>((params, _) => BuySellViewModel(
+      params: params, getIt.get<AppStore>(), fiatConversionStore: getIt.get<FiatConversionStore>()));
 
-  getIt.registerFactoryParam<NewBuySellAmountPage, BuySellPageMode, void>((mode, _) =>
-      NewBuySellAmountPage(buySellViewModel: getIt.get<BuySellViewModel>(param1: mode)));
+  getIt.registerFactoryParam<NewBuySellAmountPage, NewBuySellParams, void>((params, _) =>
```
OmarHatem28, line 1341: I don't see the calls to this instance updated tho, only the new one is correct

lib/new-ui/viewmodels/charts/charts_bloc.dart
```diff
@@ -0,0 +1,173 @@
...
+      if (newCurrencies.isEmpty) {
+        throw Exception(
+          "removed the last currency ${kDebugMode ? "- your ui should block this! what did you do?" : ""}",
+        );
+      }
+
+      final CryptoCurrency newPin;
+      if (s.pinnedCurrency == event.currency) {
+        newPin = newCurrencies.first;
+        await ChartsAsset(asset: s.pinnedCurrency, isFavorite: false).insert();
```
OmarHatem28, line 136: why do we add it then remove it?

lib/new-ui/viewmodels/charts/charts_bloc.dart
OmarHatem28, line 1: clean and nice, good job 🙏

lib/new-ui/widgets/charts_page/chart_modal.dart
OmarHatem28, line 1: reformat

OmarHatem28 (changes requested):

lib/new-ui/model/charts/price_data.dart
malik1004x: From what I remember, `ignore` was chosen for performance, but shouldn't matter much. To be clear, I tested the app while the server was being fixed and it all updated properly, but just in case I can replace it with `.replace`.

lib/new-ui/viewmodels/charts/charts_bloc.dart
malik1004x: It's to prevent a race condition where there are two pinned currencies – this is already guarded by the Bloc, but technically a user could shut down the app while this event handler was running. Chance is very low, but doesn't hurt to have an additional check, and I wanted to safeguard it after the build was rejected by testers because of a broken database.

lib/entities/default_settings_migration.dart
malik1004x: What if the user has no wallets when this runs? What if the user adds or removes wallets? This was how UI did it, they can always add more currencies to the page, it's just the default.

OmarHatem28 (approved):

PR #3188, opened 2026-04-16
add missing mutex in monero

malik1004x:

# Description

Add a missing mutex in getAllSubaddresses which may occasionally cause a segfault when retrieving transaction info.

# Pull Request - Checklist  

- [x] Format code

### DIFF
--- a/cw_monero/lib/api/subaddress_list.dart
+++ b/cw_monero/lib/api/subaddress_list.dart
@@ -65,7 +65,8 @@ int lastWptr = 0;
 int lastTxCount = 0;
 List<TinyTransactionDetails> ttDetails = [];
 
-List<Subaddress> getAllSubaddresses() {
+Future<List<Subaddress>> getAllSubaddresses() async {
+  await txHistoryMutex.acquire();
   txhistory = currentWallet!.history();
   final txCount = txhistory!.count();
   if (lastTxCount != txCount && lastWptr != currentWallet!.ffiAddress()) {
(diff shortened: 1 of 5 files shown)


cw_monero/lib/api/subaddress_list.dart
```diff
@@ -85,6 +86,7 @@ List<Subaddress> getAllSubaddresses() {
     ttDetails.clear();
     ttDetails.addAll(newttDetails);
   }
+  txHistoryMutex.release();
```
MrCyjaneK: `txHistoryMutex.release();` - this whole function should be wrapped in try {} finally so even if something throws we release the mutex to not freeze the wallet

lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -770,8 +770,8 @@ abstract class DashboardViewModelBase with Store {
         "public view key is 0",
       // if (wallet.seed == null) "wallet seed is null",
       // if (wallet.seed == "") "wallet seed is empty",
-      if (monero!.getSubaddressList(wallet).getAll(wallet)[0].address ==
-          "41d7FXjswpK1111111111111111111111111111111111111111111111111111111111111111111111111111112KhNi4")
+      // if (monero!.getSubaddressList(wallet).getAll(wallet)[0].address ==
+      //     "41d7FXjswpK1111111111111111111111111111111111111111111111111111111111111111111111111111112KhNi4")
         "primary address is invalid, you won't be able to receive / spend funds",
```
OmarHatem28: you commented the if, but left this, so it will always have this error, it's probably not showing on the new UI tho
```suggestion
        // "primary address is invalid, you won't be able to receive / spend funds",
```

OmarHatem28 (approved):

PR #3199, opened 2026-04-27
Cw 1438 adapt batch fetching for balance and unspents

serhii-bor:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_bitcoin/lib/electrum.dart
```diff
@@ -340,6 +341,80 @@ class ElectrumClient {
     return historyMap;
   }
 
+  Future<Map<String, List<Map<String, dynamic>>>> getBatchUnspent(
```
OmarHatem28, line 344: Nit: Would rather simplify it

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -3405,7 +3513,23 @@ abstract class ElectrumWalletBase
...
+    // that matches this received transaction, mark it as being from a peg out:
+    for (final tx in transactionHistory.transactions.values) {
+      final heightDiff = ((tx.height ?? 0) - (tx.height ?? 0)).abs();
+    // this isn't a perfect matching algorithm since we don't have the right input/output information from these transaction models (the addresses are in different formats), but this should be more than good enough for now as it's extremely unlikely a user receives the EXACT same amount from 2 different sources and one of them is a peg out and the other isn't WITHIN 5 blocks of each other
+      if (tx.additionalInfo["isPegOut"] == true &&
+          tx.amount == tx.amount &&
+          heightDiff <= 5) {
+        tx.additionalInfo["fromPegOut"] = true;
+      }
+    }
```
OmarHatem28: this will shadow the parameter passed to the function 
```suggestion
    for (final tx2 in transactionHistory.transactions.values) {
      final heightDiff = ((tx2.height ?? 0) - (tx.height ?? 0)).abs();
    // this isn't a perfect matching algorithm since we don't have the right input/output information from these transaction models (the addresses are in different formats), but this should be more than good enough for now as it's extremely unlikely a user receives the EXACT same amount from 2 different sources and one of them is a peg out and the other isn't WITHIN 5 blocks of each other
      if (tx2.additionalInfo["isPegOut"] == true &&
          tx2.amount == tx.amount &&
          heightDiff <= 5) {
        tx.additionalInfo["fromPegOut"] = true;
      }
    }
```

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -2552,65 +2618,108 @@ abstract class ElectrumWalletBase
...
+        return newAddresses
+            .where((addressRecord) => addressRecord.isUsed)
+            .map((addressRecord) => addressRecord.address)
+            .toSet();
+      },
+      type: type,
+    );
+
+    if (newAddresses.isNotEmpty) {
+      currentBranch.addAll(newAddresses);
```
OmarHatem28: you are adding the new addresses here but you're not calling fetch transactions for address type on the new addresses

cw_bitcoin/lib/electrum.dart
```diff
@@ -172,7 +172,8 @@ class ElectrumClient {
       unterminatedString += message;
 
       if (isJSONStringCorrect(unterminatedString)) {
-        final response = json.decode(unterminatedString);
+        final response = json.decode(unterminatedString) as Map<String, dynamic>;
```
OmarHatem28: why specific casting? handleResponse function already handles that

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: The new addresses are already fetched inside the discoverAddressesBatch callback, but I agree this is not clear. I’ll rename the variable and remove currentBranch.addAll() because it is not used after discovery.

cw_bitcoin/lib/electrum.dart
serhii-bor: I restored it because it was removed in a previous batch PR, but you’re right — it’s not needed anymore

OmarHatem28:
will continue in #3252

PR #3214, opened 2026-05-05
fix: Windows file save dialog appearing behind app window

### LINKED ISSUE #3107: Backup export on Windows does not prompt for save location
Bug Description
No follow-up dialog appears on Windows when attempting to export a backup file. After completing the backup steps, there is no prompt to choose a file location or name.

Steps to Reproduce

Go to "Create backup"

Enter your PIN

Click on "Export backup"

Confirm with "Yes, I have"

Wait → No dialog appears to save the file

Expected Behavior
A standard Windows file dialog should open, allowing the user to:

Choose a save location

Specify a file name

Actual Behavior
No dialog or feedback appears. The user is not informed where (or if) the backup file is saved.

Platform

OS: Windows 11 (Build 26200.8037)

Device: PC (x86 architecture, non-ARM)

Cake Wallet Version: Cake_Wallet_v5.9.0_Windows.exe

Additional Context
There is no indication of where the backup file is stored or how to access it afterward. This makes it unclear whether the backup process was successful or where the file can be retrieved for import on another device.

SetupCoding:
## Summary

Fixes #3107  Backup export on Windows does not prompt for save location.

## Root Cause

The Win32 save file dialog was opening **behind** the Flutter window, making it invisible to users. This happened because `FilePicker.platform.saveFile()` defaults `lockParentWindow` to `false`, so when the dialog is spawned in a separate Dart isolate, it has no parent window handle and appears behind the app.

## Changes

- Added `lockParentWindow: true` to all `FilePicker.platform.saveFile()` calls, which sets the Flutter window as the dialog's owner so it appears in front of the application
- Replaced force-unwrap (`outputFile!`) with a null check + early return, so user cancellation is handled gracefully instead of throwing an unhandled exception

### Files changed

- `lib/src/screens/backup/backup_page.dart`  primary fix (backup export)
- `lib/src/screens/settings/mweb_logs_page.dart`  same latent bug
- `lib/src/screens/settings/silent_payments_logs_page.dart`  same latent bug
- `lib/src/screens/seed/wallet_seed_page.dart`  same latent bug (already had null check, just needed `lockParentWindow`)

## Testing

1. Build on Windows  Create Backup  Export Backup  save dialog now appears in front of the app
2. Cancel the dialog  no crash, returns gracefully
3. Complete the save  file is written to chosen location

### DIFF
--- a/lib/src/screens/backup/backup_page.dart
+++ b/lib/src/screens/backup/backup_page.dart
@@ -172,10 +172,15 @@ class BackupPage extends BasePage {
 
   Future<void> _saveFile(BackupExportFile backup) async {
     String? outputFile = await FilePicker.platform
-        .saveFile(dialogTitle: 'Save Your File to desired location', fileName: backup.name);
+        .saveFile(
+            dialogTitle: 'Save Your File to desired location',
+            fileName: backup.name,
+            lockParentWindow: true);
+
+    if (outputFile == null) return;
 
     try {
-      await backup.file.copy(outputFile!);
+      await backup.file.copy(outputFile);
     } catch (exception, stackTrace) {
       await ExceptionHandler.onError(FlutterErrorDetails(
         exception: exception,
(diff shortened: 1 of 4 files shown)


OmarHatem28 (approved):
Good catch, Thank you for contributing 🙏

PR #3224, opened 2026-05-11
Pay anything fixes

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (approved):

OmarHatem28:
I would just make it also detect the contract address belong to which chain, since we already have the default tokens list, so if I scan a QR with USDT on Eth, while on a Polygon wallet, it should detect that instead of falling back to Polygon

PR #3249, opened 2026-05-21
fix: monero address display in tx history

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28:
@Blazebrain why are we returning the account with the address and then regex it, I am confused
CC @malik1004x if this is something intended please elaborate

malik1004x:
This appears correct – it's a fix for the backend returning the address together with the label as one string. You could also do this inside the Monero module, but then it could potentially have some side effects.

OmarHatem28 (approved):

PR #3262, opened 2026-05-27
linux fixes

MrCyjaneK:
fix: use ints instead of bools in SQLite to prevent crashes on linux fix: disable lightning if it's unsupported
ui: start linux app in portrait mode instead of landscape so it looks nicer
fix: bitcoin lightning removal on linux

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (approved):
@malik1004x so it turns out that the SQL bool will be handled differently on other platforms, so my previous comment was kinda correct, so let's standardize the approach that Czarek made here from now on

PR #3289, opened 2026-06-03
tx-confirmation-options-on-dev [cw-1463]

malik1004x:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/view_model/send/send_view_model.dart
```diff
@@ -1078,6 +1099,48 @@ abstract class SendViewModelBase extends WalletChangeListenerViewModel with Stor
         });
       }
 
+      // FIXME(malik) ideally, this should be done wallet-side.
+      // it is required because evm, solana and tron don't actually save the transaction info when you send something.
+      // instead, they rely on the tx to eventually get fetched at sync time, which can take a while
```
OmarHatem28, line 1104: why not add it wallet-side then? when the transactions are fetched from the API anyway it will clear the list and addAll

OmarHatem28 (approved):

lib/view_model/send/send_view_model.dart
malik1004x: Because the commitTransaction function is not available from wallet code but rather passed as a function pointer into the PendingTransaction object. Doing this wallet-side would require significant modification to the transaction committing code, which I decided is too risky for this one button.

malik1004x: We cannot wait for transactions to get fetched from the API here. The point is to have the transaction details be accessible the second the transaction is made, which is not possible when you're waiting for the API call.

PR #3291, opened 2026-06-03
Address resolver refactoring

serhii-bor:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/view_model/send/output.dart
```diff
@@ -397,22 +402,48 @@ abstract class OutputBase with Store {
...
-    extractedAddress = await extractAddressFromParsed(context, parsedAddress);
-    note = parsedAddress.description;
   }
 
-  void loadContact(ContactBase contact) {
-    address = contact.name;
-    parsedAddress = ParsedAddress.fetchContactAddress(address: contact.address, name: contact.name);
-    extractedAddress = parsedAddress.addresses.first;
-    note = parsedAddress.description;
+  Future<void> loadContact((String, String) selectedContact) async {
```
OmarHatem28: I would rather pass the Contact as it was, than passing contact.x and contact.y and then treating them as just Strings without context

lib/view_model/send/output.dart
```diff
@@ -397,22 +402,48 @@ abstract class OutputBase with Store {
...
+    address = selectedContact.$1;
+    parsedAddress = ParsedAddress(
+        parsedAddressByCurrencyMap: {}, addressSource: AddressSource.contact);
+    extractedAddress = selectedContact.$2;
+    note = parsedAddress.description ?? '';
```
OmarHatem28: wouldn't this be always null/empty? like the object was just declared above it, and it had no description

lib/core/address_validator.dart
```diff
@@ -36,7 +36,66 @@ class AddressValidator extends TextValidator {
...
+  static const List<CryptoCurrency> reliableValidateCurrencies = [
+    CryptoCurrency.xmr,
+    CryptoCurrency.btc,
+    CryptoCurrency.ltc,
+    CryptoCurrency.bch,
+    CryptoCurrency.trx,
+    CryptoCurrency.nano,
+    CryptoCurrency.banano,
+    CryptoCurrency.sol,
+    CryptoCurrency.wow,
```
OmarHatem28:
```suggestion
    CryptoCurrency.sol,
```

lib/core/address_validator.dart
```diff
@@ -72,10 +131,12 @@ class AddressValidator extends TextValidator {
...
             '|(${MwebAddress.regex.pattern})(\$|\s)';
       case CryptoCurrency.nano:
-        pattern = '[0-9a-zA-Z_]+';
+        pattern = '(?:nano_|xrb_)[13][13456789abcdefghijkmnopqrstuwxyz]{59}';
       case CryptoCurrency.banano:
-        pattern = '[0-9a-zA-Z_]+';
+        pattern = 'ban_[13][13456789abcdefghijkmnopqrstuwxyz]{59}';
       case CryptoCurrency.usdc:
+      case CryptoCurrency.sol:
+      pattern = r'[1-9A-HJ-NP-Za-km-z]{43,44}';
```
OmarHatem28: this is placed incorrectly, because case of usdc will also use the sol regex, sol should be separate

lib/core/address_validator.dart
```diff
@@ -134,11 +198,13 @@ class AddressValidator extends TextValidator {
       case CryptoCurrency.usdterc20:
       case CryptoCurrency.xlm:
       case CryptoCurrency.trx:
+      pattern = '^(?:T[1-9A-HJ-NP-Za-km-z]{33}|41[0-9A-Fa-f]{40}|0x41[0-9A-Fa-f]{40})';
```
OmarHatem28: if this is specific to trx, then it should be moved out from here, as it will also resolve usdterc20 and xlm and all the above

lib/core/address_validator.dart
```diff
@@ -134,11 +198,13 @@ class AddressValidator extends TextValidator {
...
       case CryptoCurrency.dai:
       case CryptoCurrency.dash:
       case CryptoCurrency.eos:
       case CryptoCurrency.wow:
-        pattern = '[0-9a-zA-Z]+';
+      pattern = r'(?:W(?:o|m|W)[1-9A-HJ-NP-Za-km-z]{94,96}'
+      r'|Wo[1-9A-HJ-NP-Za-km-z]{106,107})';
```
OmarHatem28: same here

lib/core/address_validator.dart
```diff
@@ -172,7 +238,9 @@ class AddressValidator extends TextValidator {
       case CryptoCurrency.btcln:
         pattern = r'(lightning:)?(lnbc|lntb|lnbs|lnbcrt|lnurl|LNBC|LNTB|LNBS|LNBCRT|LNURL)[a-zA-Z0-9]+';
       case CryptoCurrency.zano:
-        pattern = r'([1-9A-HJ-NP-Za-km-z]{90,200})|(@[\w\d.-]+)';
+        pattern = ignoreZanoAlias == true
+            ? r'(?:Z[1-9A-HJ-NP-Za-km-z]{96}|iZ[1-9A-HJ-NP-Za-km-z]{106})'
+            : r'(?:Z[1-9A-HJ-NP-Za-km-z]{96}|iZ[1-9A-HJ-NP-Za-km-z]{106}|@[a-z0-9]{6,32})';
```
OmarHatem28: shouldn't zano alias be already resolved by the address resolver? why is it in here

lib/src/screens/buy/buy_sell_page.dart
```diff
@@ -410,8 +412,23 @@ class BuySellPage extends BasePage {
...
         borderColor: Theme.of(context).colorScheme.outlineVariant,
-        onPushPasteButton: (context) async {},
-        onPushAddressBookButton: (context) async {},
+        onPushPasteButton: (context) async {
+          final domain = cryptoCurrencyKey.currentState!.addressController.text;
+          final parsed =
+          await fetchParsedAddress(context, domain, buySellViewModel.cryptoCurrency);
+          if (parsed.isNotEmpty) {
+            buySellViewModel.cryptoCurrencyAddress = parsed;
+          }
```
OmarHatem28, line 421: NIT: just extract it in a function and use it instead of same code 3 times

lib/core/address_resolver/address_resolver_service.dart
```diff
@@ -0,0 +1,762 @@
...
+
+        final coins = currency == null
+            ? entry.currencies.toList()
+            : (entry.currencies.contains(currency) ? [currency] : const <CryptoCurrency>[]);
+
+        if (coins.isEmpty) continue;
+        tasks.add(entry.run(query, coins, wallet));
+      }
+
+      final results = await Future.wait(tasks);
```
OmarHatem28: What if one of the tasks throws? Do we get zero address resolving then?

Also, you can just put `Future.wait<ParsedAddress>` to avoid a dynamic-typed list.

lib/core/address_resolver/address_resolver_service.dart
OmarHatem28, line 1: Instead of having a lookup table and loose functions for each resolver, consider doing `abstract class Resolver` with virtual function `Future<ParsedAddress> resolve`. This way they don't have to be in one file and the lookup table is 15 lines instead of 150.

We already have a separate class for each resolver API or so, so we should make them implement this abstract class, and only expose the resolve override, and keep the other functions there but private

lib/core/address_resolver/address_resolver_service.dart
```diff
@@ -0,0 +1,762 @@
...
+        applies: (q) => settingsStore.lookupsNostr && isEmailFormat(q),
+        // Nostr handle example: name@domain
+        run: _lookupsNostr,
+      ),
+    ];
+  }
+
+  static String _stripHtmlTags(String value) {
+    return value
+        .replaceAll(RegExp(r'[\u2028\u2029]'), '\n')
```
OmarHatem28: How does stripping HTML tags relate to resolving addresses? Consider doing an extension for the `String` class so we can reuse this whenever needed.

lib/core/address_resolver/address_resolver_service.dart
```diff
@@ -0,0 +1,762 @@
...
+          parsedAddressByCurrencyMap: domainResult.parsedAddressByCurrencyMap,
+          addressSource: AddressSource.twitter,
+          handle: text,
+          profileImageUrl: twitterUser.profileImageUrl,
+          profileName: twitterUser.name,
+        );
+      }
+    }
+
+    String pinnedTweet = twitterUser.pinnedTweet?.text ?? '';
```
OmarHatem28: Once you refactor and add `TwitterResolver extends Resolver` consider splitting the pinned tweet, bio etc into separate functions and call them in `resolve()`.

lib/core/address_resolver/address_resolver_service.dart
```diff
@@ -0,0 +1,762 @@
...
+      final address = await fetchUnstoppableDomainAddress(text, cur.title);
+      if (address.isNotEmpty) {
+        result[cur] = address;
+      }
+    }
+
+    if (result.isNotEmpty) {
+      return ParsedAddress(
+        parsedAddressByCurrencyMap: result,
+        profileImageUrl: 'assets/images/profile.png',
```
OmarHatem28: 1. Don't hardcode a placeholder image, UI should detect there's no image and put a placeholder.
2. Always use SVGs in the new UI.

lib/view_model/send/send_view_model.dart
```diff
@@ -1261,13 +1263,14 @@ abstract class SendViewModelBase extends WalletChangeListenerViewModel with Stor
 
 
   ContactRecord? newContactAddress() {
+    final contacts = contactListViewModel.contacts;
     final Set<String> contactAddresses =
         Set.from(contactListViewModel.contacts.map((contact) => contact.address))
```
OmarHatem28, line 1366: Why didn't you use the variable you just created?

lib/view_model/send/output.dart
```diff
@@ -397,22 +402,48 @@ abstract class OutputBase with Store {
...
   }
 }
 
+Future<bool> showParsedAddressConfirmationAlert(
```
OmarHatem28: Don't pass `BuildContext` to data classes. I know we do already, but don't add more tech debt now so we have less to refactor.

lib/view_model/send/output.dart
OmarHatem28, line 1: Since you are refactoring this, could you also take care of the `address` field value containing the alias if the user resolved the address from that? Currently, to display the address in UI we need to check `isParsedAddress` every time. You could make a getter for this and simplify the UI address display a bit (see `SendConfirmSheet` for example).

OmarHatem28 (changes requested):

lib/view_model/send/output.dart
serhii-bor: fixed

lib/view_model/send/output.dart
serhii-bor: fixed

lib/core/address_validator.dart
serhii-bor: done

lib/core/address_resolver/address_resolver_service.dart
serhii-bor: done

lib/core/address_resolver/address_resolver_service.dart
serhii-bor: done

lib/core/address_resolver/address_resolver_service.dart
serhii-bor: done

lib/view_model/send/send_view_model.dart
serhii-bor: looks like it was a merge issue. Removed.

lib/view_model/send/output.dart
serhii-bor: removed from the data classe

lib/core/address_validator.dart
serhii-bor: reverted

lib/core/address_validator.dart
serhii-bor: reverted

lib/core/address_validator.dart
serhii-bor: revferted

lib/core/address_validator.dart
serhii-bor: reverted

lib/core/address_resolver/address_resolver_service.dart
serhii-bor: fixed

lib/view_model/send/output.dart
serhii-bor: done

lib/core/address_resolver/fio/fio_address_provider.dart
```diff
@@ -0,0 +1,125 @@
...
+      printV('FioAddressProvider.resolve error: $e');
+      return [];
+    }
+  }
+
+  static const apiAuthority = 'fio.blockpane.com';
+  static const availCheck = '/v1/chain/avail_check';
+  static const getAddress = '/v1/chain/get_pub_address';
+
+  static Future<bool> checkAvail(String fioAddress) async {
```
malik1004x, line 64: [minor] Can't these be private?

lib/core/address_resolver/twitter/twitter_address_provider.dart
```diff
@@ -0,0 +1,97 @@
...
+  /// from the user's, profile.
+  /// fallback to resolving Unstoppable Domains if no addresses are found in each of the Twitter user's bio, location, and pinned tweet.
+
+  @override
+  Future<List<ParsedAddress>> resolve({
+    required String query,
+    required List<CryptoCurrency> currencies,
+    required WalletBase wallet,
+  }) async {
+    final formattedName = query.substring(1); // Remove '@' from the beginning of the handle
```
malik1004x:
```suggestion
    final formattedName = query.replaceFirst("@", "");
```

lib/core/address_resolver/unstoppable/unstoppable_address_provider.dart
```diff
@@ -0,0 +1,235 @@
...
+          handle: formattedName,
+        ),
+      ];
+    } catch (e) {
+      printV('[address resolver] Error resolving Unstoppable Domain: $e');
+      return [];
+    }
+  }
+
+  static Future<String> fetchUnstoppableDomainAddress(String domain, String ticker) async {
```
malik1004x, line 210: Can't this be private as well?

lib/core/address_resolver/zano/zano_alias_address_provider.dart
```diff
@@ -0,0 +1,74 @@
...
+  @override
+  Future<List<ParsedAddress>> resolve({
+    required String query,
+    required List<CryptoCurrency> currencies,
+    required WalletBase wallet,
+  }) async {
+    try {
+      if (!currencies.contains(CryptoCurrency.zano)) return [];
+
+      final formattedName = query.substring(1); // Remove '@' from the beginning of the alias
```
malik1004x:
```suggestion
      final formattedName = query.replaceFirst("@", "");
```

lib/core/address_resolver/zano/zano_alias_address_provider.dart
```diff
@@ -0,0 +1,74 @@
...
+      ];
+    } catch (e) {
+      printV('[address resolver] Error resolving Zano Alias: $e');
+      return [];
+    }
+  }
+
+  static Future<String?> fetchZanoAliasAddress(String alias) async {
+    try {
+      final uri = Uri.parse('http://37.27.100.59:10500/json_rpc');
```
malik1004x: [minor] Probably better to put this URL into a `static const`.

lib/core/address_resolver/address_resolver_service.dart
```diff
@@ -0,0 +1,141 @@
...
+            printV('[address resolver service] ...skipping $src (cannot handle)');
+            continue;
+          }
+          if (!provider.supportedCurrencies.contains(currency)) {
+            printV('[address resolver service] ...skipping $src (unsupported currency)');
+            continue;
+          }
+
+          printV('[address resolver service] ...resolving $src - $cur - $query');
+          final results = await provider.resolve(
```
malik1004x: [minor] Noticed you wrapped every single resolve function in a try/catch. Consider wrapping this one call in a try/catch instead so that whoever's adding resolvers in the future doesn't have to remember to also try/catch them.

lib/core/address_resolver/address_resolver_service.dart
```diff
@@ -0,0 +1,141 @@
...
+            query: query,
+            currencies: [currency],
+            wallet: wallet,
+          );
+
+          if (results.isNotEmpty) {
+            final parsedAddress = results.first.parsedAddressByCurrencyMap[currency];
+            final parsedAddressStr = parsedAddress == null || parsedAddress.isEmpty ? 'N/A' : parsedAddress;
+            printV('[address resolver service] resolved $src - $cur - $parsedAddressStr');
+            return [results.first];
```
malik1004x: Why are we returning a list of only the first element?

lib/core/address_resolver/address_resolver_service.dart
```diff
@@ -0,0 +1,141 @@
...
+          }
+        }
+
+        return [];
+      }
+
+      return []; // TODO: Implement multi-currency resolution if no specific currency is provided.
+
+      final tasks = <Future<List<ParsedAddress>>>[];
```
malik1004x, line 121: What is this? It appears unreachable since there is a `return` right before it.

lib/src/screens/buy/buy_sell_page.dart
```diff
@@ -28,9 +27,10 @@ import 'package:keyboard_actions/keyboard_actions.dart';
 import 'package:mobx/mobx.dart';
 
 class BuySellPage extends BasePage {
-  BuySellPage(this.buySellViewModel);
+  BuySellPage(this.buySellViewModel) : _resolver = getIt<AddressResolverService>();
```
malik1004x: Can we wire this in `di.dart` instead? I know this is a legacy page anyway but will make refactoring it easier later.

lib/di.dart
```diff
@@ -1592,10 +1592,9 @@ Future<void> setup({
 
   getIt.registerFactory(() => YatService());
 
-  getIt.registerFactory(() => AddressResolver(
-      yatService: getIt.get<YatService>(),
-      wallet: getIt.get<AppStore>().wallet!,
-      settingsStore: getIt.get<SettingsStore>()));
+  getIt.registerLazySingleton<AddressResolverService>(
```
OmarHatem28: why as a singleton?

lib/core/address_resolver/address_resolver_service.dart
```diff
@@ -0,0 +1,141 @@
...
+      EnsAddressProvider(),
+      ZcashNameAddressProvider(),
+      ZcashMeAddressProvider(),
+      WellKnownAddressProvider(),
+      ZanoAliasAddressProvider(),
+      Bip353AddressProvider(),
+      FioAddressProvider(),
+      ThorchainAddressProvider(),
+      NostrAddressProvider(),
+    ];
```
OmarHatem28: I don't see why this is static, why keep all of those instances live all the time, they can get initiated in the resolve function and destroyed after it's done

lib/view_model/settings/connection_sync_view_model.dart
```diff
@@ -68,6 +88,99 @@ abstract class ConnectionSyncViewModelBase with Store {
   @computed
   bool get builtinTor => _settingsStore.currentBuiltinTor;
 
+  List<AddressSource> get domainLookupSources => AddressResolverService.supportedSources;
+
+  bool lookupValue(AddressSource source) {
```
OmarHatem28, line 93: shouldn't this be annotated with @computed so it rebuilds the observer when it changes?
but then it would probably not work as well because you are returning another computed getter not an observable

OmarHatem28 (commented):
please check those comments and Robert's comments as well

lib/core/address_resolver/address_resolver_service.dart
OmarHatem28: https://github.com/cake-tech/cake_wallet/pull/3291/changes#r3466048000
wrap it in a try and catch and continue to the next provider in the catch

lib/core/address_resolver/unstoppable/unstoppable_address_provider.dart
serhii-bor: no, it’s also used in a utility class

lib/core/address_resolver/fio/fio_address_provider.dart
serhii-bor: done

lib/core/address_resolver/zano/zano_alias_address_provider.dart
serhii-bor: done

lib/core/address_resolver/address_resolver_service.dart
serhii-bor: done

lib/core/address_resolver/address_resolver_service.dart
serhii-bor: Initially, this refactor was part of another feature where we needed to return all addresses associated with the handle. But currently, we only use the first entry.

lib/core/address_resolver/address_resolver_service.dart
serhii-bor: same here. This is for supporting multi-provider and multi-currency resolving, in case we decide to come back to it later.

lib/di.dart
serhii-bor: yes, it makes sense to keep it as a factory since it is stateless now and wallet is passed directly to resolve()

lib/core/address_resolver/address_resolver_service.dart
serhii-bor: I changed it to build providers inside the resolve function.

lib/view_model/settings/connection_sync_view_model.dart
serhii-bor: domainLookupSources is not expected to change dynamically while the app is running, so it does not need to be @computed. Instead, we track each lookup value here: value: _connectionsSyncViewModel.lookupValue(source).

OmarHatem28 (approved):

PR #3311, opened 2026-06-07
fix(trezor): use PAYTOADDRESS script_type for external outputs

anipy1:
The Trezor `signTransaction` deeplink payload currently sets [script_type](https://github.com/cake-tech/cake_wallet/blob/dev/cw_bitcoin/lib/hardware/trezor_service.dart#L75) based on the on-chain output script type via the local `_getScriptType` helper. For **external** outputs (where the `address` field is set, not `addressPath`), this is incorrect. Trezor Suite rejects the deeplink as **"Invalid  parameters from calling app"** because per Trezor's protobuf spec, `script_type` is *"only used if `address_n` is provided."*

For external outputs, Trezor expects `script_type: PAYTOADDRESS` regardless of the recipient's actual on-chain script type, Suite parses the `address` string itself to determine the script type. Values like `PAYTOWITNESS` / `PAYTOP2SHWITNESS` / `PAYTOTAPROOT` specifically mean *"own change paid to that wallet type"* and require `addressPath`, not `address`. Suite reads the contradiction (external `address` + own-change `script_type`) and rejects the deeplink.

This PR fixes `BitcoinTrezorService.signTransaction`: every output goes through as external in the current implementation (the change-output ToDo at line 76 isn't implemented yet), so `script_type` is hardcoded `"PAYTOADDRESS"`. The `_getScriptType` helper stays untouched, still used by LitecoinTrezorService` for its own-change branch.

**Why this didn't surface earlier:** `_getScriptType` returns `"PAYTOADDRESS"` for `P2pkhAddressType.p2pkh` — coincidentally the right value for external outputs. So sending to legacy `1...` addresses works. The bug appears as soon as the destination is segwit (`bc1q...`), nested-segwit (`3...`), or taproot (`bc1p...`).

**Scope note:** `LitecoinTrezorService.signLitecoinTransaction` has the same shape of bug on its external branch (line ~150: `scriptType: _getScriptType(...)` even when `address` is set instead of `addressPath`). I've intentionally **not** touched it in this PR because I haven't run the Litecoin signing flow on real hardware. Same one-line conditional fix applies; happy to follow up if a Litecoin maintainer wants to verify end-to-end, or to extend this PR if you'd rather have it as one change.

**Bug demonstration (before this fix):**

<details>
<summary><strong>📹 Case 1 — P2WPKH (Native SegWit) destination → fails in Trezor Suite</strong></summary>

https://github.com/user-attachments/assets/74c7bdc0-79e2-4957-9b28-c044f53d94cf

</details>

<details>
<summary><strong>📹 Case 2 — P2SH (Nested SegWit / `3…`) destination → fails inside Cake (never reaches Suite)</strong></summary>

https://github.com/user-attachments/assets/5b1d7559-a878-4299-9ed5-b7d870d59357

</details>

<details>
<summary><strong>📹 Case 3 — P2PKH (Legacy `1…`) destination → works (coincidentally)</strong></summary>

https://github.com/user-attachments/assets/1d13cdcc-0ee9-42d2-b6ae-c75304fceb26

</details>

<details>
<summary><strong>📹 Case 4 — P2TR (Taproot `bc1p…`) destination → fails in Trezor Suite</strong></summary>

https://github.com/user-attachments/assets/1859dbd6-61a6-4e8e-8501-c7e3a3e34966

</details>

# Pull Request - Checklist
- [x] Format code
- [ ] Manual tests in accessibility mode (TalkBack on Android) passed — N/A, no UI change in this PR; the fix is in the Trezor deeplink payload
  construction (a non-visible code path).


cw_bitcoin/lib/hardware/trezor_service.dart
```diff
@@ -69,12 +69,22 @@ class BitcoinTrezorService extends HardwareWalletService with BitcoinHardwareWal
...
-          amount: psbt.getOutputAmount(i),
-          address: script.toAddress(),
-          scriptType: _getScriptType(script.getAddressType()!)
-          // ToDo: addressPath: psbt.getOutputBip32Derivation(i, pubkey).$2, // To highlight change outputs
-          ));
+        amount: psbt.getOutputAmount(i),
+        address: script.toAddress(),
+        scriptType: "PAYTOADDRESS",
+        // ToDo: when change-output detection lands, set addressPath +
+        // _getScriptType(...) for own-change outputs.
```
konstantinullrich:
```suggestion
        // ToDo: when change-output detection lands, set addressPath + _getScriptType(...) for own-change outputs.
        // ToDo: addressPath: psbt.getOutputBip32Derivation(i, pubkey).$2, // To highlight change outputs
```

konstantinullrich (approved):

OmarHatem28:
Thank you for noticing and making a quick fix 🙏

PR #3317, opened 2026-06-08
fix: update single coin route to use HardwareWalletViewModel

redsh4de:
# Description

Monero.com had a bug where it would only scan Ledger wallets even on the Trezor selection screen. Thios addressed that

# Pull Request - Checklist  

- [x] Format code


redsh4de:
@Blazebrain This fixes a bug in Monero.com where on the Trezor screen it will only scan for Ledgers.

OmarHatem28 (approved):
Yup, that was an oversight. Thanks for noticing and for contributing

PR #3319, opened 2026-06-09
26-06-09_Update Translation_de_DE

bsn21m:
I also noticed that strings_de.arb still contains a few additional or duplicate keys that no longer appear in strings_en.arb, or that seem to exist there now under slightly different names. This appears to affect keys such as duress_PIN / durres_PIN, ignore / ignor, and waiting_payment_confirmation / awaiting_payment_confirmation.

Before removing any of those older or duplicate keys, I would prefer to confirm it with you first. If they are no longer needed, the German file could also be cleaned up further in that area.


OmarHatem28 (approved):

OmarHatem28:
Thank you 🙏

PR #3322, opened 2026-06-10
feat: harden spam detection for solana and evm

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_solana/lib/solana_wallet.dart
```diff
@@ -152,6 +155,17 @@ abstract class SolanaWalletBase
...
 
+  Future<void> _checkForExistingScamTokens() async {
+    if (!splTokensBox.isOpen) return;
+
+    for (final token in splTokensBox.values) {
+      if (isTokenPropertiesSuspicious(token) && !token.isPotentialScam) {
+        token.isPotentialScam = true;
+        await token.save();
+      }
+    }
```
OmarHatem28, line 184: 1. this would result in a token that we mark as scam, but the user unmarks it, to be always marked again as scam when he opens the app.
2. how would it work for bridged currencies? like `WBTC`? you are using `contains` so it will return true, while it's a valid token (same case for evm)

cw_solana/lib/solana_wallet.dart
```diff
@@ -690,7 +704,90 @@ abstract class SolanaWalletBase
     }
   }
 
+  bool isTokenPropertiesSuspicious(SPLToken token) {
+    final defaultMints =
+        DefaultSPLTokens().initialSPLTokens.map((t) => t.mintAddress).toSet();
+    final isTokenWhitelisted = defaultMints.contains(token.mintAddress);
+
+    final defaultTokenSymbols = DefaultSPLTokens()
+        .initialSPLTokens
```
OmarHatem28: optimize this because it calls a lot of loops inside, and it's called within a loop as well
so extract it outside of this function and outside of the loop in `_checkForExistingScamTokens`

cw_solana/lib/solana_wallet.dart
```diff
@@ -690,7 +704,90 @@ abstract class SolanaWalletBase
...
+      '.org',
+      '.top',
+      '.live',
+      'airdrop',
+      'reward',
+      'distribution',
+      'www',
+      '.xyz',
+      '🎁',
+      'bot',
```
OmarHatem28: same, you are using `contains` while some tokens might have that in their names

cw_solana/lib/solana_wallet.dart
Blazebrain: 1. Added a check to make it a one time scam check for that wallet, so if the user decides to unmark a token, it'll remain unmarked if he opens the app again
2. The blacklist was already using Set<String>.contains(symbol), which is exact match,  Something like `{'BTC','ETH',…}.contains('WBTC')` returns false, so bridged tokens like WBTC don't get flagged.

cw_evm/lib/evm_chain_wallet.dart
```diff
@@ -545,67 +545,112 @@ abstract class EVMChainWalletBase
...
+    'KAS',
+  };
+
+  static bool _hasSuspiciousData(String normalized) {
+    final lower = normalized.toLowerCase();
+    if (_urlLikeSuspiciousMarkers.any(lower.contains)) return true;
+    return _suspiciousWordPattern.hasMatch(lower);
+  }
+
+  bool isTokenPropertiesSuspicious(
```
OmarHatem28, line 593: shouldn't we call this also in AddToken function

Blazebrain: Not really actually, we already have a set of checks we do to catch spams when tokens are added, so it'll flag it

Plus if it somehow passes those flags, we run the scam checks again once the user enters the wallet, so it'll get reclassified.

OmarHatem28 (approved):

PR #3324, opened 2026-06-10
fix: propagate buy/sell crypto address to providers without needing to focus the field

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (approved):
will merge this, but you should have noticed that this would affect BTC that it will get the addressForExchange first and assign it, which would uncover another issue which is that we are still using addressForExchange in the buy flow, but anyway I will push that fix separately on dev

PR #3334, opened 2026-06-13
improve address rotation button ui + fix bug when autogenerate is disabled

malik1004x:

# Description

Fixes the address rotation button when auto-generation is disabled in a Monero wallet. Ensures the latest address is always shown after rotation. Adds a clear loading indicator in case rotation takes too long.

# Pull Request - Checklist  

- [ ] Format code


lib/new-ui/pages/receive_page.dart
```diff
@@ -194,29 +196,30 @@ class _NewReceivePageState extends State<NewReceivePage> {
...
-                if (_largeQrMode) {
-                  ShareUtil.share(
-                    text: widget.addressListViewModel.uri.toString(),
-                    context: context,
-                  );
-                } else if (widget.addressListViewModel.hasAddressRotation) {
-                  widget.addressListViewModel.rotateAddress();
-                }
-              },
+              trailingWidget: Observer(
```
OmarHatem28, line 202: shouldn't we add a condition if it supports rotating or not first? and from the previous ToDo I see there was an issue with mweb rotation, is it fixed or should we hide the button in that case as well

OmarHatem28 (approved):

PR #3352, opened 2026-06-26
Improve how wallets are renamed.

OmarHatem28:
affected wallets (Electrum-like) (BTC, LTC, BCH, Doge)

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_core/lib/pathForWallet.dart
```diff
@@ -36,3 +36,41 @@ Future<String> outdatedAndroidPathForWalletDir({required String name}) async {
...
+  required String toName,
+  required WalletType type,
+}) async {
+  if (fromName == toName) return;
+
+  final typeRoot = await pathForWalletTypeDir(type: type);
+  final sourceDir = Directory('$typeRoot/$fromName');
+  if (!sourceDir.existsSync()) return;
+
+  if (File('$typeRoot/$toName/$toName').existsSync()) {
```
MrCyjaneK: I'd use path/path.dart join method because windows uses \

(also in couple other places too, I'm only leaving a comment under this occurrence)

MrCyjaneK: also I'd just do $toName once.. just in case there's anything other than a wallet file - eg corrupted directory or a wallet that failed creation which could cause some side effects in future

cw_core/lib/pathForWallet.dart
```diff
@@ -36,3 +36,41 @@ Future<String> outdatedAndroidPathForWalletDir({required String name}) async {
...
+Future<void> copyWalletFilesTo({
+  required String fromName,
+  required String toName,
+  required WalletType type,
+}) async {
+  if (fromName == toName) return;
+
+  final typeRoot = await pathForWalletTypeDir(type: type);
+  final sourceDir = Directory('$typeRoot/$fromName');
+  if (!sourceDir.existsSync()) return;
```
MrCyjaneK: probably should throw

cw_bitcoin/lib/litecoin_wallet_service.dart
```diff
@@ -139,26 +137,8 @@ class LitecoinWalletService extends WalletService<
...
-
-    await currentWallet.renameWalletFiles(newName);
-    await saveBackup(newName);
-
-    final newWalletInfo = currentWalletInfo;
-    newWalletInfo.id = WalletBase.idFor(newName, getType());
-    newWalletInfo.name = newName;
-
-    await newWalletInfo.save();
+    await LitecoinWalletBase.renameMwebBox(fromName: currentName, toName: newName);
```
MrCyjaneK: There's no "undo" action in case where mweb box gets renamed but wallet itself doesn't, we probably should do some try catch finally dark magic to undo the rename in case of failure.

Also this may affect one of the checks I suggested for directory existing before copying. Maybe instead of duplicating the logic accept optional param that will take extra filenames?

cw_core/lib/wallet_service.dart
```diff
@@ -25,7 +27,28 @@ abstract class WalletService<N extends WalletCredentials, RFS extends WalletCred
...
+    await saveBackup(newName);
+
+    currentWalletInfo.id = WalletBase.idFor(newName, getType());
+    currentWalletInfo.name = newName;
+    await currentWalletInfo.save();
+
+    final oldDir = Directory('${await pathForWalletTypeDir(type: getType())}/$currentName');
+    if (oldDir.existsSync()) {
+      try {
+        await oldDir.delete(recursive: true);
```
malik1004x, line 49: I don't see a safeguard for when `currentName` and `newName` are the same here. There is one in `copyWalletFilesTo` but it just returns instead of throwing, possibly leading to data loss (should be blocked by UI but I wouldn't rely solely on that).

cw_bitcoin/lib/litecoin_wallet_service.dart
OmarHatem28: correct, will move the delete mweb to after the whole rename is done.

and regarding the second concern, it's not valid because I just rename the box, it's not saved in the wallet directory anyway, hive boxes I believe lives in a different place all of them anyway

PR #3365, opened 2026-07-03
Buy/Sell flow (New UI)

malik1004x:

# Description

Adds the new buy/sell flow consisting of several screens (amount selection, provider selection, payment method selection, confirmation and redirecting).
Also fixes several bugs in the viewmodel found during testing.

# Pull Request - Checklist  

- [ ] Format code


lib/entities/fiat_currency.dart
```diff
@@ -68,67 +71,148 @@ class FiatCurrency extends EnumerableItem<String> with Serializable<String> impl
...
+  static const sgd =
+      FiatCurrency(symbol: "SGD", countryCode: "sgp", fullName: "Singapore Dollar", emoji: "🇸🇬");
+  static const thb = FiatCurrency(
+    symbol: "THB",
+    countryCode: "tha",
+    fullName: "New Thaiwan Dollar",
+    emoji: "🇹🇭",
+  );
+  static const twd =
+      FiatCurrency(symbol: "TWD", countryCode: "twn", fullName: "Thai Baht", emoji: "🇹🇼");
```
konstantinullrich:
```suggestion
      FiatCurrency(symbol: "TWD", countryCode: "twn", fullName: "New Thaiwan Dollar", emoji: "🇹🇼");
```

lib/entities/fiat_currency.dart
```diff
@@ -68,67 +71,148 @@ class FiatCurrency extends EnumerableItem<String> with Serializable<String> impl
...
+  static const sar =
+      FiatCurrency(symbol: "SAR", countryCode: "sau", fullName: "Saudi Riyal", emoji: "🇸🇦");
+  static const sek =
+      FiatCurrency(symbol: "SEK", countryCode: "swe", fullName: "Swedish Krona", emoji: "🇸🇪");
+  static const sgd =
+      FiatCurrency(symbol: "SGD", countryCode: "sgp", fullName: "Singapore Dollar", emoji: "🇸🇬");
+  static const thb = FiatCurrency(
+    symbol: "THB",
+    countryCode: "tha",
+    fullName: "New Thaiwan Dollar",
```
konstantinullrich:
```suggestion
    fullName: "Thai Baht",
```

lib/src/screens/connect_device/connect_device_page.dart
konstantinullrich, line 1: NIT: Unrelated change

lib/solana/cw_solana.dart
konstantinullrich, line 1: NIT: Unrelated change and why this format??

lib/new-ui/pages/buy_sell/buy_sell_confirmation_page.dart
```diff
@@ -0,0 +1,134 @@
...
+            ),
+            Expanded(child: Observer(
+              builder: (_) {
+                return Column(
+                  spacing: 24,
+                  children: [
+                    Column(
+                      spacing: 4,
+                      children: [
+                        Text(
```
konstantinullrich: Please use MoneyText here

lib/new-ui/pages/buy_sell/buy_sell_provider_page.dart
```diff
@@ -0,0 +1,183 @@
...
+        widget.buySellViewModel.changeOption(quote);
+        navigateToConfirmation(context);
+      },
+      leadingIconSize: 32,
+      trailingWidget: Column(
+        crossAxisAlignment: CrossAxisAlignment.end,
+        mainAxisAlignment: MainAxisAlignment.center,
+        spacing: 4,
+        children: [
+            Text(widget.buySellViewModel
```
konstantinullrich: Please use `MoneyText` here

lib/new-ui/widgets/floating_amount_input.dart
```diff
@@ -0,0 +1,97 @@
...
+              ),
+              style: Theme.of(context).textTheme.displayMedium?.copyWith(
+                    fontWeight: FontWeight.w400,
+                    fontSize: 45,
+                    color: Theme.of(context).colorScheme.onSurface,
+                  ),
+            ),
+          ),
+          const SizedBox(width: 8),
+          Text(
```
konstantinullrich: Please use `CurrencySymbolText`

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
```diff
@@ -0,0 +1,475 @@
...
+              ? const CupertinoActivityIndicator()
+              : Column(
+                  mainAxisAlignment: MainAxisAlignment.center,
+                  children: [
+                    Row(
+                      spacing: 4,
+                      mainAxisAlignment: MainAxisAlignment.center,
+                      children: [
+                        if (amount != null)
+                          Text(
```
konstantinullrich, line 429: Please use `MoneyText` here

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
```diff
@@ -0,0 +1,475 @@
...
+                      SizedBox(width: 30, child: Text((curr as FiatCurrency).emoji, style: const TextStyle(fontSize: 24),)),
+                    Column(
+                      crossAxisAlignment: CrossAxisAlignment.start,
+                      mainAxisAlignment: MainAxisAlignment.center,
+                      spacing: 2,
+                      children: [
+                        Text(
+                          curr.fullName ?? curr.symbol,
+                        ),
+                        Text(
```
konstantinullrich: Please use `CurrencySymbolText` here

konstantinullrich (changes requested):

konstantinullrich (approved):

lib/new-ui/widgets/buy_sell/buy_sell_selector_modal.dart
```diff
@@ -0,0 +1,146 @@
+import 'package:cake_wallet/di.dart';
+import 'package:cake_wallet/generated/i18n.dart';
+import "package:cake_wallet/main.dart";
+import 'package:cake_wallet/new-ui/modal_navigator.dart';
+import 'package:cake_wallet/new-ui/pages/buy_sell/buy_sell_amount_page.dart';
+import 'package:cake_wallet/src/widgets/cake_image_widget.dart';
+import 'package:cake_wallet/themes/core/theme_extension.dart';
+import 'package:flutter/material.dart';
+
+enum BuySellPageMode { buy, sell }
```
Blazebrain: We can move `BuySellPageMode` to a separate file, probably in `lib/entities`

Having it in this file would mean we import a widget file in the di, viewmodel etc anytime we want to access the enum

lib/new-ui/widgets/token_chain_display.dart
```diff
@@ -0,0 +1,50 @@
...
+            if ((asset.chainIconPath ?? "").isNotEmpty)
+              Align(
+                alignment: Alignment.bottomRight,
+                child: Container(
+                  decoration: ShapeDecoration(
+                    shape: RoundedSuperellipseBorder(
+                      borderRadius: BorderRadius.circular(5),
+                      side: const BorderSide(color: Colors.black),
+                    ),
+                    color: Colors.white,
```
Blazebrain, line 34: nit: Is this intentional? or should be accessed via the theme

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
```diff
@@ -0,0 +1,478 @@
...
+  final customInputFocusNode = FocusNode();
+
+  @override
+  void initState() {
+    super.initState();
+
+    // this is a hack for the "up to" display to show the actual upper limit.
+    // limits are loaded alongside quotes, and quotes for some providers only load properly with an amount and payment method selected.
+    // it'll be refactored soon anyway, i guess.
+    widget.buySellViewModel.changeFiatAmount(amount: "1000");
```
Blazebrain: Since there is no currency being set with the "1000" here, would this not result in a value being too low to trigger that limit in some currencies?

1000 JPY, VND etc might be lower than some provider minimums

We could derive this from defaultAmounts as it has both amount and currency

lib/view_model/buy/buy_sell_view_model.dart
```diff
@@ -149,6 +154,90 @@ abstract class BuySellViewModelBase extends WalletChangeListenerViewModel with S
...
+
+    return Money.trySafeParse(
+        (fiatConversionStore.prices[cryptoCurrency]! *
+                (double.tryParse(amountForQuote(quote).toString())??0))
+            .toStringAsFixed(2),
+        fiatCurrency);
+  }
+
+  // based on usd values, should have roughly equal worth (was done with ai though so it's subject to correction)
+  static final Map<FiatCurrency, List<String>> _defaultAmountsMap = {
```
Blazebrain, line 200: We sould verify this though :)

lib/view_model/buy/buy_sell_view_model.dart
```diff
@@ -149,6 +154,90 @@ abstract class BuySellViewModelBase extends WalletChangeListenerViewModel with S
...
+    FiatCurrency.tur: ["1500", "3000", "15000", "30000", "75000"],
+    FiatCurrency.twd: ["1500", "3000", "15000", "30000", "75000"],
+    FiatCurrency.usd: ["50", "100", "500", "1000", "2500"],
+    FiatCurrency.vnd: ["1000000", "2000000", "10000000", "20000000", "50000000"],
+    FiatCurrency.zar: ["1000", "2000", "10000", "20000", "50000"],
+    FiatCurrency.kes: ["5000", "10000", "50000", "100000", "250000"],
+  };
+
+  // the fallback is just the usd values.
+  // not great but this fallback shouldn't be triggered anyway
```
Blazebrain, line 256: Some currencies are currently missing compared to the `FiatCurrency.all` list. They'll get the fallback if not included too

ARS, BDT, CNY, GHS, IRR, RUB, SAR, UAH and VEF

lib/view_model/buy/buy_sell_view_model.dart
```diff
@@ -403,17 +498,24 @@ abstract class BuySellViewModelBase extends WalletChangeListenerViewModel with S
...
+            pair.to.symbol == fiatCurrency.symbol &&
+            pair.to.tag == fiatCurrency.tag);
       } else {
-        return provider.supportedFiatList
-            .any((pair) => pair.from == fiatCurrency && pair.to == cryptoCurrency);
+        return provider.supportedFiatList.any((pair) =>
+            pair.from.symbol == fiatCurrency.symbol &&
+            pair.from.tag == fiatCurrency.tag &&
+            pair.to.symbol == cryptoCurrency.symbol &&
+            pair.to.tag == cryptoCurrency.tag);
```
Blazebrain, line 524: Since the same four conditions are written twice, we could extract to a small helper 

Also `pair.to.tag == fiatCurrency.tag` is always `null == null` since FiatCurrency has no tag

lib/new-ui/pages/buy_sell/buy_sell_confirmation_page.dart
```diff
@@ -0,0 +1,134 @@
...
+          end: Alignment.bottomCenter,
+        ),
+      ),
+      child: SafeArea(
+        child: Column(
+          children: [
+            ModalTopBar(
+              title: _pageTitle,
+              leadingIcon: Icon(Icons.arrow_back_ios_new),
+              onLeadingPressed: Navigator.of(context).pop,
```
Blazebrain: lets add the semantic label

lib/new-ui/pages/buy_sell/buy_sell_payment_method_page.dart
```diff
@@ -0,0 +1,73 @@
...
+          begin: Alignment.topCenter,
+          end: Alignment.bottomCenter,
+        ),
+      ),
+      child: SafeArea(
+          child: Column(
+        children: [
+          ModalTopBar(
+            title: S.of(context).payment_method,
+            leadingIcon: Icon(Icons.arrow_back_ios_new),
```
Blazebrain: lets add the semantic label

lib/new-ui/pages/buy_sell/buy_sell_provider_page.dart
```diff
@@ -0,0 +1,181 @@
...
+          begin: Alignment.topCenter,
+          end: Alignment.bottomCenter,
+        ),
+      ),
+      child: SafeArea(
+          child: Column(
+        children: [
+          ModalTopBar(
+            title: _pageTitle,
+            leadingIcon: Icon(Icons.arrow_back_ios_new),
```
Blazebrain: lets add the semantic label

lib/new-ui/pages/buy_sell/buy_sell_provider_page.dart
```diff
@@ -0,0 +1,181 @@
...
+              }
+
+              return SingleChildScrollView(
+                controller: ModalScrollController.of(context),
+                child: Padding(
+                  padding: EdgeInsets.symmetric(horizontal: 18),
+                  child: NewListSections(showHeader: true, sections: {
+                    "": [
+                      ListItemRegularRow(
+                          keyValue: "payment method",
```
Blazebrain: nit: lets use the snake_case.._key convention for this

lib/new-ui/pages/buy_sell/buy_sell_provider_page.dart
```diff
@@ -0,0 +1,181 @@
...
+
+  String get _pageTitle =>
+      (widget.buySellViewModel.mode == BuySellPageMode.buy ? S.current.buy : S.current.sell) +
+      " " +
+      (widget.buySellViewModel.cryptoCurrency.fullName ?? "");
+
+  ListItem quoteListItem(Quote quote) {
+    final fiatAmount = widget.buySellViewModel.fiatAmountForQuote(quote);
+    return ListItemRegularRow(
+      keyValue: quote.provider.title,
```
Blazebrain, line 157: nit: might be fragile using a provider dependent value for the key

lib/new-ui/pages/buy_sell/buy_sell_confirmation_page.dart
```diff
@@ -0,0 +1,134 @@
...
+                                    height: 24,
+                                  ),
+                                  Text(
+                                    buySellViewModel.selectedQuote!.rampName ??
+                                        buySellViewModel.selectedQuote!.provider.title,
+                                    style: TextStyle(
+                                        color: Theme.of(context).colorScheme.onSurfaceVariant,
+                                        fontWeight: FontWeight.w500),
+                                  ),
+                                  SizedBox.shrink()
```
Blazebrain: Is this doing anything or was left over?

lib/di.dart
```diff
@@ -1487,7 +1491,11 @@ Future<void> setup({
...
 
-  getIt.registerFactory(() => BuySellViewModel(getIt.get<AppStore>()));
+  getIt.registerFactoryParam<BuySellViewModel, BuySellPageMode, void>((mode, _) => BuySellViewModel(
+      mode: mode, getIt.get<AppStore>(), fiatConversionStore: getIt.get<FiatConversionStore>()));
+
+  getIt.registerFactoryParam<NewBuySellAmountPage, BuySellPageMode, void>((mode, _) =>
+      NewBuySellAmountPage(buySellViewModel: getIt.get<BuySellViewModel>(param1: mode)));
 
   getIt.registerFactory(
       () => BuySellPage(getIt.get<BuySellViewModel>(), getIt.get<AddressResolverService>()));
```
Blazebrain, line 1502: We're registering`BuySellViewModel` above with the `mode` but this still calls it without and would throw an error

Also, the `BuySellPage` is still reachable via NanoGpt buy deeplink, triggered in `LinkViewModel`. We could pass the mode here or remove the old page completely

lib/src/screens/buy/buy_sell_page.dart
```diff
@@ -473,50 +462,14 @@ class BuySellPage extends BasePage {
...
-          return DesktopExchangeCardsSection(
-            firstExchangeCard: cryptoExchangeCard,
-            secondExchangeCard: fiatExchangeCard,
-            onBuyTap: () =>
-                !buySellViewModel.isBuyAction ? buySellViewModel.changeBuySellAction() : null,
-            onSellTap: () => null,
-            isBuySellOption: true,
-          );
-        }
+        return Placeholder();
```
Blazebrain, line 472: Lets delete the entire page since we're retiring it for the new buy/sell page

If the nanogpt deeplink routes here, it would result in the debug widget showing on prod

its better to remove everything in this PR and route that case to the new page

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
```diff
@@ -0,0 +1,478 @@
...
+    ),
+  );
+
+  void selectCryptoCurrency(BuildContext context) => CurrencyPickerSheet.show(
+      context: context,
+      args: CurrencyPickerArgs(
+        items: widget.buySellViewModel.activeWalletCurrencies.toList(),
+        onSelected: (item) => widget.buySellViewModel.changeCryptoCurrency(currency: item),
+        symbolResolver: widget.buySellViewModel.amountParsingProxy.getCryptoSymbol,
+      ),);
```
Blazebrain: A potential issue here which would result in a buy without a destination address set

The picker displays `wallet.balance.keys` which on any EVM wallet includes all the enabled erc20 tokens. However if we pick a token(say usdt), the `buyViewModel` calls   `_getInitialCryptoCurrencyAddress`, and it returns an empty string since its not native. The empty string then gets passed to the provider url and the flow would go on to checkout with no address set

Seems the old buy page prevented this by validating the form input on the address textfield, so it couldn't pass an empty address when trying to create the quote, we'll need to add a similar check or we add the address text field back

lib/view_model/buy/buy_sell_view_model.dart
```diff
@@ -81,7 +87,7 @@ abstract class BuySellViewModelBase extends WalletChangeListenerViewModel with S
     final formattedFiatAmount = double.tryParse(fiatAmount);
     final formattedCryptoAmount = double.tryParse(_cryptoAmount);
 
-    return isBuyAction
+    return mode == BuySellPageMode.buy
         ? formattedFiatAmount ?? 200.0
         : formattedCryptoAmount ?? (cryptoCurrency == CryptoCurrency.btc ? 0.001 : 1);
   }
```
Blazebrain, line 95: In sell mode, the amount reads `_cryptoAmount`, and the only thing that writes `_cryptoAmount` is `changeFiatAmount`, which sets it to the `"..."` and returns whenever quotes aren't loaded. 

But then, `changeCryptoAmount` isn't called from any of the new pages, so nothing recalculates it afterwards. `double.tryParse("...")` is null, so amount falls back to 0.001 for BTC or 1 for everything else.

This is the actual flow when you enter the sell page and it means that tapping a preset before rates land makes `_cryptoAmount` stays "...".

The result is that every screen shows the crypto equivalent of the fiat amount the user picked, but `launchTrade` sends something else.

By the way, moonpay does `amount.toStringAsFixed(2)` so it sets the 0.001 above to 0.00

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
```diff
@@ -0,0 +1,478 @@
...
+      if(context.mounted) {
+        unawaited(Navigator.of(context).push(CupertinoPageRoute(
+          builder: (context) => Material(color: Colors.transparent, child: page),),));
+      }
+
+    } finally {
+      setState(() {
+        _isLoadingPaymentMethods = false;
+      });
+    }
```
Blazebrain, line 231: `_isLoadingPaymentMethods` is set to true and then blocks on `asyncWhen` with no timeout

That means that if `_getAvailablePaymentTypes` throws then the state never leaves `PaymentMethodLoading` and `asyncWhen` never completes. 

That means the finally would never run, and the continue button would continue spinning forever.

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
```diff
@@ -0,0 +1,478 @@
...
+                          amount?.currency.symbol ?? S.of(context).custom,
+                          style: TextStyle(
+                              fontSize: 18,
+                              fontWeight: FontWeight.w600,
+                              color: amount == null
+                                  ? Theme.of(context).colorScheme.primary
+                                  : Theme.of(context).colorScheme.onSurfaceVariant,),
+                        ),
+                      ],
+                    ),
```
Blazebrain: This would lead to something like `100 USD USD` being displayed here as MoneyText has `showSymbol` as `true` by default and then we're adding another symbole text here

By the way, `MoneyText` also adds a mask when the balances are hidden, not sure we want to have that here as this ones are just presents and not balances

lib/new-ui/pages/buy_sell/buy_sell_confirmation_page.dart
```diff
@@ -0,0 +1,134 @@
...
+                    Column(
+                      spacing: 4,
+                      children: [
+                        Text(
+                          "${buySellViewModel.fiatAmount} ${buySellViewModel.fiatCurrency}",
+                          style: TextStyle(fontSize: 32),
+                        ),
+                        Text(
+                          "≈ ${buySellViewModel.amountForQuote(buySellViewModel.selectedQuote!)?.toStringWithSymbol(fractionalDigits: 8)}",
+                          style: TextStyle(
```
Blazebrain: This would display null directly to the user here if `amountForQuote` returns null

Also, should we not be passing `cryptoCurrency.decimal` instead of hardcoding it to 8?

lib/src/widgets/cake_image_widget.dart
```diff
@@ -130,7 +131,10 @@ class CakeImageWidget extends StatelessWidget {
   /// A caller-supplied [loadingWidget] owns its own semantics; the built-in
   /// spinner is purely visual and must not become an unnamed focus stop.
   Widget _buildLoadingWidget() =>
-      loadingWidget ?? const ExcludeSemantics(child: Center(child: CircularProgressIndicator()));
+      loadingWidget ?? ExcludeSemantics(child:SizedBox(
+          height: height,
+          width: width,
+          child: Center(child: CupertinoActivityIndicator())));
```
Blazebrain, line 138: This changes the loader for every image in the app, and it isn't related to buy/sell

lib/view_model/buy/buy_sell_view_model.dart
```diff
@@ -403,17 +498,24 @@ abstract class BuySellViewModelBase extends WalletChangeListenerViewModel with S
...
+            pair.from.tag == fiatCurrency.tag &&
+            pair.to.symbol == cryptoCurrency.symbol &&
+            pair.to.tag == cryptoCurrency.tag);
       }
     }).toList();
 
     if (validProviders.isEmpty) {
-      buySellQuotState = BuySellQuotFailed();
+      buySellQuotState = BuySellQuotFailed(
+          errorMessage: "Couldn't find a provider that supports ${cryptoCurrency.fullName}.");
```
Blazebrain: Let's localize the error message

lib/view_model/buy/buy_sell_view_model.dart
```diff
@@ -441,11 +543,12 @@ abstract class BuySellViewModelBase extends WalletChangeListenerViewModel with S
         .toList();
 
     if (validQuotes.isEmpty) {
-      buySellQuotState = BuySellQuotFailed();
+      buySellQuotState = BuySellQuotFailed(
+          errorMessage: "No provider could create a quote for ${cryptoCurrency.fullName}");
```
Blazebrain: Let's localize the error message

lib/buy/buy_quote.dart
```diff
@@ -13,11 +14,11 @@ extension RecommendationTitle on ProviderRecommendation {
   String get title {
     switch (this) {
       case ProviderRecommendation.bestRate:
-        return 'BEST RATE';
+        return S.current.best_rate;
       case ProviderRecommendation.lowKyc:
         return 'LOW KYC';
```
Blazebrain: lets localize this

res/values/strings_en.arb
```diff
@@ -200,6 +205,8 @@
   "choose_a_provider": "Choose a provider",
   "choose_account": "Choose account",
   "choose_address": "\n\nPlease choose the address:",
+  "choose_amount_to_buy": "Choose amount to buy",
+  "choose_amount_to_sell": "Choose amount to sell",
```
Blazebrain, line 210: Seems its only localized in english alone for now, how about the others?

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
```diff
@@ -0,0 +1,478 @@
...
+
+  @override
+  State<NewBuySellAmountPage> createState() => _NewBuySellAmountPageState();
+}
+
+class _NewBuySellAmountPageState extends State<NewBuySellAmountPage> {
+  bool _customAmountMode = false;
+  bool _isLoadingPaymentMethods = false;
+  final customInputController = TextEditingController();
+  final customInputFocusNode = FocusNode();
```
Blazebrain, line 43: lets dispose these

cw_core/lib/amount/money.dart
```diff
@@ -25,6 +25,14 @@ class Money implements Comparable<Money> {
     return Money(amount, currency);
   }
 
+  factory Money.safeParse(source, Currency currency, {bool isBaseUnit = false}) {
+    final amount = isBaseUnit
+        ? BigInt.parse(source.toString())
+        : parseFixed(source.toString().withDecimals(currency.decimals), currency.decimals);
+
+    return Money(amount, currency);
+  }
```
Blazebrain: Can't see this being used anywhere

lib/new-ui/pages/buy_sell/buy_sell_redirecting_page.dart
```diff
@@ -0,0 +1,127 @@
+import 'package:cake_wallet/buy/sell_buy_states.dart';
+import 'package:cake_wallet/generated/i18n.dart';
+import 'package:cake_wallet/new-ui/widgets/new_primary_button.dart';
+import 'package:cake_wallet/src/widgets/cake_image_widget.dart';
+import 'package:cake_wallet/view_model/buy/buy_sell_view_model.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_mobx/flutter_mobx.dart';
```
Blazebrain: lets use double quotes as set in the lints

lib/new-ui/pages/buy_sell/buy_sell_provider_page.dart
```diff
@@ -0,0 +1,181 @@
...
+import 'package:cake_wallet/new-ui/pages/buy_sell/buy_sell_confirmation_page.dart';
+import 'package:cake_wallet/new-ui/pages/buy_sell/buy_sell_payment_method_page.dart';
+import 'package:cake_wallet/new-ui/widgets/buy_sell/buy_sell_selector_modal.dart';
+import "package:cake_wallet/new-ui/widgets/money/money_text.dart";
+import 'package:cake_wallet/new-ui/widgets/receive_page/receive_top_bar.dart';
+import 'package:cake_wallet/src/widgets/new_list_row/new_list_section.dart';
+import 'package:cake_wallet/view_model/buy/buy_sell_view_model.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_mobx/flutter_mobx.dart';
```
Blazebrain: lets use double quotes as set in the lints

lib/new-ui/pages/buy_sell/buy_sell_payment_method_page.dart
```diff
@@ -0,0 +1,73 @@
+import 'package:cake_wallet/buy/payment_method.dart';
+import 'package:cake_wallet/entities/new_ui_entities/list_item/list_item_regular_row.dart';
+import 'package:cake_wallet/generated/i18n.dart';
+import 'package:cake_wallet/new-ui/widgets/receive_page/receive_top_bar.dart';
+import 'package:cake_wallet/src/widgets/new_list_row/new_list_section.dart';
+import 'package:cake_wallet/view_model/buy/buy_sell_view_model.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_mobx/flutter_mobx.dart';
```
Blazebrain: lets use double quotes as set in the lints

lib/new-ui/pages/buy_sell/buy_sell_confirmation_page.dart
```diff
@@ -0,0 +1,134 @@
...
+import 'package:cake_wallet/new-ui/pages/buy_sell/buy_sell_redirecting_page.dart';
+import 'package:cake_wallet/new-ui/widgets/buy_sell/buy_sell_selector_modal.dart';
+import 'package:cake_wallet/new-ui/widgets/new_primary_button.dart';
+import 'package:cake_wallet/new-ui/widgets/receive_page/receive_top_bar.dart';
+import 'package:cake_wallet/src/widgets/cake_image_widget.dart';
+import 'package:cake_wallet/src/widgets/new_list_row/new_list_section.dart';
+import 'package:cake_wallet/view_model/buy/buy_sell_view_model.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_mobx/flutter_mobx.dart';
```
Blazebrain: lets use double quotes as set in the lints

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
```diff
@@ -0,0 +1,478 @@
...
+                textColor: Theme.of(context).colorScheme.onPrimary,),
+          ),
+        ],
+      ),
+    );
+}
+
+class BuySellDefaultAmountSelector extends StatelessWidget {
+  const BuySellDefaultAmountSelector(
+      {required this.onFiatSelectorPressed, required this.defaultAmounts, required this.fiatCurrency, required this.mode, required this.onSelected, required this.isLoading, required this.hasCurrencySelector, required this.onCurrencySelectorPressed, required this.cryptoCurrency, super.key,
```
Blazebrain: lets format this

lib/new-ui/widgets/floating_amount_input.dart
```diff
@@ -0,0 +1,98 @@
...
+  final Function(String)? onChanged;
+  final FormFieldValidator<String>? validator;
+
+  @override
+  State<FloatingAmountInput> createState() => _FloatingAmountInputState();
+}
+
+class _FloatingAmountInputState extends State<FloatingAmountInput> {
+  bool _amountFocused = false;
+  late FocusNode focusNode = widget.focusNode ?? FocusNode();
```
Blazebrain, line 30: lets dispose this

Blazebrain (changes requested):
in addition to this, lets apply formatting to the new files as some seemed to not be properly formatted, along with a couple lint issues

lib/new-ui/widgets/buy_sell/buy_sell_selector_modal.dart
malik1004x: And how is that a problem? Dart's `import` is compile-time and does not cause any overhead. Having a separate file for a single-line enum is not great.

lib/new-ui/widgets/token_chain_display.dart
malik1004x: What should be accessed by the theme? The colors Black and White? No, those were taken directly from the Figma a while back – they should be hardcoded.

lib/view_model/buy/buy_sell_view_model.dart
malik1004x: What method are you proposing to verify this? It doesn't matter much anyway – if a user wants a different amount they can use the custom amount option.

lib/view_model/buy/buy_sell_view_model.dart
malik1004x: "Could" is the right word – there's no point of "small helpers" like this.

lib/new-ui/pages/buy_sell/buy_sell_provider_page.dart
malik1004x: the `keyValue` is not used anywhere and the only reason I add it is because the widget requires it.

lib/new-ui/pages/buy_sell/buy_sell_confirmation_page.dart
malik1004x: The `SizedBox.shrink()`? This is so the column has some padding at the bottom, it's a common pattern to use this in conjunction with `spacing`.

lib/new-ui/pages/buy_sell/buy_sell_confirmation_page.dart
malik1004x: Absolutely not. If the currency has 18 decimals, this is a guaranteed overflow.

lib/src/widgets/cake_image_widget.dart
malik1004x: Yes, it is – this is shown when the provider images are loading. I noticed this looked ugly and changed it.

res/values/strings_en.arb
malik1004x: Google Translate localization is terrible, I'd rather not have it translated.

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
malik1004x: if `_getAvailablePaymentTypes` throws, we get an error popup anyway.

cw_core/lib/amount/money.dart
malik1004x: This is kept for the sake of completeness since it's the correct behavior when dealing with amounts where we have no control over the decimals (e.g. the ones received from trade providers). Felt weird to have `trySafeParse` but not `safeParse`. It's used in the Swap refactor anyway.

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
malik1004x: Good catch. ` _getInitialCryptoCurrencyAddress` was overcomplicated and didn't fit the new flow anyway, so I just stubbed it as `.addressForBuy`.

lib/di.dart
malik1004x: What do you mean by "calls it without"? `mode` is passed the whole way.

lib/new-ui/pages/buy_sell/buy_sell_confirmation_page.dart
malik1004x: As for the other point – generally you're right, but `amountForQuote` loads near-instantly after this page is opened so the UI difference is miniscule. I added a condition just in case, though.

lib/new-ui/widgets/buy_sell/buy_sell_selector_modal.dart
Blazebrain: Not overhead, layering. `buy_sell_selector_modal.dart` imports `di.dart` and `main.dart`, and then `di.dart` imports the modal, so basically those two import each other. The view model then picks up main.dart just to read an enum.

Also, `lib/entities` is full of small enum files already, this would sit fine next to them.

Blazebrain: Not overhead, layering. `buy_sell_selector_modal.dart` imports `di.dart` and `main.dart`, and then `di.dart` imports the modal, so basically those two import each other. The view model then picks up main.dart just to read an enum.

Also, `lib/entities` is full of small enum files already, this would sit fine next to them.

lib/new-ui/widgets/token_chain_display.dart
Blazebrain: We already draw this exact badge in `list_item_regular_row_widget.dart` and it comes off the theme there, the border `surfaceContainerHigh`, the fill `onSurface`, and the icon tint `surface`

Each of those tokens change when the theme changes, and if its static, it'll result in some of those wierd light theme images Jack shares to the channel because we didn't consider light theme when adding the colors

Plus, if the Figma changed, then we should update the list row too then, but they shouldn't disagree

lib/view_model/buy/buy_sell_view_model.dart
Blazebrain: That's fair, makes sense

By the way, we should probably remove the "was done with ai though so it's subject to correction" comment : )

lib/view_model/buy/buy_sell_view_model.dart
Blazebrain: Fair enough, they're pretty small

Only the null to null check left for this then

lib/new-ui/pages/buy_sell/buy_sell_provider_page.dart
Blazebrain: Right, its not used in the row widget itself

Though it seems the `NewListSections` uses `item.keyValue` in a couple of places though, plus the integration tests setup which would be merged in soon uses the keys

Overall, we could leave as is, still fine

lib/di.dart
Blazebrain: I'm referring to the line 1501, which registers the `BuySellPage`

```
getIt.registerFactory(
    () => BuySellPage(getIt.get<BuySellViewModel>(), getIt.get<AddressResolverService>()));
```

No param is passed to it, so its null as BuySellPageMode and throws an error

You can test it out using the nanoGpt deeplink. Trigger `nano-gpt://buy`, it routes to `Routes.buySellPage` through `LinkViewModel` and there's a full blown error on that page

We could also just delete the old page like I mentioned in the other comment, and everything becomes fixed

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
Blazebrain: I don't think we should leave that to the popup, we should handle it

Mostly because after the user dismisses the popup, the continue button and any other tap is blocked till they close the sheet, because the finally block never ran and the guard at the beginning of the function would keep returning, so the sheet remains "stuck"

Since `asyncWhen` takes a timeout, we can add it and also handle the error if `_getAvailablePaymentTypes` throws

lib/src/widgets/cake_image_widget.dart
Blazebrain: It's not only the provider images that use this, it cuts across a lot of other screens

The widget already has a `loadingWidget` param, you can pass the loader you want there and you'll get the look you want, but it shouldn't be here and touch everywhere

res/values/strings_en.arb
Blazebrain: Yeah, I agree its terrible, but that's what we've been using so far while a replacement is being worked on

Leaving this in English-only would mean all other language speakers see a language they possibly don't understand in one of the key features of the app

cw_core/lib/amount/money.dart
Blazebrain: True, but It's still not used currently, plus `unreachable_from_main` is in the lint set you added :)

If it'll be needed in swap refactor, then we should add it in that PR and not here

lib/new-ui/widgets/token_chain_display.dart
malik1004x: Cool, make a PR with that.

lib/new-ui/widgets/buy_sell/buy_sell_selector_modal.dart
malik1004x: Imports in Dart are not transitive.

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
malik1004x: Okay, I fixed it the elegant way: if `_getAvailablePaymentTypes` throws it just sets the failure state immediately.

lib/src/widgets/cake_image_widget.dart
malik1004x: Why would I keep other flows using the old, ugly widget if I can provide them the nicer one with zero effort? `CircularProgressIndicator` is not used by us anywhere in New UI since `CupertinoActivityIndicator` fits much better, and the `height` and `width` are fixes for an overflow bug that, yes, could occur on other flows as well.

Blazebrain (approved):

lib/view_model/buy/buy_sell_view_model.dart
```diff
@@ -200,12 +302,18 @@ abstract class BuySellViewModelBase extends WalletChangeListenerViewModel with S
...
+  @computed
+  bool get hasMultipleCurrencies => activeWalletCurrencies.length > 1;
+
   @action
   void changeCryptoCurrencyAddress(String address) => cryptoCurrencyAddress = address;
 
   @action
   Future<void> changeFiatAmount({required String amount}) async {
-    fiatAmount = amount;
+    fiatAmount = amount.replaceAll(",", ".");
```
OmarHatem28, line 327: also not a blocker, but let's try to standardize the use of `.sanitized()`

PR #3369, opened 2026-07-05
fix: make receive actions usable on compact screens

Arbeitszeit:
Issue Number (if Applicable): N/A

# Description

This change keeps the receive-page actions usable on compact Android screens and avoids displaying unavailable address metadata as literal `null` values.

## Changes

- Use equal-width flexible action cells in the receive-page bottom row.
- Reduce button size, padding, and spacing below 400 logical pixels.
- Allow `ModernButton` labels to use up to two centered lines with ellipsis.
- Render transaction count and balance independently only when each value is available.
- Add widget coverage for Spanish labels at 320 px and 360 px widths.
- Add widget coverage for address rows whose metadata is unavailable.

## Validation

- `git diff --check` passes.
- Widget tests were added, but could not be executed locally because the generated root pubspec and Flutter toolchain are not available in the analysis environment. CI validation is requested.

# Pull Request - Checklist

- [ ] Format code

### DIFF
--- a/lib/new-ui/pages/addresses_page.dart
+++ b/lib/new-ui/pages/addresses_page.dart
@@ -421,22 +421,32 @@ class AddressRow extends StatelessWidget {
                             ),
                           ],
                         ),
-                        Row(
-                          mainAxisSize: MainAxisSize.max,
-                          mainAxisAlignment: MainAxisAlignment.spaceBetween,
-                          children: [
-                            Text(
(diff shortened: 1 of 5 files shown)


kotelgg:
We’ll see if it fixes the display issue on my phone too (Unihertz Titan, 1432×1436 pixels). I have no way to use landscape mode. So I can’t enter the PIN because I can’t see all the numbers, and I can’t send a transaction because I can’t see the Send button.

PR #3381, opened 2026-07-08
feat: pin all dependencies to git sources

MrCyjaneK:

# Description

# Pull Request - Checklist  

- [ ] Format code


pubspec.lock
OmarHatem28, line 1: shouldn't this be deleted as well since it's in gitIgnore?

konstantinullrich (approved):

OmarHatem28 (changes requested):
universal_ble, ledger_flutter_plus, ledger_usb_plus are using the wrong commit hashes.

also the way of finding the correct commit hash has 2 potential issues
1. the way it looks for a tag and fallsback to finding the latest commit that has a version, is just a best effort, so I would rather stop the script here and show an error so the dev go and get the right commit hash himself
2. wouldn't ignored files in a package repo which are generated on publish to pub dev, cause an issue when we get them directly from git? things like translations for example and any other thing that gets generated on publish to pubdev, but I guess it's easily discoverable by whoever would be working on that, so I guess we can skip this

pubspec.lock
MrCyjaneK: No, the root one stays

OmarHatem28 (approved):

PR #3383, opened 2026-07-09
CW-1523-AnyPay Redesign

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/new-ui/widgets/anypay/evm_address_detected_sheet.dart
```diff
@@ -0,0 +1,128 @@
...
+                  border: Border.all(color: colors.surface, width: 2),
+                ),
+                child: CakeImageWidget(imageUrl: iconPaths[i], fit: BoxFit.contain),
+              ),
+            ),
+        ],
+      ),
+    );
+  }
+}
```
konstantinullrich: Isn't this a duplicate of eg `_IconWithBadge` in `currency_picker_row.dart`? Why not make it into a separate widget that can be reused?

Since we have similar implementations scattered across the UI.

lib/new-ui/widgets/anypay/swap_from_network_page.dart
```diff
@@ -0,0 +1,137 @@
...
+                children: [
+                  NewPrimaryButton(
+                    onPressed: onProceed,
+                    image: primaryHasSwapIcon
+                        ? SvgPicture(
+                            AssetBytesLoader('assets/new-ui/swap_arrows.svg.vec'),
+                            width: 24,
+                            height: 24,
+                            colorFilter: ColorFilter.mode(colors.onPrimary, BlendMode.srcIn),
+                          )
```
konstantinullrich: Why not use `CakeImageWidget`?

lib/new-ui/widgets/anypay/send_to_network_page.dart
```diff
@@ -0,0 +1,119 @@
...
+                mainAxisSize: MainAxisSize.min,
+                children: [
+                  NewPrimaryButton(
+                    onPressed: onSwitchWallet,
+                    image: SvgPicture(
+                      AssetBytesLoader('assets/new-ui/wallet_filled.svg.vec'),
+                      width: 24,
+                      height: 24,
+                      colorFilter: ColorFilter.mode(colors.onPrimary, BlendMode.srcIn),
+                    ),
```
konstantinullrich: Why not use `CakeImageWidget` here?

lib/new-ui/widgets/anypay/send_to_network_page.dart
```diff
@@ -0,0 +1,119 @@
...
+                  ),
+                  const SizedBox(height: 12),
+                  NewPrimaryButton(
+                    onPressed: onSwap,
+                    image: SvgPicture(
+                      AssetBytesLoader('assets/new-ui/swap_arrows.svg.vec'),
+                      width: 24,
+                      height: 24,
+                      colorFilter: ColorFilter.mode(colors.primary, BlendMode.srcIn),
+                    ),
```
konstantinullrich: Why not use CakeImageWidget here?

lib/new-ui/widgets/currency_picker/single_network_currency_picker.dart
```diff
@@ -41,14 +41,27 @@ class _SingleNetworkCurrencyPickerState extends State<SingleNetworkCurrencyPicke
...
   Widget build(BuildContext context) {
     final native = walletTypeToCryptoCurrency(_network);
     final query = _searchController.text.trim();
-    final tokens =
-        _args.items.where((c) => c != native).where((c) => currencyMatchesQuery(c, query)).toList();
+    final tokens = _args.items
+        .where((c) => c != native)
+        .where((c) => cryptoCurrencyOrTokenToWalletType(c) == _network)
+        .where((c) => currencyMatchesQuery(c, query))
+        .toList();
```
konstantinullrich: Wouldn't one where statement be more efficient since it would only need to loop though the list once?

lib/new-ui/widgets/anypay/send_to_network_page.dart
```diff
@@ -0,0 +1,119 @@
...
+  final String destinationNetworkName;
+  final String destinationIconPath;
+  final String currentNetworkName;
+  final VoidCallback onSwitchWallet;
+  final VoidCallback onSwap;
+
+  @override
+  Widget build(BuildContext context) {
+    final colors = Theme.of(context).colorScheme;
+    return Scaffold(
```
konstantinullrich: Why not use `Material` instead of `Scaffold` since `Scaffold` would be a bit overkill here if you just use the body.

lib/new-ui/pages/send_page.dart
```diff
@@ -224,42 +231,47 @@ class _NewSendPageState extends State<NewSendPage> {
...
+                uri = paymentRequest.address;
+              } else {
+                final chainSuffix = (scheme == 'ethereum' &&
+                        paymentRequest.chainId != null &&
+                        paymentRequest.chainId != 1)
+                    ? "@${paymentRequest.chainId}"
+                    : "";
+                final amount =
+                    paymentRequest.amount.isNotEmpty ? "?amount=${paymentRequest.amount}" : "";
+                uri = "${paymentRequest.scheme}:${paymentRequest.address}$chainSuffix$amount";
```
konstantinullrich: This would still cause the issue reported by Jack, where 20 USDC would turn into 20 ETH

lib/new-ui/widgets/anypay/evm_address_detected_sheet.dart
Blazebrain: No, it's not a duplicate of `_IconWithBadge`. This widget handles stacking multiple evm network images side by side, in an overlapping format. `IconWithBadge` shows for a single currency with its network image stacked.

lib/new-ui/widgets/anypay/swap_from_network_page.dart
Blazebrain: `NewPrimaryButton` accepts `SvgPicture` as its param, and its passed similarly in other calls, that's why I used `SvgPicture` here. I'll expand it to a generic `Widget` so either `SvgPicture` or `CakeImageWidget` can be passed to it

lib/new-ui/widgets/anypay/send_to_network_page.dart
Blazebrain: Same reason as above

lib/new-ui/widgets/anypay/send_to_network_page.dart
Blazebrain: Same reason as above

lib/new-ui/widgets/currency_picker/single_network_currency_picker.dart
Blazebrain: Yeah, I'll collapse it into a single where statement

lib/new-ui/widgets/anypay/send_to_network_page.dart
Blazebrain: Right, I'll switch it to `Material`

lib/new-ui/widgets/anypay/evm_address_detected_sheet.dart
konstantinullrich: I see, that was my mistake

lib/new-ui/pages/send_page.dart
Blazebrain: Right, adjusted the uri creation to pull from the erc681 uri directly

Blazebrain: By the way, this won't affect the flow currently (that's why Jack doesn't report it again in the anypay builds), because it has been fixed via the changes. this is because I added the `contractAddress` to the payment request model directly, and that would always fetch the right token, giving 20 USDC if its the one. 

Adjusted this to handle here too just in case(another confident maybe edgecase?), but it was already fixed in the actual revamp flows

konstantinullrich: I was wondering why we need the URI at this point anyways, because all relevant Infos are in the payment request

konstantinullrich (approved):

lib/new-ui/pages/swap_page.dart
```diff
@@ -693,140 +908,271 @@ class _NewSwapPageState extends State<NewSwapPage> {
...
+                  ],
+                ),
+              ),
+            ),
+          ],
+        );
+      },
+    );
   }
 }
```
konstantinullrich: Why not put that in a different file?

lib/new-ui/pages/send_page.dart
```diff
@@ -1039,274 +1154,322 @@ class _NewSendPageState extends State<NewSendPage> {
...
+      scheme: "",
+      isValid: true,
+    );
+    widget.paymentViewModel.applyManualEvmSelection(detection);
+
+    final currentType = widget.sendViewModel.wallet.type;
+    final currentChainId =
+        isEVMCompatibleChain(currentType) ? evm!.getChainIdByWalletType(currentType) : null;
+    if (currentChainId != null && target.chainId == currentChainId) {
+      await _applyPaymentSelectingCurrency(paymentRequest, target.currency);
```
konstantinullrich, line 1528: Why do you have the native asset here as fallback? 

If the receiver wants EURC and we can't resolve that we should not send ETH instead.

lib/new-ui/pages/send_page.dart
```diff
@@ -963,8 +988,88 @@ class _NewSendPageState extends State<NewSendPage> {
...
+        return targetChainId == evm!.getChainIdByWalletType(currentType);
+      }
+    }
+    return widget.sendViewModel.walletCurrencyName == req.scheme.toLowerCase();
+  }
+
+  int? _evmTargetChainId(PaymentRequest req) {
+    final scheme = req.scheme.toLowerCase();
+    if (scheme == "ethereum") {
+      return req.chainId ?? 1;
```
konstantinullrich: Why fallback to Ehtereum here? Why not use the chain the user is on right now

konstantinullrich: `chain_id is optional and contains the decimal chain ID, such that transactions on various test- and private networks can be requested. If no chain_id is present, the client’s current network setting remains effective.`

https://eips.ethereum.org/EIPS/eip-681

lib/new-ui/widgets/currency_picker/single_network_currency_picker.dart
```diff
@@ -41,65 +44,106 @@ class _SingleNetworkCurrencyPickerState extends State<SingleNetworkCurrencyPicke
...
   }
 
-  double _fiatValueFor(CryptoCurrency c) => _args.balanceByAsset?[c]?.fiatValue ?? 0;
+  late final Map<String, CurrencyPickerBalance> _balanceByKey = {
+    for (final e
+        in (_args.balanceByAsset ?? const <CryptoCurrency, CurrencyPickerBalance>{}).entries)
+      _assetKey(e.key): e.value,
+  };
+
+  String _assetKey(CryptoCurrency c) => c.title.toUpperCase();
```
konstantinullrich: NIT(Claude): Is this a good key" since you could get a symbol collision? Since there is nothing enforcing the uniques of a symbols.

lib/new-ui/widgets/currency_picker/single_network_currency_picker.dart
```diff
@@ -41,65 +44,106 @@ class _SingleNetworkCurrencyPickerState extends State<SingleNetworkCurrencyPicke
     Navigator.of(context).maybePop();
   }
 
-  double _fiatValueFor(CryptoCurrency c) => _args.balanceByAsset?[c]?.fiatValue ?? 0;
+  late final Map<String, CurrencyPickerBalance> _balanceByKey = {
```
konstantinullrich: NIT: Also why not make this a getter?

lib/new-ui/pages/send_page.dart
```diff
@@ -1039,274 +1154,322 @@ class _NewSendPageState extends State<NewSendPage> {
...
+          result.chainId != null) {
+        final appStore = getIt.get<AppStore>();
+        final node = appStore.settingsStore
+            .getCurrentNode(widget.sendViewModel.wallet.type, chainId: result.chainId);
+        await evm!.selectChain(widget.sendViewModel.wallet, result.chainId!, node: node);
       }
+      await widget.sendViewModel.wallet.updateBalance();
+    } catch (e, s) {
+      printV("completeWalletSwitch failed: $e\n$s");
+    } finally {
```
konstantinullrich: If it fails to switch here the user is not notified. This could lead to the user sending funds on the wrong chain without the user knowing.

konstantinullrich (changes requested):

lib/new-ui/pages/swap_page.dart
```diff
@@ -1250,24 +1749,53 @@ class SwapAmountBoxState extends State<SwapAmountBox> {
         ? widget.exchangeViewModel.receiveCurrencies
         : widget.exchangeViewModel.depositCurrencies;
     final currencies = rawCurrencies.whereType<CryptoCurrency>().toList();
-    appendEvmDefaultTokens(currencies);
+    final restrictToCurrentBalances = !widget.isReceiverCard &&
+        widget.balanceByAsset != null &&
+        widget.balanceByAsset!.isNotEmpty;
+    if (!restrictToCurrentBalances) {
```
OmarHatem28: why?

lib/new-ui/pages/send_page.dart
```diff
@@ -963,8 +988,88 @@ class _NewSendPageState extends State<NewSendPage> {
...
+    final req = widget.initialPaymentRequest;
+    if (req == null) {
+      return false;
+    }
+
+    final currentType = widget.sendViewModel.wallet.type;
+    if (evm != null && isEVMCompatibleChain(currentType)) {
+      final targetChainId = _evmTargetChainId(req);
+      if (targetChainId != null) {
+        return targetChainId == evm!.getChainIdByWalletType(currentType);
```
OmarHatem28: why are we setting this to the current type chain id after we get the target?

OmarHatem28 (commented):
The parsing of contracts QRs is disregarding the token decimals, try scannign this QR for 1 USDT, and it parses it to 0.000000000001 USDT

`ethereum:0xdAC17F958D2ee523a2206206994597C13D831ec7@1/transfer?address=0xdAC17F958D2ee523a2206206994597C13D831ec7&uint256=1000000`


also to be honest, I don't like that it's tightly coupled with the swap and the send flow, but for the importance of this feature, I can still merge it after the comments are fixed, but let's work on the improvements and separating it, ASAP

cw_core/lib/currency_for_wallet_type.dart
```diff
@@ -111,3 +111,48 @@ String getCryptoCurrencyIconForWalletListItem(WalletType type,
 
   return walletTypeToCryptoCurrency(type, isTestnet: isTestnet, chainId: chainId).iconPath ?? "";
 }
+
+String? symbolIconPathForWalletType(WalletType type) {
+  const prefix = 'assets/new-ui/card_icons/symbol_icons';
```
malik1004x: Double quotes please. That's what we have the lints for.

lib/new-ui/pages/send_page.dart
```diff
@@ -1319,32 +1573,82 @@ class _NewSendPageState extends State<NewSendPage> {
...
+      walletSwitcherViewModel: widget.walletSwitcherViewModel,
+      fromSend: SwapFromSendArgs(
+        recipientAddress: result.addressDetectionResult?.address ?? "",
+        receiveCurrency: receiveCurrency,
+        targetWalletType: destWalletType,
+        depositBalanceByAsset: depositBalanceByAsset,
+        receiveAmount: receiveAmount,
+      ),
+    );
+    await showModalBottomSheet<void>(
```
malik1004x: Why not just `.push` the page? We're inside a `ModalNavigator`. Showing another page breaks swipe return gestures, animations, and is a worse UX in general.

res/pictures/wallet_filled.svg
malik1004x, line 1: We also already have this graphic.

res/pictures/swap_arrows.svg
malik1004x, line 1: We already have this graphic.

lib/view_model/send/output.dart
```diff
@@ -263,6 +263,7 @@ abstract class OutputBase with Store {
   void updateWallet(
       WalletBase<Balance, TransactionHistoryBase<TransactionInfo>, TransactionInfo> newWallet) {
     _wallet = newWallet;
+    estimatedFee = Money.zero(cryptoCurrencyHandler());
```
malik1004x, line 266: Why this? The fee is auto-calculated by the constructor later anyway.

lib/utils/token_utilities.dart
```diff
@@ -222,6 +222,24 @@ class TokenUtilities {
...
+    for (final token in loadDefaultSolTokensForSwap()) {
+      if (_matchesCurrency(token, currency)) return token.mintAddress;
+    }
+    return null;
+  }
+
+  static String? findTronTokenContract(CryptoCurrency currency) {
+    if (currency is TronToken) return currency.contractAddress;
+
+    for (final token in loadDefaultTronTokensForSwap()) {
```
malik1004x, line 237: Have you tested performance of this? If I remember correctly, these functions perform many database operations.

lib/src/widgets/new_list_row/list_item_regular_row_widget.dart
```diff
@@ -89,16 +89,28 @@ class ListItemRegularRowWidget extends StatelessWidget {
...
-                            colorFilter: iconColor == null
-                                ? null
-                                : ColorFilter.mode(iconColor!, BlendMode.srcIn),
+                          right: -2,
+                          bottom: -2,
+                          child: Container(
+                            width: 16,
+                            height: 16,
+                            decoration: BoxDecoration(
+                              borderRadius: BorderRadius.circular(6),
```
malik1004x, line 98: Wouldn't this add a border to all instances of `ListItemRegularRow`?

lib/new-ui/widgets/swap_page/swap_amount_box.dart
```diff
@@ -0,0 +1,801 @@
+import "package:cake_wallet/core/amount_validator.dart";
```
malik1004x, line 1: Please don't move this – it'll cause conflicts with the swap refactor. I know it should be in its own file eventually, but 70% of this code is different now and we should only move the new version.

lib/new-ui/widgets/anypay/switch_network_wallet_page.dart
```diff
@@ -0,0 +1,107 @@
...
+    final colors = Theme.of(context).colorScheme;
+    return Material(
+      color: colors.surface,
+      child: SafeArea(
+        child: Column(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: [
+            ModalTopBar(
+              title: "",
+              leadingIcon: const Icon(Icons.arrow_back_ios_new),
```
malik1004x, line 49: This needs a semantic label.

lib/new-ui/widgets/anypay/swap_from_network_page.dart
```diff
@@ -0,0 +1,138 @@
+import "package:cake_wallet/generated/i18n.dart";
+import "package:cake_wallet/new-ui/widgets/new_primary_button.dart";
+import "package:cake_wallet/new-ui/widgets/receive_page/receive_top_bar.dart";
+import "package:cake_wallet/src/widgets/cake_image_widget.dart";
+import "package:cw_core/currency_for_wallet_type.dart";
+import "package:flutter/material.dart";
+
+class SwapFromNetworkPage extends StatelessWidget {
```
malik1004x, line 8: I checked the Figma and it seems like `SendToNetworkPage` and `SwapFromNetworkPage` are only different in one graphic and button texts, are you sure these have to be separate?

lib/new-ui/widgets/anypay/select_recipient_network_sheet.dart
```diff
@@ -0,0 +1,241 @@
...
+class NetworkListCard extends StatelessWidget {
+  const NetworkListCard({required this.rows});
+
+  final List<Widget> rows;
+
+  @override
+  Widget build(BuildContext context) {
+    final colors = Theme.of(context).colorScheme;
+    final children = <Widget>[];
+    for (var i = 0; i < rows.length; i++) {
```
malik1004x: Why not just `ListView.separated`?

lib/new-ui/widgets/anypay/select_recipient_network_sheet.dart
```diff
@@ -0,0 +1,241 @@
...
+            const SizedBox(height: 24),
+            Text(
+              S.of(context).select_recipient_network,
+              textAlign: TextAlign.center,
+              style: Theme.of(context)
+                  .textTheme
+                  .titleLarge
+                  ?.copyWith(fontSize: 20, fontWeight: FontWeight.w500, letterSpacing: -0.1),
+            ),
+            const SizedBox(height: 24),
```
malik1004x, line 94: Why all the hardcoded `SizedBox`es? This could be solved with `spacing`. You can make nested columns to get different sizes and at the same time make the layout more flexible.

lib/new-ui/widgets/anypay/recipient_network_row.dart
```diff
@@ -0,0 +1,95 @@
...
+  final WalletBase wallet;
+  final void Function(ChainInfo currentChain) onTap;
+
+  @override
+  Widget build(BuildContext context) {
+    if (evm == null) return const SizedBox.shrink();
+
+    final currentChainId = evm!.getChainIdByWalletType(wallet.type);
+    final currentChain = evm!.getChainInfoByChainId(currentChainId);
+    if (currentChain == null) return const SizedBox.shrink();
```
malik1004x: Don't fall back to `SizedBox.shrink()` for error cases, let it throw. If it throws we have a stacktrace we can do something with instead of a "I got a blank page, can someone help me?" report.

lib/new-ui/widgets/anypay/evm_address_detected_sheet.dart
```diff
@@ -0,0 +1,128 @@
...
+
+class _StackedNetworkIcons extends StatelessWidget {
+  const _StackedNetworkIcons({required this.iconPaths});
+
+  final List<String> iconPaths;
+
+  static const double _size = 50;
+  static const double _step = 34;
```
malik1004x: I'd love these to be constructor fields with default values – we could reuse this in quite a few places.

lib/new-ui/widgets/swap_page/swap_from_send_args.dart
```diff
@@ -0,0 +1,19 @@
...
+    required this.targetWalletType,
+    this.depositBalanceByAsset,
+    this.receiveAmount,
+  });
+
+  final String recipientAddress;
+  final CryptoCurrency receiveCurrency;
+  final WalletType targetWalletType;
+  final Map<CryptoCurrency, CurrencyPickerBalance>? depositBalanceByAsset;
+  final String? receiveAmount;
```
malik1004x: Why not use `Money`?

lib/new-ui/widgets/swap_page/swap_from_send_args.dart
```diff
@@ -0,0 +1,19 @@
...
+    required this.recipientAddress,
+    required this.receiveCurrency,
+    required this.targetWalletType,
+    this.depositBalanceByAsset,
+    this.receiveAmount,
+  });
+
+  final String recipientAddress;
+  final CryptoCurrency receiveCurrency;
+  final WalletType targetWalletType;
```
malik1004x, line 17: Why pass a `WalletType` instead of getting it from `CryptoCurrency`?

lib/new-ui/widgets/swap_page/swap_amount_box.dart
```diff
@@ -0,0 +1,801 @@
...
+    fiatAmountController.dispose();
+    amountFocusNode.dispose();
+    super.dispose();
+  }
+
+  bool _fiatInputMode = false;
+
+  /// Whether the amount field is currently taking fiat input, the page's
+  /// best-rate reaction needs this to know which controller holds the amount.
+  bool get fiatInputMode => _fiatInputMode;
```
malik1004x, line 124: Why? Just rename the field if you need access outside of it.

lib/new-ui/pages/send_page.dart
```diff
@@ -1039,274 +1181,386 @@ class _NewSendPageState extends State<NewSendPage> {
...
+    final currentNetworkName = _networkDisplayName(currentType, currentChainId);
+    final currentNetworkIcon = symbolIconPathForWalletType(currentType) ?? "";
+
+    final hasSingleWallet =
+        result.type == PaymentFlowType.singleWallet || result.wallets.length == 1;
+    final hasMultipleWallets =
+        result.type == PaymentFlowType.multipleWallets || result.wallets.length > 1;
+    final hasWallet = hasSingleWallet || hasMultipleWallets || result.wallet != null;
+
+    final decisionTitle = isEvmTarget
```
malik1004x, line 1226: This should all be handled in the new page's code instead of this helper function.

lib/new-ui/pages/send_page.dart
```diff
@@ -1003,29 +1134,40 @@ class _NewSendPageState extends State<NewSendPage> {
           );
           break;
         case PaymentFlowType.evmNetworkSelection:
-          await _showTokenSelectionFlow(
-            widget.paymentViewModel,
-            widget.walletSwitcherViewModel,
-            paymentRequest,
-            fixedNetwork: result.walletType,
-          );
+          if (result.chainId != null &&
```
malik1004x: Way too much logic for a switch case inside UI-side helper. Either make its own helper function or, preferably, put it in its own class.

lib/utils/token_utilities.dart
```diff
@@ -302,7 +318,17 @@ class TokenUtilities {
         }
       }
 
-      // Then add tokens from CryptoCurrency.all that don't duplicate user tokens
+      if (evm != null) {
```
malik1004x, line 321: How could `evm` be null here?

lib/new-ui/widgets/anypay/send_to_network_page.dart
```diff
@@ -0,0 +1,120 @@
...
+  Widget build(BuildContext context) {
+    final colors = Theme.of(context).colorScheme;
+    return Material(
+      color: colors.surface,
+      child: SafeArea(
+        child: Column(
+          children: [
+            ModalTopBar(
+              title: "",
+              leadingIcon: const Icon(Icons.arrow_back_ios_new),
```
malik1004x, line 36: Semantic labels please.

lib/new-ui/widgets/currency_picker/currency_picker_args.dart
```diff
@@ -52,6 +63,55 @@ class CurrencyPickerBalance {
   final double? fiatValue;
 }
 
+String? _assetAddressKey(CryptoCurrency c) {
+  if (c is Erc20Token) {
```
malik1004x, line 67: Put a getter in a superclass instead of type checks. Will help on future PRs too.

lib/utils/payment_request.dart
```diff
@@ -6,14 +6,27 @@ import 'package:cw_core/payment_uris.dart';
...
 
   factory PaymentRequest.fromString(String input) {
     try {
       return PaymentRequest.fromBolt11(input);
     } catch (_) {
-      return PaymentRequest.fromUri(Uri.parse(input));
+      try {
+        return PaymentRequest.fromUri(Uri.parse(input));
+      } catch (_) {
+        return PaymentRequest(input, '', '', '', null);
```
malik1004x: So, if URI fails to parse, just cram it there? Doesn't seem like the best idea.

Also, this class should probably get a wallet type and derive a chain ID and scheme from that.

lib/new-ui/widgets/swap_page/swap_amount_box.dart
```diff
@@ -0,0 +1,801 @@
...
+      });
+    }
+
+    super.initState();
+  }
+
+  @override
+  void dispose() {
+    if (_memoListener != null) {
+      memoController.removeListener(_memoListener!);
```
malik1004x: This is pointless. Listeners get auto-removed when the field is disposed.

lib/new-ui/widgets/swap_page/swap_amount_box.dart
```diff
@@ -0,0 +1,801 @@
...
+            .getCryptoSymbol(widget.currency as CryptoCurrency)
+        : widget.currency.name.toUpperCase();
+
+    final chainIconPath = (widget.currency is CryptoCurrency)
+        ? _getCurrencyChainIconPath(widget.currency as CryptoCurrency)
+        : null;
+
+    final colors = Theme.of(context).colorScheme;
+
+    if (widget.sourceSelectorMode) {
```
malik1004x, line 147: This is a completely separate tree and thus should be its own widget. That way we can also properly connect it with the swap refactor.

lib/new-ui/widgets/swap_page/swap_amount_box.dart
```diff
@@ -0,0 +1,801 @@
...
+                          : widget.exchangeViewModel.roundedDepositAmount(6),
+                      fiatAmount: widget.isReceiverCard
+                          ? widget.exchangeViewModel.roundedReceiveAmountFiat(6)
+                          : widget.exchangeViewModel.roundedDepositAmountFiat(6),
+                      cryptoCurrencySymbol: currencyToShow,
+                      fiatCurrencySymbol: widget.exchangeViewModel.fiat.symbol,
+                    );
+                  },
+                ),
+                if (!widget.hideWalletPicker)
```
malik1004x: Don't call the field `hideWalletPicker`, the name should be related to the flow you're making. A year from now nobody will know why you'd want to hide the wallet picker.

lib/new-ui/widgets/swap_page/anypay_swap_footer.dart
```diff
@@ -0,0 +1,106 @@
...
+                  color: colors.surfaceContainer,
+                  borderRadius: BorderRadius.circular(24),
+                ),
+                child: Row(
+                  mainAxisSize: MainAxisSize.min,
+                  children: [
+                    if (provider != null)
+                      CakeImageWidget(imageUrl: provider.description.image, width: 24, height: 24)
+                    else
+                      const CupertinoActivityIndicator(),
```
malik1004x, line 84: Isn't this copied from the provider preview widget? Why can't it use that?

lib/new-ui/pages/swap_page.dart
```diff
@@ -272,18 +315,32 @@ class _NewSwapPageState extends State<NewSwapPage> {
...
                 .changeDepositAmount(amount: widget.exchangeViewModel.depositAmount);
-          else
+          } else {
             widget.exchangeViewModel.changeDepositAmount(amount: depositAmountController.text);
+          }
+        }
+      });
+
+      reaction((_) => widget.exchangeViewModel.forcedProviderRate, (double rate) {
+        if (widget.exchangeViewModel.forcedProvider != null && rate == 0) {
```
malik1004x, line 326: Why are we adding this?

lib/new-ui/pages/swap_page.dart
```diff
@@ -123,13 +127,22 @@ class _NewSwapPageState extends State<NewSwapPage> {
       final depositFiatAmountController = depositKey.currentState!.fiatAmountController;
       final receiveFiatAmountController = receiveKey.currentState!.fiatAmountController;
 
+      mobx.ReactionDisposer reaction<T>(T Function(mobx.Reaction) fn, void Function(T) effect) {
```
malik1004x: Why? Just call `_disposers.add()`. We're getting rid of MobX soon anyway. Don't overcomplicate.

lib/new-ui/pages/send_page.dart
```diff
@@ -1039,274 +1181,386 @@ class _NewSendPageState extends State<NewSendPage> {
...
+        alertContent: S.of(context).unsupported_network_requested(chainId.toString()),
+        buttonText: S.of(context).ok,
+        buttonAction: () => Navigator.of(context).pop(),
       ),
     );
   }
 
-  Future<void> _showTokenSelectionFlow(
+  String _networkDisplayName(WalletType type, int? chainId) {
+    if (evm != null && isEVMCompatibleChain(type)) {
```
malik1004x: Why is this here? This could be reused in quite a few places if it weren't part of `SendPage`.

lib/new-ui/widgets/swap_page/swap_from_send_args.dart
```diff
@@ -0,0 +1,19 @@
...
+  final Map<CryptoCurrency, CurrencyPickerBalance>? depositBalanceByAsset;
```
malik1004x, line 18: What's this for? Can't you just get the balances from the wallet? How is it needed if it's not when we open the Swap page the regular way?

lib/new-ui/widgets/currency_picker/currency_picker_args.dart
```diff
@@ -52,6 +63,55 @@ class CurrencyPickerBalance {
...
+    return c.mintAddress.toLowerCase();
+  }
+  return null;
+}
+
+CurrencyPickerBalance? balanceForAsset(
+  Map<CryptoCurrency, CurrencyPickerBalance>? balances,
+  CryptoCurrency asset,
+) {
+  if (balances == null || balances.isEmpty) {
```
malik1004x, line 83: Again, I see no need for this, especially with this amount of safeguards. Just get the balances from the wallet, or if you need a balance for something else, put that in the viewmodel.

lib/new-ui/widgets/anypay/send_to_network_page.dart
```diff
@@ -0,0 +1,120 @@
...
+    required this.onSwap,
+    super.key,
+  });
+
+  final String title;
+  final String destinationNetworkName;
+  final String destinationIconPath;
+  final String currentNetworkName;
+  final VoidCallback onSwitchWallet;
+  final VoidCallback onSwap;
```
malik1004x, line 24: Why pass callbacks instead of just passing in a viewmodel? Do you have more than one callback here?

lib/new-ui/widgets/anypay/evm_address_detected_sheet.dart
```diff
@@ -0,0 +1,128 @@
...
+class EvmAddressDetectedSheet extends StatelessWidget {
+  const EvmAddressDetectedSheet({super.key, required this.networks});
+
+  final List<RecipientNetworkItem> networks;
+
+  static Future<int?> show({
+    required BuildContext context,
+    required List<RecipientNetworkItem> networks,
+  }) {
+    return showModalBottomSheet<int>(
```
malik1004x, line 16: What's with this pattern? We don't use it anywhere else. Just call `showModalBottomSheet`. We can intend to make this a page later.

lib/new-ui/pages/swap_page.dart
```diff
@@ -434,62 +524,78 @@ class _NewSwapPageState extends State<NewSwapPage> {
...
     if (confirmed) {
       widget.exchangeViewModel.feesViewModel.setDefaultTransactionPriority();
     }
   }
 
-  void disposeBestRateSync() => widget.exchangeViewModel.bestRateSync.cancel();
-
   @override
   Widget build(BuildContext context) {
+    final colorScheme = Theme.of(context).colorScheme;
```
malik1004x, line 547: Stop doing that. Just write `Theme.of(context).colorScheme`. You're causing merge conflicts, unnecessary rebuilds and clutter for no apparent reason.

lib/new-ui/pages/swap_page.dart
```diff
@@ -434,62 +524,78 @@ class _NewSwapPageState extends State<NewSwapPage> {
...
+
+    final fromSend = widget.fromSend;
+    final fromType = widget.exchangeViewModel.wallet.type;
```
malik1004x, line 550: Are these not accessible later? Why are you assigning them here?

lib/new-ui/pages/swap_page.dart
```diff
@@ -378,15 +436,34 @@ class _NewSwapPageState extends State<NewSwapPage> {
...
+      }
     });
   }
 
   @override
   void dispose() {
+    for (final disposer in _disposers) {
+      disposer();
+    }
+    _disposers.clear();
```
malik1004x: Why clear the list?

lib/new-ui/pages/send_page.dart
```diff
@@ -1319,32 +1573,82 @@ class _NewSendPageState extends State<NewSendPage> {
...
+
+    final receiveCurrency = resolvedToken ??
+        result.detectedCurrency ??
+        walletTypeToCryptoCurrency(destWalletType, chainId: result.chainId);
+
+    final receiveAmount = paymentRequest.amount.isNotEmpty ? paymentRequest.amount : null;
+
+    final isFiatDisabled = widget.sendViewModel.isFiatDisabled;
+    final depositBalanceByAsset = <CryptoCurrency, CurrencyPickerBalance>{
+      for (final r in widget.sendViewModel.balanceViewModel.formattedBalances)
```
malik1004x, line 1616: Again, we can get these from the wallet at any time instead of passing this.

lib/new-ui/pages/send_page.dart
```diff
@@ -1039,274 +1181,386 @@ class _NewSendPageState extends State<NewSendPage> {
...
-        }
+    bool chainSwitchFailed = false;
+    try {
+      if (evm != null &&
+          isEVMCompatibleChain(widget.sendViewModel.wallet.type) &&
+          result.chainId != null) {
+        final appStore = getIt.get<AppStore>();
+        final node = appStore.settingsStore
+            .getCurrentNode(widget.sendViewModel.wallet.type, chainId: result.chainId);
+        await evm!.selectChain(widget.sendViewModel.wallet, result.chainId!, node: node);
```
malik1004x, line 1400: How does `selectChain` work? What if the user has separate wallets for different EVM chains (which most of them do as wallet groups aren't that popular)?

lib/new-ui/pages/send_page.dart
```diff
@@ -1003,29 +1134,40 @@ class _NewSendPageState extends State<NewSendPage> {
...
+              evm != null &&
+              evm!.getChainInfoByChainId(result.chainId!) == null) {
+            _showUnsupportedNetworkAlert(result.chainId!);
+            return;
+          }
+
+          final targetChainId =
+              paymentRequest.scheme.isNotEmpty ? _evmTargetChainId(paymentRequest) : null;
+          final targetChain = targetChainId != null && evm != null
+              ? evm!.getChainInfoByChainId(targetChainId)
```
malik1004x: How can `evm` be null here?

lib/new-ui/pages/send_page.dart
```diff
@@ -1003,29 +1134,40 @@ class _NewSendPageState extends State<NewSendPage> {
...
+              : null;
+          final currentChainId =
+              evm != null && isEVMCompatibleChain(widget.sendViewModel.wallet.type)
+                  ? evm!.getChainIdByWalletType(widget.sendViewModel.wallet.type)
+                  : null;
+          final isCrossChain = targetChain != null && targetChain.chainId != currentChainId;
+
+          if (isCrossChain) {
+            await _handleEvmNetworkFlow(targetChain, paymentRequest);
+          } else if (widget.sendViewModel.isEVMWallet) {
```
malik1004x, line 1155: All these extra conditions for EVM (not just this one) should be handled as one function that just redirects you to the EVM-specific page.

malik1004x (changes requested):

lib/new-ui/pages/swap_page.dart
Blazebrain: Extracted it

lib/new-ui/pages/send_page.dart
Blazebrain: Yeah, that's true

I've added a stop when its an unsupported token

lib/new-ui/pages/send_page.dart
Blazebrain: Added the fallback to the current chain the user is on according to the spec

lib/new-ui/widgets/currency_picker/single_network_currency_picker.dart
Blazebrain: Already modfied this, so its no more needed here

lib/new-ui/pages/send_page.dart
Blazebrain: Yeah, I've added an alert to notify the user, and abort all pending operations

lib/new-ui/pages/send_page.dart
Blazebrain: We're not setting it to the current chain type

its the comparison sign `==` that's used here, not the assignment sign`=`

lib/new-ui/pages/swap_page.dart
Blazebrain: Its for the deposit picker on the AnyPay swap page, we're restricting the list to the assets that the user has, in his balance, as that's what he can currently fund the swap from.

Adding the others would mean the user would be seeing 0 balances, and we can't use that in the AnyPay flow as this is the part where the user sends the asset from their own balance.

cw_core/lib/currency_for_wallet_type.dart
Blazebrain: Alright

lib/new-ui/pages/send_page.dart
Blazebrain: True, would switch to `.push`

res/pictures/wallet_filled.svg
Blazebrain: Would check the existing graphics for it

res/pictures/swap_arrows.svg
Blazebrain: Would check the existing graphics for it

lib/view_model/send/output.dart
Blazebrain: No, your assumption is off. 

We don't have any constructor run on wallet switch and the `updateWallet` mutates the same `Output` instance. The recalculation would only happen later via sync status run. 

The Money.zero reset would clear up a stale fee being displayed on wallet switch, before the calculation happems. This was added in because of an earlier bug reported.

lib/utils/token_utilities.dart
Blazebrain: No, they don't do any database operation.

The functions `findTronTokenContract` and `findSolanaTokenMint` iterate over `getDefaultSPLTokens()` and `getDefaultTronTokens()`, and these are static in-memory lists we call via the proxy.

They're good to go.

lib/src/widgets/new_list_row/list_item_regular_row_widget.dart
Blazebrain: No, it won't. 

It only renders the border when `badgeIconPath != null`, and only two places in the codebase set `badgeIconPath`

lib/new-ui/widgets/swap_page/swap_amount_box.dart
Blazebrain: This was recommended by a previous review comment, and its has already been pushed. Reverting now would churn more than it saves.

You should make the adjustment in the swap refactor instead

lib/new-ui/widgets/anypay/switch_network_wallet_page.dart
Blazebrain: Alright, would add it in.

lib/new-ui/widgets/anypay/swap_from_network_page.dart
Blazebrain: They can be kept separate, a merged page for the separate flows would need more flags and all

It can be further improved in the improvement PR landing soon

lib/new-ui/widgets/anypay/select_recipient_network_sheet.dart
Blazebrain: Not really necessary

`NetworkListCard` is a shrink wrapped rounded card inside an outer SingleChildScrollView. Using `ListView.separated` there needs `shrinkWrap` + disabled physics, and that is heavier than a Column for a handful of rows.

lib/new-ui/widgets/anypay/select_recipient_network_sheet.dart
Blazebrain: fair, would switch it to spacing

lib/new-ui/widgets/anypay/recipient_network_row.dart
Blazebrain: This is tricky though, especially for Monero.com where `evm = null`, don't think we should throw here. I'll add a `printV` there

lib/new-ui/widgets/anypay/evm_address_detected_sheet.dart
Blazebrain: that makes sense, would push

lib/new-ui/widgets/swap_page/swap_from_send_args.dart
Blazebrain: Yeah, that's true. would switch

lib/new-ui/widgets/swap_page/swap_from_send_args.dart
Blazebrain: We get both the `walletType` and `CryptoCurrency` from the result, and they're non null. deriving would just introduce an avoidable nullable result that we already have available.

lib/new-ui/widgets/swap_page/swap_amount_box.dart
Blazebrain: That's not a good approach.

Adding a getter is deliberate there. the field is mutated via setState, and a public field would just invite external writes that bypass setState. Totally unneeded.

lib/new-ui/pages/send_page.dart
Blazebrain: We discussed the `AnyPayService` earlier, and work is ongoing. All of these would be in there.

lib/utils/token_utilities.dart
Blazebrain: Monero.com builds would always have `evm` as null, not an overkill.

lib/new-ui/pages/send_page.dart
Blazebrain: Ditto `AnyPayService`

lib/new-ui/widgets/anypay/send_to_network_page.dart
Blazebrain: Will add, same as the other page

lib/new-ui/widgets/currency_picker/currency_picker_args.dart
Blazebrain: Yeah, that's valid. I'll prefer having it as a separate improvement PR though, as it cuts across cw_core.

lib/utils/payment_request.dart
Blazebrain: The fallback is deliberate fail-soft, the raw input lands in the address field where the chain validator visibly rejects it. The alternative we have is an unhandled exception escaping the async scan callback

Also, deriving walletType/chainId inside PaymentRequest inverts the design, and not advisable. It's to be a parse result, simple. Detection is for `UniversalAddressDetector` to do downstream. Out of this PR's scope.

lib/new-ui/widgets/swap_page/swap_amount_box.dart
Blazebrain: Valid, would remove

lib/new-ui/widgets/swap_page/swap_amount_box.dart
malik1004x: We're refactoring this whole thing in a week. Nobody's going to do any external writes as this logic will be gone.

lib/new-ui/widgets/swap_page/swap_amount_box.dart
Blazebrain: Yep, planned for the `AnyPayService` improvement PR

lib/new-ui/widgets/swap_page/swap_from_send_args.dart
malik1004x: Why would it be null?

lib/new-ui/widgets/swap_page/swap_amount_box.dart
Blazebrain: good catch, already removed its callers in the updated design UI shared, would delete it

lib/utils/token_utilities.dart
malik1004x: Then doesn't this break on user-added tokens?

lib/src/widgets/new_list_row/list_item_regular_row_widget.dart
malik1004x: Then it makes no sense. `badgeIconPath` doesn't imply it would add a border. Pass a border as a separate field.

lib/new-ui/widgets/swap_page/anypay_swap_footer.dart
Blazebrain: The provider chip is the only shared widget there, every other thing in the footer is different, the pill, layout, rate handling etc. I don't think we should force a shared widget for one chip

lib/new-ui/pages/swap_page.dart
Blazebrain: No we aren't. It's not a new change. `forcedProviderRate` reaction is already pre-existing on dev.

lib/new-ui/pages/swap_page.dart
Blazebrain: nit, but would remove

lib/new-ui/pages/send_page.dart
Blazebrain: Didn't need it outside here currently, but we can move to the evm helpers I guess

lib/new-ui/widgets/swap_page/swap_amount_box.dart
malik1004x: This will cause merge conflicts. Please move it back, it costs nothing – perhaps changing a few imports.

lib/new-ui/widgets/swap_page/swap_from_send_args.dart
Blazebrain: Partially valid, but the maps carry BalanceViewModel-formatted amounts + fiat (including fiat-disabled handling), which a dumb widget cannot compute.

I can do a modification of the vm computed balances in the refactor.

lib/new-ui/widgets/anypay/swap_from_network_page.dart
malik1004x: It won't need new flags if you add a proper service layer.

lib/new-ui/widgets/currency_picker/currency_picker_args.dart
Blazebrain: Direct lookup would miss because picker items are default-list token instances while balance keys are plain constants (Erc20Token ≠ CryptoCurrency.usdterc20), so address-first resolution is what makes balances show at all. 

I'll fold this into the VM-computed balances in the refactor.

lib/new-ui/widgets/anypay/send_to_network_page.dart
Blazebrain: Two callbacks are very visible in the code and VM would be unnecessarily heavier for these

lib/new-ui/widgets/anypay/evm_address_detected_sheet.dart
Blazebrain: Factually off. We use it in other places. Some of it is `CurrencyPickerSheet.show` and `FiatCurrencyPickerSheet.show`.

lib/new-ui/widgets/anypay/select_recipient_network_sheet.dart
malik1004x: I know it's not necessary. But that's the Flutter way of doing things, it makes the UI more readable and maintainable. At least use `.map` if you don't like it.

lib/new-ui/widgets/anypay/recipient_network_row.dart
malik1004x: How would this ever be reached in Monero.com?

lib/utils/token_utilities.dart
malik1004x: Where would we load tokens in Monero.com? How is this reachable in that variant?

lib/new-ui/pages/swap_page.dart
Blazebrain: The rebuild claim is technically incorrect (Theme.of registers the dependency either way), and the local variable style is used in multiple places even in new-ui.

lib/new-ui/pages/swap_page.dart
Blazebrain: Each is used many times below; the alternative is `widget.fromSend` in 8 other places. Non-issue.

lib/new-ui/pages/swap_page.dart
Blazebrain: valid nit, i'll remove that

lib/utils/payment_request.dart
malik1004x: > unhandled exception

Then handle the exception.

lib/new-ui/pages/send_page.dart
Blazebrain: Same as the balances comment above, VM-computed in the refactor

All in the improvement PR landing soon

lib/new-ui/pages/send_page.dart
Blazebrain: The flow switches to a wallet of the target type first (if there are separate wallets of that evm chain, then the user gets a page to choose which one to switch to)

All of these is handled and reflected in the code.

lib/new-ui/pages/send_page.dart
Blazebrain: Oh shoot, yeah. `evmNetworkSelection` already makes it non-null, will remove

lib/new-ui/pages/send_page.dart
Blazebrain: Same, this lands in the AnyPayService PR

lib/new-ui/widgets/swap_page/swap_from_send_args.dart
malik1004x: Passing pre-formatted amounts across pages is bad. You can get the balances from the wallet and format them however is required for the given widget.

lib/utils/token_utilities.dart
Blazebrain: No it doesn't

User-added tokens are `SPLToken`/`TronToken` instances, and the first branch in both finder functions returns their mint/contract directly. The default lists checks is only the fallback for the plain CryptoCurrency constants. 

By the way, i handled all of these in the unit tests

lib/utils/payment_request.dart
Blazebrain: The fallback is the handling. It turns the input into to the plain address and the address validator returns a visible error instead of the scan failing silently

lib/utils/token_utilities.dart
Blazebrain: Same response as above regarding Monero.com

lib/new-ui/widgets/swap_page/swap_amount_box.dart
Blazebrain: If its gone in a week, changing it now gains nothing then. Leaving it as is

lib/new-ui/widgets/swap_page/swap_from_send_args.dart
Blazebrain: `cryptoCurrencyOrTokenToWalletType` returns `WalletType?`

Why would we go for a nullable response when we already have a valid `WalletType` response from the result?

Bypasses unnecessary null checks or expansions

lib/new-ui/widgets/anypay/recipient_network_row.dart
Blazebrain: It isn't reachable in Monero.com today, EVM detections blocks it as incompatible before any token loading. 

The guard is the convention we use for any proxy access in lib/ currently, and this makes the function stay safe regardless of what calls it, rather than assuming its callers has guards.

lib/new-ui/widgets/anypay/swap_from_network_page.dart
Blazebrain: True, and the service layer is coming in with the improvement PR as I mentioned earlier

lib/src/widgets/new_list_row/list_item_regular_row_widget.dart
Blazebrain: The path doesn't add a border onto anything existing, it renders the whole badge, and the pill is the badge's design, same as the icon inside it. 

There's exactly one badge style in the app (the two chain badges), so a separate styling field would be config that nothing varies on. The moment a second badge style exists I'll parameterize it.

OmarHatem28 (approved):

OmarHatem28:
Looking forward for the improvements PR and making this very loosely coupled with everything else and separated in its own service layer and each function in that layer has a single responsibility

PR #3389, opened 2026-07-09
payjoin: exclude 0-conf inputs from receiver candidates

### LINKED ISSUE #3388: Payjoin receiver can contribute unconfirmed (0-conf) inputs
the payjoin receiver's candidate list filters only `isSending && !isFrozen`, never confirmation depth, so the receiver can contribute a 0-conf input.

this is a temporal fingerprint: an unconfirmed-ancestor spend is observable on-chain and lets an analyst partition a payjoin's inputs by owner (sender and receiver run different software; any axis where they diverge separates their inputs). A 0-conf parent can also be RBF'd out, invalidating the payjoin.

fix: exclude 0-conf from the payjoin candidates only (`getUtxoWithPrivateKeys(confirmedOnly: true)` in the receiver path). Scoped to the payjoin receiver; signing and normal sends are unaffected.

ref: https://github.com/payjoin/rust-payjoin/issues/1597#issuecomment-4649767163

bc1cindy:
the payjoin receiver's candidate list (`getCandidateInputs` → `getUtxoWithPrivateKeys()`) filtered only `isSending && !isFrozen`, never confirmation depth, so the receiver could contribute a 0-conf input, a temporal fingerprint (an unconfirmed-ancestor spend is observable and lets an analyst partition the payjoin's inputs by owner), and a 0-conf parent can be RBF'd out, invalidating the payjoin.

adds an opt-in `confirmedOnly` flag to `getUtxoWithPrivateKeys`, set only on the receiver candidate path. Scoped to payjoin receiving; signing and normal sends are unaffected.

ref: https://github.com/payjoin/rust-payjoin/issues/1597#issuecomment-4649767163

closes #3388

### DIFF
--- a/cw_bitcoin/lib/bitcoin_wallet.dart
+++ b/cw_bitcoin/lib/bitcoin_wallet.dart
@@ -551,8 +551,8 @@ abstract class BitcoinWalletBase extends ElectrumWallet with Store {
     return tx;
   }
 
-  List<UtxoWithPrivateKey> getUtxoWithPrivateKeys() => unspentCoins
-      .where((e) => (e.isSending && !e.isFrozen))
+  List<UtxoWithPrivateKey> getUtxoWithPrivateKeys({bool confirmedOnly = false}) => unspentCoins
+      .where((e) => e.isSending && !e.isFrozen && (!confirmedOnly || (e.confirmations ?? 0) > 0))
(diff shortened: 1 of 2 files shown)


OmarHatem28 (approved):
Yup, Thanks for updating it

PR #3393, opened 2026-07-11
check if it will affect the performance of wallet loading time and wa…

OmarHatem28:
…llet usage in general

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_core/lib/wallet_addresses.dart
+++ b/cw_core/lib/wallet_addresses.dart
@@ -71,11 +71,12 @@ abstract class WalletAddresses {
   Future<void> saveAddressesInBox() async {
     try {
       walletInfo.address = address;
-      walletInfo.setAddresses(addressesMap);
-      walletInfo.setAddressInfos(addressInfos);
-      walletInfo.setUsedAddresses(usedAddresses.toList());
-      walletInfo.setHiddenAddresses(hiddenAddresses.toList());
-      walletInfo.setManualAddresses(manualAddresses.toList());
+      // TODO: check if it will affect the performance of each wallet
+      await walletInfo.setAddresses(addressesMap);
+      await walletInfo.setAddressInfos(addressInfos);
+      await walletInfo.setUsedAddresses(usedAddresses.toList());
+      await walletInfo.setHiddenAddresses(hiddenAddresses.toList());
+      await walletInfo.setManualAddresses(manualAddresses.toList());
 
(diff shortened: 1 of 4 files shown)


cw_core/lib/wallet_addresses.dart
```diff
@@ -71,11 +71,12 @@ abstract class WalletAddresses {
   Future<void> saveAddressesInBox() async {
     try {
       walletInfo.address = address;
-      walletInfo.setAddresses(addressesMap);
-      walletInfo.setAddressInfos(addressInfos);
-      walletInfo.setUsedAddresses(usedAddresses.toList());
-      walletInfo.setHiddenAddresses(hiddenAddresses.toList());
-      walletInfo.setManualAddresses(manualAddresses.toList());
+      // TODO: check if it will affect the performance of each wallet
```
MrCyjaneK, line 74: ~~Performance aside we may throw errors when things aren't getting saved in db or cause display issues. definitely should be either awaited or not done at all~~

OmarHatem28: yeah this is why I added the await because it indeed was throwing some errors, but not sure how frequent this save function is called, but I assume a lot, so we might need to make that in a different thread or so

MrCyjaneK: ~~I can retry that with a profiler but I wouldn't just remove the awaits.~~

~~Also for app performance wise it shouldn't matter if we do await or not as both of these are still on the main thread. The only difference is when the function return - and I'd assume that callers of this function would prefer for this data to be persisted before returning.~~

OmarHatem28: if not awaited, the function completes and the main thread is not blocked, they are just added to the queue and whenever the main thread is idle it takes from the queue.
and I don't think all the caller want this function to fully complete before returning, but you are right, they can skip awaiting this function as a whole in that case.
so yeah, please check the callers of this function in profile mode as well, and see if they need to await it or not, or if it affects the performance first or not, cuz if it doesn't then no need to check if we should await the function call itself or not

MrCyjaneK: sorry lol I've fully misread the diff and confused + and -. Change looks good, there are no noticeable differences

MrCyjaneK (approved):

PR #3395, opened 2026-07-11
fix android CI

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/pubspec_base.yaml
+++ b/pubspec_base.yaml
@@ -205,8 +205,10 @@ dependency_overrides:
       ref: 7bcb00380783f8611f68f5cd63fec96261fea2c2
   ledger_usb_plus:
     git:
-      url: https://github.com/konstantinullrich/ledger-usb-plus
-      ref: dad482c847c0d0eedceabbb40d3478e64b65e630
+      # url: https://github.com/konstantinullrich/ledger-usb-plus
+      # ref: dad482c847c0d0eedceabbb40d3478e64b65e630
+      url: https://github.com/cake-tech/ledger-usb-plus
+      ref: 734c3d6f8267ba52ad73b6e1e78e570c07298890
   socks5_proxy:
     git:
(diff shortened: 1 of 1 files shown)


MrCyjaneK:
probably will get fixed by #3390

Blazebrain:
> probably will get fixed by #3390

This is to get the CI back up and running as its slowing down tests with it failing. #3390 is bringing in all the fixes ledger related. 

We don't have to wait for it (if its not ready for merge), as the CI failing is delaying other builds

MrCyjaneK:
Yes I know but it changes the ledger dependency - so we need to run full tests on ledger on this build before merging as the git hash differs.

MrCyjaneK:
Git hash that I would be more supportive of merging now is 

46d146db5d3d79b1c14051c477b9ddef9e4dc287

as that's the upstream merge commit. Two commits on top shouldn't affect anything but since this is a native plugin and it touches native code in android/ dir upstream I wouldn't just assume it does nothing (and definitely wouldn't be for merging it as just CI fix)

Blazebrain:
> Yes I know but it changes the ledger dependency - so we need to run full tests on ledger on this build before merging as the git hash differs.

Yeah, we can always run the tests as needed, however the difference between the old hash and the new hash is a changelog file. By the way, the old hash dad482c8 (konstantinullrich/ledger-usb-plus) was force-pushed on upstream and that's what caused the failure, this simply replaces it with minimal difference

here's exactly what the hash change is
```
$ git diff dad482c8 734c3d6 --stat
 CHANGELOG.md | 4 ++++
 pubspec.yaml | 2 +-      # version 1.0.6 -> 1.0.7
```

MrCyjaneK:
oh I had a wrong tab open and check diff against the actual upstream... Force push on main branch..

MrCyjaneK (approved):

PR #3397, opened 2026-07-13
add-exchange-rate

konstantinullrich:
# Description

This adds an exchange rate class for future use in Swap flow and the fiat api

# Pull Request - Checklist  

- [x] Format code


cw_core/lib/amount/exchange_rate.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:cw_core/amount/money.dart';
+import 'package:cw_core/currency.dart';
+
+class ExchangeRate {
+  /// The currency being priced (e.g. BTC in a BTC/USD pair).
+  final Currency baseCurrency;
```
malik1004x: Nit: probably should be called `base` to allow shorter lines and for parity with naming in charts.

malik1004x (approved):

OmarHatem28 (commented):
@konstantinullrich fix the tests parameter name and add tests for USDTerc20 from CryptoCurrency class and from ERC20 token. probably it will fail.
so either need to modify the condition and make it explicit or make sure that this is only used in swaps, where we only get the currencies from CryptoCurrency.all
I would prefer the first

PR #3408, opened 2026-07-15
Privacy: randomized coin selection (BnB changeless + single random draw)

### LINKED ISSUE #3407: Coin selection follows a predictable order (privacy fingerprint)
Coin selection for BTC/LTC sends walks `unspentCoins` in a fixed, predictable order, which acts as a wallet fingerprint.

`unspentCoins` is built address by address, in `allAddresses` order (the wallet's address generation order). Within a single address, `fetchUnspent` resolves coins concurrently, but with a warm cache they complete in the order the server returned them, `listunspent` is blockchain order, i.e. by block height ≈ oldest first. So the pool order is: address generation order, then coin age.

the greedy selection in _createUTXOS (electrum_wallet.dart) accumulates over that pool in order, so the input set is deterministic: coins from earlier addresses and older coins are always spent first. An analyst who clusters the wallet's addresses can predict/recognize its selection pattern. The only randomization today is network jitter on cold fetches, accidental and gone once the cache warms.

selection should be deliberately non-deterministic (e.g. single random draw, as in Core/bdk), ideally with branch-and-bound for changeless exact matches.

bc1cindy:
coin selection walked `unspentCoins` in a fixed order (address generation order, then coin age), so the input set was deterministic, a fingerprint an analyst can exploit, and always produced residual change.

selection now works like Core/bdk (`BranchAndBound<SingleRandomDraw>`):

- branch-and-bound: searches over effective values for an input set whose excess over amount + fee stays below dust → the change output is dropped and the residue goes to fee (changeless tx, overpay bounded by dust).
- single random draw fallback: no match → the pool is shuffled with `Random.secure()` before the existing accumulate-until-covered loop.
- the RBF fee-bump top-up and payjoin receiver candidates are also shuffled.

closes #3407 

related to https://github.com/payjoin/rust-payjoin/issues/1597

### DIFF
--- a/cw_bitcoin/lib/electrum_wallet.dart
+++ b/cw_bitcoin/lib/electrum_wallet.dart
@@ -899,6 +899,16 @@ abstract class ElectrumWalletBase
   bool _isBelowDust(BigInt amount) =>
       amount <= networkDustAmount && network != BitcoinNetwork.testnet;
 
+  // Random draw priority per outpoint. Assigned lazily with a secure RNG and kept
+  // until the next createTransaction call, so the recursive estimate passes of one
+  // transaction build all see the same input order (reshuffling between passes made
+  // fee estimation unstable). Cleared per transaction so each send is a fresh draw.
+  final Random _coinSelectionRng = Random.secure();
+  final Map<String, int> _coinSelectionOrder = {};
+
+  int _coinSelectionPriority(BitcoinUnspent utx) => _coinSelectionOrder.putIfAbsent(
+      '${utx.hash}:${utx.vout}', () => _coinSelectionRng.nextInt(1 << 32));
+
   UtxoDetails _createUTXOS({
     required bool sendAll,
     required bool paysToSilentPayment,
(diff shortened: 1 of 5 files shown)


cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -907,8 +911,41 @@ abstract class ElectrumWalletBase
...
+    // stays below dust, so the caller drops the change output and the send is changeless.
+    // When no such set exists, the shuffled pool below acts as a single random draw.
+    final canTryChangeless = !sendAll &&
+        inputsCount == null &&
+        credentialsAmount > 0 &&
+        feeRate > 0 &&
+        outputsCount > 0 &&
+        !availableInputs.any((u) => u.bitcoinAddressRecord.type == SegwitAddresType.mweb);
+    if (canTryChangeless) {
+      final perInputVBytes = estimatedTransactionSize(1, 0) - estimatedTransactionSize(0, 0);
```
OmarHatem28: well, this only takes into consideration Segwit size, but it will overestimate for taproot and will not calculate it correctly for legacy address types and other coins that inherit electrum functionality (doge for example)

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -907,8 +911,41 @@ abstract class ElectrumWalletBase
...
 
-    // sort the unconfirmed coins so that mweb coins are last:
-    availableInputs.sort((a, b) => a.bitcoinAddressRecord.type == SegwitAddresType.mweb ? 1 : -1);
+    // Single Random Draw: shuffle the pool so selection is non-deterministic, removing the
+    // predictable address/scan order (a fingerprint). MWEB coins are kept last afterwards.
+    availableInputs.shuffle(Random.secure());
+    availableInputs = [
+      ...availableInputs.where((u) => u.bitcoinAddressRecord.type != SegwitAddresType.mweb),
+      ...availableInputs.where((u) => u.bitcoinAddressRecord.type == SegwitAddresType.mweb),
+    ];
```
OmarHatem28, line 959: This would shuffle the list for each trial, I think we should persist the inputs order until the transaction is created, since this function is called multiple times.
Given that BNB should find the best outcome in its recursive nature, then it wouldn't need this function to be recursive as well, but for other types and wallets that uses this function (BCH, Doge for example) it wouldn't act correctly and will go into multiple recursive calls

cw_bitcoin/lib/electrum_wallet.dart
bc1cindy: fixed in 578b5994

each input is now charged its own script-type size (legacy 148, P2SH 91, P2WPKH 68, taproot 58), and outputs are priced by type too

cw_bitcoin/lib/electrum_wallet.dart
bc1cindy: fixed in 9e9d59b5

inputs now keep a fixed random order for the whole transaction build, re-drawn per transaction

OmarHatem28 (approved):
It's a bit over-engineered and there are some cases it will miss, and it will probably cause it to into un-necessary recursive calls, but it's a good base to enhance on top of it.
Thanks @bc1cindy 🙏 
but please test the changes you make and ensure whatever tool you are using, have more context to the other parts of the project. not just what it modifies

PR #3441, opened 2026-07-24
Fix mismatched support links for SwapTrade and Telegram

sethforprivacy:

# Description

Two entries in `SupportViewModel`'s link list displayed one destination but opened another.

### SwapTrade — wrong provider

`lib/view_model/support_view_model.dart` pointed SwapTrade's link at `mailto:support@exolix.com`, while the visible label read `help.swaptrade.io`. A user with a SwapTrade trade problem tapped SwapTrade support and emailed **Exolix** — who cannot help them, and who absorbed another provider's support traffic.

The bug is older than the SwapTrade name. `git log -S` shows the entry was added as **Quantex** with `linkTitle: 'help.myquantex.com'` and `link: 'mailto:support@exolix.com'`, so Exolix's mailto was copy-pasted from the entry directly above it from the start. ef377fb53 ("Change Quantex to SwapTrade") updated the title, icon, and `linkTitle`, but carried the wrong `link` forward.

Now points at `https://help.swaptrade.io/contact/`:

- Matches what `docs.cakewallet.com/support/swap` already lists for SwapTrade ("Live Chat" via `help.swaptrade.io/contact/`), so the app and the public docs agree.
- Live chat is the **only** contact method that help center publishes — no SwapTrade support email exists — so a `mailto:` alternative isn't available even if we'd prefer one.
- A path below the displayed host matches existing style in this file: the commented-out SideShift entry pairs `linkTitle: 'help.sideshift.ai'` with `link: 'https://help.sideshift.ai/en/'`.

### Telegram — stale label

The inverse problem: `linkTitle` read `t.me/cakewallet` while the link went to `https://t.me/cakewalletannouncements`.

Here the **link** was the correct half. `t.me/cakewallet` exists as a group named "Cake Wallet Community", but self-describes as *"Inactive Cake Wallet Community"* and tells visitors to go to the announcements channel instead. `t.me/cakewalletannouncements` is an active channel (851 subscribers). Sending users to the group the group itself disowns would be a regression, so the stale label was updated to match the destination rather than the reverse.

### Full audit

All 17 entries were checked, not just the reported one. The remaining 15 are consistent: the `mailto:` entries all display the address they open (Email, ChangeNow, SimpleSwap, Exolix, Trocador, DFX, Kryptonim), the `https://` entries display their host (Website, Forum, Discord, Telegram Support Bot), and four use intentionally generic labels where a literal match doesn't apply — GitHub (`S.current.apk_update` → releases), Onramper (`View exchanges` → docs), MoonPay and Robinhood (`S.current.submit_request` → their request forms). Each of those four lands where its label promises.

## For reviewers

**Worth confirming with whoever owns the provider relationships.** `help.swaptrade.io` loads and live chat is present, but the page's footer and social links still show **Quantex** branding — it even lists `admin@myquantex.com` and `help.myquantex.com/contact/`. It's the right destination today and the one our docs advertise, but SwapTrade's help center looks only half-rebranded, so it's worth a check that this is the surface we want users landing on.

Data-only change, so `support_view_model.g.dart` needs no regeneration.

# Pull Request - Checklist

- [x] Format code

Both destination URLs were verified as loading and correct by fetching them directly, but the app itself was not built or launched, so the two manual-test boxes are left unchecked for whoever picks this up.

### DIFF
--- a/lib/view_model/support_view_model.dart
+++ b/lib/view_model/support_view_model.dart
@@ -44,7 +44,7 @@ abstract class SupportViewModelBase with Store {
           LinkListItem(
               title: 'Telegram',
               icon: 'assets/images/Telegram.png',
-              linkTitle: 't.me/cakewallet',
+              linkTitle: 't.me/cakewalletannouncements',
               link: 'https://t.me/cakewalletannouncements'),
           LinkListItem(
               title: 'Telegram Support Bot',
@@ -75,7 +75,7 @@ abstract class SupportViewModelBase with Store {
               title: 'SwapTrade',
               icon: 'assets/images/swap_trade.png',
               linkTitle: 'help.swaptrade.io',
-              link: 'mailto:support@exolix.com'),
+              link: 'https://help.swaptrade.io/contact/'),
           LinkListItem(
               title: 'Trocador',
               icon: 'assets/images/trocador.png',


sethforprivacy:
### Note on the failing `internal-build / PR_test_build (29)` check

This failure is **not caused by this PR** — it's the pre-existing CI bug with slash-containing branch names, being fixed in #3442.

The job ran for 34m48s and got all the way through `flutter build apk` successfully. It died afterwards, in the "Rename apk file" step:

```
cp: cannot create regular file 'test-apk/claude/swaptrade-support-link-fix-8d2ed5.apk': No such file or directory
```

`reusable-build.yml` interpolates the raw branch name into a file path, but the preceding line creates only a flat `mkdir test-apk`. This branch is `claude/swaptrade-support-link-fix-8d2ed5`, so `cp` needs a `test-apk/claude/` subdirectory that was never created. Any branch with a `/` in its name fails at this step regardless of its contents.

So the red check here is APK-packaging/Slack-upload plumbing, not code. Supporting evidence that this change itself is fine:

- the Flutter build completed before the failing step, so the modified Dart compiled
- the other four `PR_test_build` jobs pass
- `check_restricted_imports` passes

Once #3442 lands, rebasing this branch on `dev` should turn the check green. This PR is a two-line data-only edit to `lib/view_model/support_view_model.dart` and deliberately does not touch CI.

OmarHatem28 (approved):

PR #3442, opened 2026-07-24
ci: sanitize branch name when naming test APK in reusable-build

sethforprivacy:
## Problem

The `internal-build` job's **Rename apk file** / **Find APK file** steps in `reusable-build.yml` use the raw `${BRANCH_NAME}` (`github.head_ref`) as the APK filename:

```bash
cp app-arm64-v8a-release.apk test-apk/${BRANCH_NAME}.apk
```

For any branch whose name contains a `/` (e.g. `claude/my-fix`), this expands to `test-apk/claude/my-fix.apk`, and since the intermediate `test-apk/claude/` directory is never created, `cp` fails with:

```
cp: cannot create regular file 'test-apk/<branch>/<name>.apk': No such file or directory
```

which fails the build for slash-named branches.

## Fix

Derive a sanitized branch name from `android/app.properties` (already written earlier in the job as `id=com.cakewallet.test_<sanitized>`), and use it for the APK filenames. This is the exact approach `automated_integration_test.yml` already uses, so the two workflows now behave consistently.

The Slack upload filename still uses the human-readable branch name — that's metadata, not a local path, so slashes there are harmless.

## Notes

Split out of #3439, where this failure first surfaced. This is CI-only; no app code changes.

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -176,16 +176,18 @@ jobs:
 
       - name: Rename apk file
         run: |
+          sanitized_branch_name=$(grep '^id=' android/app.properties | cut -d'=' -f2 | sed 's/com\.cakewallet\.test_//')
           cd build/app/outputs/flutter-apk
           mkdir test-apk
-          cp app-arm64-v8a-release.apk test-apk/${BRANCH_NAME}.apk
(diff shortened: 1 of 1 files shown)


MrCyjaneK (approved):
👍, though I'd prefer for this PR to have `/` in the name since it fixes the issue :p

PR #3446, opened 2026-07-27
Cw 1551 quick bitcoin wallet sync

serhii-bor:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -92,66 +92,25 @@ abstract class ElectrumWalletBase
...
+            currency != null ? {currency: initialBalance ?? _zeroBalance(currency)} : {}),
         this.unspentCoinsInfo = unspentCoinsInfo,
         this.isTestnet = !network.isMainnet,
         this._mnemonic = mnemonic,
+        this.electrumClient = electrumClient ?? electrum.ElectrumClient(),
         _useLightning = useLightning,
         super(walletInfo, derivationInfo) {
-    this.electrumClient = electrumClient ?? electrum.ElectrumClient();
-    this.walletInfo = walletInfo;
-    this.derivationInfo = derivationInfo;
```
OmarHatem28, line 112: why are we not setting this anymore? you are still using it below

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -92,66 +92,25 @@ abstract class ElectrumWalletBase
...
-        sideHdByType[type] = _masterHD!.derivePath("$accountPath/1") as Bip32Slip10Secp256k1;
-      }
-    } else {
-      // View-only wallet (xpub only)
-      for (final type in supportedTypes) {
-        mainHdByType[type] = mainHd;
-        sideHdByType[type] = sideHd;
-      }
-    }
+    _balanceDisplayedForAccount = currentAccountIndex;
```
OmarHatem28, line 111: misleading name for what it is, it indicates that this is a balance
name it something like `diaplayedBalanceAccountIndex` or so.
but more importantly why are we using a different variable? isn't that the same job for `currentAccountIndex` why not just use that and remove `_balanceDisplayedForAccount`

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -338,12 +303,15 @@ abstract class ElectrumWalletBase
...
+  ObservableMap<int, ElectrumBalance> accountBalances = ObservableMap<int, ElectrumBalance>();
+
   final Bip32Slip10Secp256k1? _masterHD;
   final Bip32Slip10Secp256k1 accountHD;
   final String? _mnemonic;
 
-  late final Map<BitcoinAddressType, Bip32Slip10Secp256k1> mainHdByType;
-  late final Map<BitcoinAddressType, Bip32Slip10Secp256k1> sideHdByType;
+  final mainHdByTypeAndAccount = <int, Map<BitcoinAddressType, Bip32Slip10Secp256k1>>{};
+  final sideHdByTypeAndAccount = <int, Map<BitcoinAddressType, Bip32Slip10Secp256k1>>{};
```
OmarHatem28, line 314: Just a note no changes needed:

but if we ever add more functionality it's better to make Accounts a separate class and have it here instead of this

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -372,6 +340,155 @@ abstract class ElectrumWalletBase
...
+  int? _balanceDisplayedForAccount;
+
+  Future<List<int>> loadAccountIndexes() async {
+    if (type != WalletType.bitcoin) return [0];
+
+    final accounts = await walletInfo.getAccounts();
+    return accounts.map((account) => account.accountIndex).toList();
+  }
+
+  void _prepareHdForAccount(int accountIndex, CryptoCurrency? currency) {
```
OmarHatem28, line 355: didn't we settle on making the accounts only segwit?

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -372,6 +340,155 @@ abstract class ElectrumWalletBase
...
+      }
+    }
+  }
+
+  ElectrumBalance balanceForAccount(int accountIndex) {
+    final fromUnspents = accountBalances[accountIndex];
+    if (fromUnspents != null) return fromUnspents;
+
+    if (accountIndex == currentAccountIndex && _balanceDisplayedForAccount == accountIndex) {
+      return balance[currency] ?? _zeroBalance(currency);
```
OmarHatem28, line 387: this case should not be possible, but anyway, I guess we should at least set the `accountBalances[accountIndex] = balance[currency] ?? _zeroBalance(currency);` and then `return accountBalances[accountIndex];`

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -372,6 +340,155 @@ abstract class ElectrumWalletBase
...
+    for (final coin in unspentCoins) {
+      final accountIndex = coin.bitcoinAddressRecord.accountIndex;
+
+      final current = newBalances[accountIndex] ?? _zeroBalance(currency);
+
+      if (coin.confirmations != null && coin.confirmations! > 0) {
+        current.confirmed += Money.fromInt(coin.value, currency);
+      } else {
+        current.unconfirmed += Money.fromInt(coin.value, currency);
+      }
```
OmarHatem28, line 406: we need a way to keep track of frozen coins as well, so this also should be per account, and I guess at this point, make an account class, or see if it fits in a different class, I am still reviewing, so not sure yet

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -372,6 +340,155 @@ abstract class ElectrumWalletBase
...
+
+    for (final accountIndex in walletAddresses.accountIndexes) {
+      newBalances.putIfAbsent(accountIndex, () => _zeroBalance(currency));
+    }
+
+    accountBalances = ObservableMap<int, ElectrumBalance>.of(newBalances);
+
+    for (final accountIndex in walletAddresses.accountIndexes) {
+      final computed = newBalances[accountIndex];
+      final existing = accountBalances[accountIndex];
```
OmarHatem28, line 419: at this point, aren't they equal?

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -372,6 +340,155 @@ abstract class ElectrumWalletBase
...
+      if (computed != null && existing != null) {
+        final computedIsZero = computed.confirmed == Money.zero(currency) &&
+            computed.unconfirmed == Money.zero(currency);
+        final existingIsNonZero = existing.confirmed != Money.zero(currency) ||
+            existing.unconfirmed != Money.zero(currency);
+        if (computedIsZero && existingIsNonZero) {
+          newBalances[accountIndex] = existing;
+        }
```
OmarHatem28, line 427: why? what if the new values are after he sent the funds, so the balance is actually zero, why keep the existing, this will display a wrong balance

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -390,8 +507,67 @@ abstract class ElectrumWalletBase
...
+    }
+
+    await walletInfo.setSelectedAccount(accountIndex);
+
+    _prepareHdForAccount(accountIndex, currency);
+
+    walletAddresses.currentAccountIndex = accountIndex;
+
+    if (isNewAccount) {
+      await walletAddresses.prepareAccountAddresses(accountIndex);
```
OmarHatem28, line 529: same

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -417,6 +593,9 @@ abstract class ElectrumWalletBase
   bool _isTryingToConnect = false;
   bool? _isBatchSupported;
   DateTime? _syncBenchmarkStartTime;
+  DateTime? _syncStartedAt;
+
+  static const Duration maxSyncDuration = Duration(minutes: 3);
```
OmarHatem28, line 598:
```suggestion
  static const Duration maxSyncDuration = Duration(minutes: 2);
```

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -489,7 +668,9 @@ abstract class ElectrumWalletBase
     String? privateKey;
     String? publicKey;
 
-    final hd = mainHdByType[SegwitAddresType.p2wpkh] ?? mainHd;
+    // `keys` is wallet-level metadata
+    final hd =
+        mainHdByTypeAndAccount[0]?[SegwitAddresType.p2wpkh] ?? accountHD.childKey(Bip32KeyIndex(0));
```
OmarHatem28, line 673:
```suggestion
        mainHdByTypeAndAccount[0]?[SegwitAddresType.p2wpkh] ?? mainHd;
```

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -770,6 +976,13 @@ abstract class ElectrumWalletBase
         _syncBenchmarkStartTime = null;
         syncStatus = SyncedSyncStatus();
       }
+
+      await updateFeeRates();
+
+      _updateFeeRateTimer ??=
+          Timer.periodic(const Duration(minutes: 1), (timer) async => await updateFeeRates());
+
+      unawaited(resolveQueuedTransactionDetails());
```
OmarHatem28, line 985: you call this with `unawaited` but we are still do `await updateTransactions();` so what's the point?

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -1878,6 +2097,11 @@ abstract class ElectrumWalletBase
 
   @action
   Future<void> updateAllUnspents() async {
+    if (type == WalletType.bitcoin && !electrumClient.isConnected) {
```
OmarHatem28, line 2100: the wallet type shouldn't matter here then

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -1937,6 +2163,8 @@ abstract class ElectrumWalletBase
     }
 
     await updateCoins(unspentCoins);
+    _updateAccountBalancesFromUnspents();
+    _updateCurrentAccountBalance();
```
OmarHatem28, line 2167: I prefer that `_updateAccountBalancesFromUnspents` internally calls `_updateCurrentAccountBalance` cuz I don't see a case where we would update the accounts balances, and not need to update the current one `balance[currency]`.


another note, what I would prefer more is that we don't have 2 balances maps, but this is not to be implemented, just remind me to discuss it in our call.

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -1999,6 +2229,8 @@ abstract class ElectrumWalletBase
 
       return addresses.map((address) {
         final scriptHash = address.getScriptHash(network);
+        if (failedScriptHashes.contains(scriptHash)) return null;
+        if (!unspentByScriptHash.containsKey(scriptHash)) return null;
         return coinsByScriptHash[scriptHash] ?? <BitcoinUnspent>[];
```
OmarHatem28, line 2234: what's the difference between `unspentByScriptHash` and `coinsByScriptHash`

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -3599,8 +4129,37 @@ abstract class ElectrumWalletBase
...
+    try {
+      if (type == WalletType.bitcoin) {
+        await fetchBalances();
+
+        if (!electrumClient.isConnected) {
+          printV("updateBalance: not connected, keeping existing balances");
+          return;
+        }
+
+        _updateCurrentAccountBalance();
```
OmarHatem28, line 4141: 1. the check should be first
2. why are we no longer setting `balance[currency]` with the result of fetchBalances? is it a useless call at this point, since _updateCurrentAccountBalance, but then it's not generic, I want fetchBalances to be called regardless of the wallet, and it should handle it for the accounts we have, which would be only the primary for other currencies (LTC, Doge, BCH)

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -3599,8 +4129,37 @@ abstract class ElectrumWalletBase
...
+      final newBalance = await fetchBalances();
+      final currentBalance = balance[currency];
+      final isZeroBalance = newBalance.confirmed == Money.zero(currency) &&
+          newBalance.unconfirmed == Money.zero(currency);
+      final hadPreviousBalance = currentBalance != null &&
+          (currentBalance.confirmed != Money.zero(currency) ||
+              currentBalance.unconfirmed != Money.zero(currency));
+
+      if (!isZeroBalance || !hadPreviousBalance) {
+        balance[currency] = newBalance;
```
OmarHatem28, line 4155: same here

tool/configure.dart
```diff
@@ -322,6 +327,7 @@ abstract class Bitcoin {
 Future<void> generateMonero(bool hasImplementation) async {
   final outputFile = File(moneroOutputPath);
   const moneroCommonHeaders = """
+import 'package:cw_core/account.dart';
```
OmarHatem28, line 330: so if you already moved account class to cw_core, why not use it in Balance and generalize it so we don't have to keep 2 maps.
how is it done for monero?

cw_core/lib/wallet_info.dart
```diff
@@ -503,8 +612,94 @@ class WalletInfo {
     await WalletInfoAddress.insert(internalId, type, address);
   }
 
+  Future<List<WalletInfoAccount>> getAccounts() async {
+    final accounts = await WalletInfoAccount.selectList(internalId);
+
+    if (accounts.isEmpty) {
+      const initialAccountsCount = 1;
+
+      for (var accountIndex = 0; accountIndex < initialAccountsCount; accountIndex++) {
```
OmarHatem28, line 621: ?!

cw_core/lib/wallet_info.dart
```diff
@@ -503,8 +612,94 @@ class WalletInfo {
...
+        await WalletInfoAccount.insertOrUpdate(
+          walletInfoId: internalId,
+          accountIndex: accountIndex,
+          label: "Primary account",
+          isSelected: accountIndex == 0,
+        );
+      }
+
+      final defaultAccounts = await WalletInfoAccount.selectList(internalId);
```
OmarHatem28, line 630: repettitive, just use `accounts`

cw_core/lib/wallet_info.dart
```diff
@@ -503,8 +612,94 @@ class WalletInfo {
...
+        accountIndex: selected.accountIndex,
+      );
+      selected.isSelected = true;
+    }
+
+    selectedAccount = selected.accountIndex;
+    return accounts;
+  }
+
+  Future<void> setSelectedAccount(int accountIndex) async {
```
OmarHatem28, line 652: just add `selected.isSelected = true;` to the `setSelectedAccount` and call it, why duplicate the code

cw_bitcoin/lib/litecoin_wallet.dart
```diff
@@ -115,8 +115,10 @@ abstract class LitecoinWalletBase extends ElectrumWallet with Store {
       initialRegularAddressIndex: initialRegularAddressIndex,
       initialChangeAddressIndex: initialChangeAddressIndex,
       initialMwebAddresses: initialMwebAddresses,
-      mainHdByType: mainHdByType,
-      sideHdByType: sideHdByType,
+      mainHdByTypeAndAccount: mainHdByTypeAndAccount,
+      sideHdByTypeAndAccount: sideHdByTypeAndAccount,
+      accountIndexes: [currentAccountIndex],
+      currentAccountIndex: currentAccountIndex,
```
OmarHatem28, line 121: **Minor:**
no need to change it, but I would rather have it have a default value and not pass it for other wallets other than bitcoin

OmarHatem28 (changes requested):

cw_core/lib/db/sqlite.dart
```diff
@@ -63,11 +63,13 @@ Future<void> _initDb({String? pathOverride}) async {
     }
   }
   await db?.close();
+
   db = await openDatabase(dbFile.path, version: 10,
```
OmarHatem28, line 67: this should be pumped

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: that was just reassigning, those values are already set via super(wallet Info, derivation Info) into the parent class field.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: yeah, makes sense. I removed _balanceDisplayedForAccount. We can just use accountBalances directly here, so there’s no need to track the displayed account separately.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: I removed the fallback completely. A missing account balance now just returns _zeroBalance(currency)

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: I’ll add frozen balance tracking per account. For the account class, I’ll wait until you finish the review and then we can decide what should live there.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: you’re right. accountBalances has already been replaced with newBalances at that point, so computed and existing are the same. I removed that redundant loop.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: This was meant to protect against incomplete UTXO results, but _hasCompleteUnspentSet already handles that. I removed this logic

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: prepareAccountAddresses prepares addresses for a specific account, so I think it makes sense to keep accountIndex explicit here instead of relying on currentAccountIndex.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: updateTransactions() is still awaited because we need to fetch the address histories and populate _missingHistoryQueue before finishing the initial sync. It no longer waits for the expensive transaction detail resolution — _fetchAddressHistories only queues missing transactions

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: you’re right,  removed the wallet type check

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: I moved _updateCurrentAccountBalance() inside.
And yeah, having two balance maps definitely makes things messy.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: unspentByScriptHash is the raw response from Electrum, while coinsByScriptHash contains the parsed BitcoinUnspent objects we use in the wallet.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: Updated. The connection check is first now, and fetchBalances() is used for all Electrum wallets and updates balance[currency].

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: done

cw_core/lib/db/sqlite.dart
serhii-bor: done

cw_core/lib/wallet_info.dart
serhii-bor: this is leftover from when I was creating 3 accounts by default, removed.

cw_core/lib/wallet_info.dart
serhii-bor: same here

cw_bitcoin/lib/electrum_wallet_addresses.dart
```diff
@@ -313,7 +346,14 @@ abstract class ElectrumWalletAddressesBase extends WalletAddresses with Store {
...
 
-    final mainHd = mainHdByType[addrType] ?? mainHdByType.values.first;
+    final mainMap = walletInfo.type == WalletType.bitcoin
+        ? mainHdByTypeAndAccount[currentAccountIndex]
+        : mainHdByTypeAndAccount[0];
+
+    final mainHd = mainMap?[addrType] ?? mainMap?.values.first;
+
+    if (mainHd == null) return '';
+
```
rafael-xmr, line 356: Idea: a common function for the re-usable logic in these? edit: for both ` get primaryAddress`  and ` get payjoinCompatibleAddress` when getting the mainHd

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -3006,35 +3273,233 @@ abstract class ElectrumWalletBase
...
+        await walletInfo.addAccount(
+          accountIndex: accountIndex,
+          label: "Account $accountIndex",
+        );
+      }
+      // Move on to the next account index
+      accountIndex++;
+    }
+    if (electrumClient.isConnected) {
+      await walletInfo.setAccountDiscoveryLimit(maxProbAccounts);
```
rafael-xmr, line 3373: if `accountDiscoveryLimit` is only ever set here, and it is always set as the static `maxProbAccounts` can't the uses of `accountDiscoveryLimit` be removed to just use `maxProbAccounts` directly?

cw_bitcoin/lib/electrum_wallet_addresses.dart
```diff
@@ -366,21 +410,24 @@ abstract class ElectrumWalletAddressesBase extends WalletAddresses with Store {
...
+          // Non need to generate legacy addresses for other accounts than the first one, since accounts were added later.
+          if (accountIndex == 0) {
+            await _generateInitialAddresses(
+              accountIndex: accountIndex,
+              type: type,
+              isLegacyDerivation: true,
+            );
+          }
+        }
```
rafael-xmr, line 430: Could use ` prepareAccountAddresses`  here?

lib/new-ui/widgets/coins_page/cards/cards_view.dart
```diff
@@ -103,6 +189,103 @@ class _CardsViewState extends State<CardsView> {
...
+          final walletBalanceRecord = widget.dashboardViewModel.balanceViewModel
+              .getMainBalanceRecord(widget.lightningMode);
+
+          late final String walletBalance;
+          late final String walletFiatBalance;
+          if (widget.dashboardViewModel.mwebEnabled && widget.dashboardViewModel.hasMweb) {
+            if (widget.dashboardViewModel.balanceViewModel.displayMode ==
+                BalanceDisplayMode.hiddenBalance) {
+              walletBalance = '●●●●●●';
+              walletFiatBalance = '●●●●●●';
```
rafael-xmr, line 229: wouldn't it be better to use a const string like ` hidden_value`  so the string can be re-used consistently? this same string appears more times across the codebase but very minor

lib/src/screens/settings/wallet_accounts_page.dart
```diff
@@ -0,0 +1,514 @@
...
+    required this.order,
+    required this.accountListItem,
+  });
+
+  final BalanceCard card;
+  final int order;
+  final AccountListItem accountListItem;
+}
+
+class WalletAccountsPage extends StatefulWidget {
```
malik1004x, line 41: This class seems to just be `AccountCustomizer` – I don't think there's much advantage in moving it here compared to just leaving it in `new-ui` – this is a New UI page after all, and this will cause conflicts with the new Accounts UI PR. Could you consider moving it back and changing the name back?

lib/src/screens/settings/wallet_accounts_page.dart
```diff
@@ -0,0 +1,514 @@
...
+                              TextStyle(color: Theme.of(context).colorScheme.onSurfaceVariant),
+                        ),
+                      ),
+                      Expanded(
+                        child: Stack(
+                          children: [
+                            SingleChildScrollView(
+                              controller: ModalScrollController.of(context),
+                              padding: const EdgeInsets.only(bottom: 196),
+                              child: CardsView(
```
malik1004x, line 240: Using `CardsView` to show the cards in the customization screen causes a significant change to the UI (verified this on a build) – previously, cards shown in `AccountCustomizer` were all rendered in the same size, and with this change, background accounts get smaller than the foreground ones. This means that if you have a large number of accounts, the background cards get really tiny and hard to click or grab.

Image


Back when we were doing closed testing for 6.0, I got several complaints of the same problem on the home screen – this is why I introduced compact mode for `CardsView`. Here you disable it, which of course is necessary to show all cards properly, but also means UI suffers. If you decided to switch this page to `CardsView` because of the bugs in card ordering, it should be noted those bugs are fixed in the Accounts UI PR along with the flow being substantially more stable in general, thefore this change might be unnecessary. If there's some other reason you switched this to `CardsView` rather than instantiating `BalanceCard` directly, you can consider making the internal card size calculation always return the same multiplier if `allowCompactMode` is `false` – this would result in a UI very similar to the previous one.



Edit: I spoke to @OmarHatem28 about this and we think that this workaround should be reverted – if this change was done to prevent bugs with ordering, keep the current production UI for this and just add the changes necessary for the accounts feature you're adding (most likely just the "enable accounts" toggle). It won't be more buggy than it currently is, and the new Accounts UI PR will fix the bugs very soon.

lib/new-ui/utils/show_card_customizer.dart
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  if (useCupertinoScaffold) {
+    await CupertinoScaffold.showCupertinoModalBottomSheet(
+      barrierColor: Colors.black.withAlpha(60),
+      context: context,
+      builder: buildCustomizer,
+    );
+  } else {
+    await showCupertinoModalBottomSheet(
```
malik1004x, line 46: I'm not sure I understand why this is now a modal sheet – previously, opening `CardCustomizer` from `AccountCustomizer` would open it as a page, same as any other page within settings. Using `showCupertinoModalBottomSheet` here means we're opening a moda sheet _inside_ a modal sheet, which makes for some quite confusing animations. You can always `await` a `Navigator.push` call – the Future returned by that is completed when the page is popped.

lib/new-ui/utils/show_card_customizer.dart
```diff
@@ -0,0 +1,64 @@
+import 'package:cake_wallet/di.dart';
+import 'package:cake_wallet/new-ui/modal_navigator.dart';
+import 'package:cake_wallet/new-ui/pages/card_customizer.dart';
+import 'package:cake_wallet/new-ui/viewmodels/card_customizer/card_customizer_bloc.dart';
+import 'package:cake_wallet/view_model/dashboard/dashboard_view_model.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:modal_bottom_sheet/modal_bottom_sheet.dart';
+
+Future<void> showCardCustomizer({
```
malik1004x, line 10: [minor] Instead of a loose function, have you considered making this a static method of `CardCustomizer`?

lib/src/screens/settings/widgets/account_creation_modal.dart
```diff
@@ -0,0 +1,133 @@
...
+import 'package:cw_core/generate_name.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+class AccountCreationModal extends StatefulWidget {
+  const AccountCreationModal({
+    required this.onPressed, required this.state, super.key,
+  });
+
+  final Future<void> Function(String label) onPressed;
```
malik1004x, line 16: You seem to have refactored this widget to not take in the ViewModel – I don't think I quite understand the purpose of this change, it's not clear what `onPressed` or `state` do and it means may be harder to instantiate this modal from anywhere else if we need that. Could you consider passing the ViewModel here like before?

lib/view_model/wallet_account_list/bitcoin_account_list/bitcoin_account_list_view_model.dart
```diff
@@ -0,0 +1,130 @@
...
+import 'package:cake_wallet/view_model/wallet_account_list/wallet_account_list_view_model.dart';
+import 'package:collection/collection.dart';
+import 'package:cw_core/crypto_currency.dart';
+import 'package:cw_core/wallet_base.dart';
+import 'package:cw_core/wallet_info.dart';
+import 'package:mobx/mobx.dart';
+
+part 'bitcoin_account_list_view_model.g.dart';
+
+class BitcoinAccountListViewModel = BitcoinAccountListViewModelBase
```
malik1004x, line 17: Leaving this comment here, but this applies to the PR in general.

You're creating a separate implementation of this ViewModel for Bitcoin specifically – this means there's now two independent implementations. If we implement accounts this way for EVM chains, then for Litecoin and Dogecoin, we now have (counting Monero) 5 different implementations of the same ViewModel. The potential issues this approach entails are already visible across this PR – you have to instantiate this ViewModel in a different way depending on wallet types, you do wallet type checks to call different variants of the same function and some important logic is delegated into the wallet integration classes (`CWBitcoin`, `CWMonero`) instead of calling the wallet directly. This was okay as long as Monero was the only wallet type that had account support, but I don't think it would scale well considering we're planning to introduce accounts to more wallet types soon.

With this in mind, I'd like to propose an alternate approach. Consider creating `mixin AccountWallet for WalletBase`, then adding `with AccountWallet` to `BitcoinWallet` and `MoneroWallet` classes. This mixin would have functions like `int get currentAccountIndex`, `void setCurrentAccount()`, etc that would be overridden with the wallet-specific implementations for Bitcoin, Monero and any other chains that we'd want accounts for in the future. This way, the app-side code can call, for example, `wallet.setAccountName("something")` and not worry about what type `wallet` actually is – on UI PRs we can trust that the wallet-side implementation will handle the task, and PRs that add account support to more chains don't have to go through UI and add `if(wallet.type == WalletType.something)` everywhere.

The slight disadvantage is you'd also need to do some minor adjustments in Monero code, but I believe even with that it would mean less complexity in the PR than copying all logic for Monero and Bitcoin.

I should note that this comment is based on the knowledge that Wownero is gone from the app anyway and Nano accounts aren't functional – I'm aware they have their implementations of this feature as well, but since they are de-facto dead code I don't think we should base structural decisions on them.

If you'd like to see an example, I used a similar (although simpler) pattern in the Coin Control refactor, where supported wallets have a mixin `CoinControlWallet` and implement their own callbacks for each of the Coin Control features. This way, I avoided wallet-specific code in the ViewModel and UI completely – a 100% wallet-independent version may not be possible in this case, but I think it's worth trying to make it somewhat genericized so we don't spend too much time adding account support to more chains in the future.

lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -364,8 +361,17 @@ abstract class DashboardViewModelBase with Store {
 
     transactions.clear();
 
+    final allTransactions = wallet.transactionHistory.transactions.values;
+    final filteredTransactions = allTransactions.where((tx) {
```
malik1004x, line 365: I believe you're already doing this filtering below, as part of `bitcoin!.isTransactionForCurrentAccount`.

lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -427,60 +433,63 @@ abstract class DashboardViewModelBase with Store {
...
+
+    final orderableCount = (wallet.type == WalletType.bitcoin) ? numAccounts - 1 : numAccounts;
 
     for (int i = 0; i < numAccounts; i++) {
       late final int index;
-      if (balanceViewModel.hasAccounts) {
+      final isLightning = wallet.type == WalletType.bitcoin && i == lightningCardIndex;
+      if (isLightning) {
+        index =
+            -2; // using -2 to differentiate lightning card from regular accounts, which use 0, 1, 2, etc.
```
malik1004x, line 461: You use -2 for Lightning, I believe we previously did 0 for Bitcoin and 1 for Lightning – this is the correct approach, but I haven't seen any migration code to move the index for old Lightning designs. Wouldn't this mean that:
- on update, designs for Lightning cards will be reset for all users,
- when a user creates a new Bitcoin account, that account will have the design of their pre-update Lightning card?

lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -602,6 +614,12 @@ abstract class DashboardViewModelBase with Store {
   @observable
   ObservableMap<int, int> cardOrder;
 
+  @observable
+  bool lightningMode = false;
```
malik1004x, line 618: You have `lightningMode` here and also `_lightningMode` in `NewHomePage` – this means they have to be kept in sync and updated independently.

I would prefer `lightningMode` to stay as a purely UI thing – the ViewModel can always deliver both data related and unrelated to Lightning, and we can filter it out in `CardsView` or other UI elements however needed.

If there's some reason this _has to_ be inside the ViewModel, consider making `_lightningMode` in `NewHomePage` a getter based on this field instead of its own variable that could potentially desync.

lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -1381,6 +1412,21 @@ abstract class DashboardViewModelBase with Store {
     }, _transactionDisposerCallback, delay: 300);
   }
 
+  void _setupBitcoinAccountChangeReaction(WalletBase wallet) {
+    if (wallet.type != WalletType.bitcoin) {
+      return;
+    }
+
+    _onBitcoinAccountChangeReaction?.reaction.dispose();
+    _onBitcoinAccountChangeReaction = reaction(
```
malik1004x, line 1421: Why is `_onBitcoinAccountChangeReaction` separate from `_onMoneroAccountChangeReaction`? From what I understand, they can't be non-null at the same time – this means you could rename the latter to `_onAccountChangeReaction` and use it for both Bitcoin and Monero, eliminating some more wallet-specific logic.

lib/view_model/dashboard/balance_view_model.dart
```diff
@@ -159,7 +159,7 @@ abstract class BalanceViewModelBase with Store {
   bool get isEVMCompatible => isEVMCompatibleChain(wallet.type);
 
   @computed
-  bool get hasAccounts => [WalletType.monero, WalletType.wownero].contains(wallet.type);
+  bool get hasAccounts => hasAccountsWalletTypes.contains(wallet.type);
```
malik1004x, line 162: In relation to my previous comment about the wallet-side `AccountWallet` mixin – getters like these and the `hasAccountsWalletTypes` constant would be essentially rendered irrelevant by this, since you could just do `wallet is AccountWallet`.

lib/di.dart
```diff
@@ -578,6 +580,13 @@ Future<void> setup({
   getIt.registerFactory(() => DashboardViewModel(
       tradeMonitor: getIt.get<TradeMonitor>(),
       balanceViewModel: getIt.get<BalanceViewModel>(),
+      accountListViewModelFactory: () {
```
malik1004x, line 583: Couldn't we avoid this by calling `getIt.get<WalletAccountListViewModel>()` inside the DashboardViewModel?

lib/di.dart
```diff
@@ -1152,6 +1158,31 @@ Future<void> setup({
   getIt.registerFactory(() => SecurityBackupPage(getIt.get<SecuritySettingsViewModel>(),
       getIt.get<AuthService>(), getIt.get<AppStore>().wallet!.isHardwareWallet));
 
+  getIt.registerFactory(() {
```
malik1004x, line 1161: Since `getIt.get<WalletAccountListViewModel>()` already filters for the proper wallet type, I don't think it's necessary to create per-type options in this factory.

lib/router.dart
```diff
@@ -970,6 +957,9 @@ Route<dynamic> createRoute(RouteSettings settings) {
     case Routes.manageNodes:
       return MaterialPageRoute<void>(builder: (_) => getIt.get<ManageNodesPage>(param1: false));
 
+    case Routes.walletAccountsPage:
```
malik1004x, line 960: You seem to have replaced `handleRouteWithPlatformAwareness` with `MaterialPageRoute` – are you sure this wouldn't break the swipe-back gesture? Remember that we currently intend to have the iOS-style swipe gestures work on _both_ iOS and Android – it's the reason `ModalNavigator` has a special wrapper with overrides for all platforms. This also means that using `CupertinoPageRoute` is generally better.

cw_bitcoin/lib/bitcoin_wallet.dart
```diff
@@ -724,6 +732,16 @@ abstract class BitcoinWalletBase extends ElectrumWallet with Store {
       return hasSilentPaymentsScanning;
     }
 
+    // Restrict extra Bitcoin accounts (accountIndex > 0) to the address types
+    if (type == WalletType.bitcoin && currentAccountIndex > 0) {
```
malik1004x, line 736: [minor] Why check `type == WalletType.bitcoin` inside `BitcoinWalletBase`?

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -92,66 +93,24 @@ abstract class ElectrumWalletBase
...
-      }
-    } else {
-      // View-only wallet (xpub only)
-      for (final type in supportedTypes) {
-        mainHdByType[type] = mainHd;
-        sideHdByType[type] = sideHd;
-      }
-    }
+    // Probably not needed as init will be called right after creating the wallet.
+    _prepareHdForAccount(currentAccountIndex, currency);
```
malik1004x, line 113: If it's "probably not needed", could you consider doing some tests after commenting out this call and dropping the function if it's still functional?

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -317,16 +275,22 @@ abstract class ElectrumWalletBase
...
   static const int transactionBatchTimeoutMs = 15000;
-
   static const int batchTestTimeoutMs = 4000;
   static const int batchTestHashesCount = 2;
-
   static const bool useBatchForHistory = true;
 
+  static const int maxProbAccounts = 3;
+
+  static ElectrumBalance _zeroBalance(CryptoCurrency currency) => ElectrumBalance(
```
malik1004x, line 286: [minor] This would be better as a constructor inside `ElectrumBalance` – `ElectrumBalance.zero`.

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -338,12 +302,15 @@ abstract class ElectrumWalletBase
...
+  @observable
+  ObservableMap<int, ElectrumBalance> accountBalances = ObservableMap<int, ElectrumBalance>();
+
   final Bip32Slip10Secp256k1? _masterHD;
   final Bip32Slip10Secp256k1 accountHD;
   final String? _mnemonic;
 
-  late final Map<BitcoinAddressType, Bip32Slip10Secp256k1> mainHdByType;
-  late final Map<BitcoinAddressType, Bip32Slip10Secp256k1> sideHdByType;
+  final mainHdByTypeAndAccount = <int, Map<BitcoinAddressType, Bip32Slip10Secp256k1>>{};
```
malik1004x, line 312: I don't think it's a good idea to have a nested `Map` – it's not immediately clear what the keys are (especially since one of them is a plain `int`) and this could lead to some messy logic.

A `Map` is the right choice here in general since it allows O(1) lookups – however, instead of nesting maps you could consider simplifying this by making a class `HDKey` with fields `int accountIndex` and `BitcoinAddressType addressType` – then using that to key the map. This way, we avoid nested maps and have potential for cleaner, more easily understandable logic across the board.

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -372,6 +339,146 @@ abstract class ElectrumWalletBase
   @observable
   SyncStatus syncStatus;
 
+  bool _hasCompleteUnspentSet = false;
+  bool _isSyncing = false;
+
+  Future<WalletInfoAccount> loadCurrentAccount() async {
+    final accounts = await walletInfo.getAccounts();
```
malik1004x, line 346: Since `walletInfo.getAccounts` is a database operation, this means we're getting all rows from the database and then filtering them out wallet-side. Have you considered adding another getter like `getSelectedAccount` and doing an SQL `WHERE`? Probably not a huge performance hit, but worth considering in case someone wants to create an obscene number of accounts.

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -372,6 +339,146 @@ abstract class ElectrumWalletBase
...
+    final current = accounts.firstWhere(
+          (account) => account.isSelected,
+      orElse: () => accounts.first,
+    );
+
+    return current;
+  }
+
+  Future<List<int>> loadAccountIndexes() async {
+    if (type != WalletType.bitcoin) return [0];
```
malik1004x, line 356: [minor] Instead of checking the wallet type here, you could override this in `BitcoinWallet` with the proper logic and just keep this function always returning `[0]`.

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -372,6 +339,146 @@ abstract class ElectrumWalletBase
...
+  }
+
+  void _prepareHdForAccount(int accountIndex, CryptoCurrency? currency) {
+    // For the first account, we support all address types. For extra accounts, we only support SegwitAddresType.p2wpkh.
+    final supportedTypes =
+    accountIndex == 0 ? supportedAddressTypes(walletInfo.type) : EXTRA_ACCOUNT_ADDRESS_TYPES;
+    final isElectrumDerivation = derivationInfo.derivationType == DerivationType.electrum;
+    final canDeriveFromSeed = _masterHD != null && currency != null;
+
+    final mainMap = mainHdByTypeAndAccount[accountIndex] ??= {};
```
malik1004x, line 369: [minor] Not sure how Dart's type inference will behave here – please make sure it doesn't fall back to `Map<dynamic, dynamic>` if no type is provided.

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -372,6 +339,146 @@ abstract class ElectrumWalletBase
...
+      // case for Electrum derivation or when we don't have seed but only xpub (view only wallet)
+      for (final type in supportedTypes) {
+        mainMap[type] = mainHd; // accountHD.child(0)
+        sideMap[type] = sideHd; // accountHD.child(1)
+      }
+    }
+  }
+
+  ElectrumBalance balanceForAccount(int accountIndex) =>
+      accountBalances[accountIndex] ?? _zeroBalance(currency);
```
malik1004x, line 392: Since we have this getter, is it necessary to expose the map? If it's also used in `BitcoinWallet`, you could add `@protected` to have the linter warn you when using it outside class methods.

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -523,16 +705,29 @@ abstract class ElectrumWalletBase
...
+
   void Function(FlutterErrorDetails)? _onError;
   Timer? _autoSaveTimer;
   StreamSubscription<dynamic>? _receiveStream;
   Timer? _updateFeeRateTimer;
   static const int _autoSaveInterval = 1;
 
   Future<void> init() async {
-    await walletAddresses.init();
+    final accountIndexes = await loadAccountIndexes();
```
malik1004x, line 721: Just to make sure I understand this feature correctly: does this mean you can have accounts 0 and 2 but not account 1? How would that happen UI-wise? If it's impossible, this could probably be replace with `walletInfo.getAccounts().length`.

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -2633,293 +2917,190 @@ abstract class ElectrumWalletBase
...
+    return future.whenComplete(() => _resolveMissingHistoryFuture = null);
+  }
 
-              _applyLitecoinPegOutTag(tx);
-              transactionHistory.addOne(tx);
-              await transactionHistory.save();
-            }
-          }
+  Future<void> _drainMissingHistoryQueue() async {
+    final watch = Stopwatch()..start();
```
malik1004x, line 3017: I assume the stopwatch was left over from testing, could you remove it before we merge?

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -3603,8 +4102,25 @@ abstract class ElectrumWalletBase
...
+    if (!electrumClient.isConnected) {
+      printV("updateBalance: not connected, keeping existing balance");
+      return;
+    }
+
+    try {
+      balance[currency] = await fetchBalances();
+
+      // For Bitcoin wallets, we also update the current account balance after fetching the overall balance.
+      // because fetchBalances() only updates the overall wallet balance, not the individual account balances.
```
malik1004x, line 4115: If `fetchBalances` updates the overall wallet balance, wouldn't this mean that in the time between the above assignment running and `_updateCurrentAccountBalance` finishing, `balance[currency]` holds the total balance of all accounts and not the balance of the current account? I believe this could lead to some UI bugs and, in rare cases where `_updateCurrentAccountBalance` takes a while, spending of coins from different accounts. Please correct me if wrong, though as I'm not 100% sure that's how the logic works.

cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -4566,3 +5080,27 @@ class UtxoDetails {
...
   });
 }
+
+class AddressHistoryScan {
+  AddressHistoryScan({
+    required this.historyByScriptHash,
+    required this.hadFailures,
+  });
+
+  final Map<String, List<Map<String, dynamic>>> historyByScriptHash;
```
malik1004x, line 5090: There's a lot of nested types here – have you considered making another helper class for this for better readability?

cw_bitcoin/lib/electrum_wallet_addresses.dart
```diff
@@ -199,8 +216,16 @@ abstract class ElectrumWalletAddressesBase extends WalletAddresses with Store {
           "Error: Unable to fetch your Lightning address, please check your network connection.";
     }
 
-    final typeMatchingAddressesAll =
-        _addresses.where((addr) => !addr.isHidden && _isAddressPageTypeMatch(addr)).toList();
+    // Should not happen, but just in case, return an empty string if the address type is not supported.
```
malik1004x, line 219: If it "should not happen", please throw an exception – this way we get an error that can be properly reported and handled and not just unexpected behavior.

cw_bitcoin/lib/electrum_wallet_addresses.dart
```diff
@@ -440,6 +504,21 @@ abstract class ElectrumWalletAddressesBase extends WalletAddresses with Store {
     return labels;
   }
 
+  Future<void> prepareAccountAddresses(int accountIndex, {List<BitcoinAddressType>? types}) async {
```
malik1004x, line 507: [minor] Wouldn't it be better to have a default parameter value instead of using `??`?

cw_core/lib/db/sqlite.dart
```diff
@@ -164,6 +164,21 @@ CREATE TABLE IF NOT EXISTS BalanceCardStyleSettings (
...
+        db,
+        table: "WalletInfo",
+        column: "accountDiscoveryLimit",
+        definition: "INTEGER DEFAULT NULL",
+      );
+      await _addColumnIfNotExists(
+        db,
+        table: "WalletInfo",
+        column: "isMultiAccountsEnabled",
+        definition: "INTEGER DEFAULT NULL",
```
malik1004x, line 179: [minor] Why would it be `DEFAULT NULL`? Shouldn't we provide a default value for this?

cw_core/lib/db/sqlite.dart
```diff
@@ -330,6 +349,25 @@ ON Trade (id);
...
+Future<void> _createWalletInfoAccountTable(Database db) async {
+  await db.execute('''
+CREATE TABLE IF NOT EXISTS WalletInfoAccount (
+  walletInfoAccountId INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+  walletInfoId INTEGER NOT NULL,
+  accountIndex INTEGER NOT NULL,
+  label TEXT NOT NULL,
+  isSelected INTEGER DEFAULT 0 NOT NULL,
+  CONSTRAINT WalletInfoAccount_WalletInfo_FK FOREIGN KEY (walletInfoId) REFERENCES WalletInfo(walletInfoId),
+  UNIQUE(walletInfoId, accountIndex)
```
malik1004x, line 367: If this is `UNIQUE` why not make it the primary key? Then you could reliably get one account based on the IDs alone.

cw_core/lib/account.dart
```diff
@@ -1,12 +1,36 @@
...
-        this.balance = (map['balance'] ?? '0.00') as String;
+  Account.fromMap(Map<String, Object?> map)
+      : id = map['id'] is int ? map['id'] as int : int.tryParse(map['id']?.toString() ?? '') ?? 0,
+        label = (map['label'] ?? '') as String,
+        balance = (map['balance'] ?? '0.00') as String;
 
   final int id;
   final String label;
-  final String? balance;
+  final String balance;
```
malik1004x, line 15: Shouldn't `balance` be a `Money` object?

lib/new-ui/widgets/coins_page/cards/cards_view.dart
```diff
@@ -113,130 +296,71 @@ class _CardsViewState extends State<CardsView> {
...
-                fiatBalance: walletFiatBalance,
-                selected: _selectedIndex == visualIndex,
-                onCustomizeTapped: _selectedIndex == visualIndex ? widget.onCustomizeTapped : null,
-                design: cardDesign,
-                actions: actions,
-              );
-            }),
-          ),
-        ),
+        child: widget.enableReorder && !widget.lightningMode && !compactMode
```
malik1004x, line 299: On the current builds, enabling Lightning mode makes the cards view awkwardly jump instead of smoothly animating into the Lightning card – could you use `AnimatedSwitcher` keyed on `lightningMode` so we don't lose the nice animation?

lib/view_model/wallet_account_list/bitcoin_account_list/bitcoin_account_list_view_model.dart
```diff
@@ -0,0 +1,130 @@
...
+
+
+  @action
+  Future<void> _recomputeBalanceStrings() {
+    final cached = _cachedStoredAccounts;
+    if (cached == null) return _loadAccounts();
+    return _runLoad(() async => cached);
+  }
+
+  Future<void> _runLoad(Future<List<WalletInfoAccount>> Function() fetch) {
```
malik1004x, line 82: I assume this is intended to allow waiting until the accounts are loaded, but I believe it's quite overkill since the database is very fast.

lib/new-ui/pages/home_page.dart
```diff
@@ -278,54 +323,118 @@ class _NewHomePageState extends State<NewHomePage> with RouteAware {
   }
 
   Future<void> openAccountCustomizer() async {
-    final accountList = accountListViewModel;
-    if (accountList == null) {
+    final accountList = widget.dashboardViewModel.accountListViewModel;
+    final accountEditOrCreate = accountEditOrCreateViewModel;
```
malik1004x, line 327: You appear to be keeping `accountEditOrCreateViewModel` in `NewHomePage` and doing special logic to reload it every time just to initiate `WalletAccountsPage` – please consider putting that in dependency injection and dropping the home page's dependency on this ViewModel.

malik1004x (changes requested):
In addition to my comments, please make sure all newly added files and lines use double quotes.

On the proposal to genericize the account creation logic – I would recommend a broader discussion on how the final shape of the `AccountWallet` class would look like, since you understand this flow now and can share some insight into how we could structure the class.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: done

cw_core/lib/wallet_info.dart
serhii-bor: The selected.isSelected field was removed completely. A better approach is to use currentAccountIndex as a single source of truth in WalletInfo.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: no, it's a "done or not" flag, not just a repeated number. If you swap it for maxProbAccounts directly, the check becomes maxProbAccounts < maxProbAccounts, which is always false — so the probe would never run again for any wallet.

cw_bitcoin/lib/electrum_wallet_addresses.dart
serhii-bor: yes, it can be improved.

cw_bitcoin/lib/electrum_wallet_addresses.dart
serhii-bor: it can be improved

lib/new-ui/widgets/coins_page/cards/cards_view.dart
serhii-bor: Makes sense. I’ll update it during the balance class refactor.

cw_core/lib/wallet_type.dart
```diff
@@ -41,6 +41,8 @@ const evmWalletTypes = [
   WalletType.bsc
 ];
 
+const hasAccountsWalletTypes = [WalletType.monero, WalletType.wownero, WalletType.bitcoin];
```
konstantinullrich, line 44: You added this feature to Dogecoin, Litecoin, and Bitcoin Cash too. Or is that just for compatibility?

lib/src/screens/settings/wallet_accounts_page.dart
serhii-bor: yes, it's essentially AccountCustomizer, I'm okay moving it back, but I'd like to keep the current name, since it does more than just customize accounts.

lib/src/screens/settings/wallet_accounts_page.dart
serhii-bor: reverted UI changes

lib/new-ui/utils/show_card_customizer.dart
serhii-bor: fixed, added an asModalSheet flag, home page still opens as a modal sheet, WalletAccountsPage now uses Navigator.push instead

lib/new-ui/utils/show_card_customizer.dart
serhii-bor: Left as is for future improvements

lib/src/screens/settings/widgets/account_creation_modal.dart
serhii-bor: reverted, AccountCreationModal now takes the WalletAccountEditOrCreateViewModel directly

lib/view_model/wallet_account_list/bitcoin_account_list/bitcoin_account_list_view_model.dart
serhii-bor: appreciate the detailed feedback. You’re right, this is the obvious next step in the refactoring.

lib/view_model/dashboard/dashboard_view_model.dart
serhii-bor: removed the redundant check, isTransactionForCurrentAccount already handles it internally.

lib/view_model/dashboard/dashboard_view_model.dart
serhii-bor: migration step added

lib/view_model/dashboard/dashboard_view_model.dart
serhii-bor: fixed

lib/view_model/dashboard/dashboard_view_model.dart
serhii-bor: fixed

lib/view_model/dashboard/balance_view_model.dart
serhii-bor: ok

lib/di.dart
serhii-bor: no, not in this case, I'd prefer to keep passing dependencies via constructor, and keep this decision logic in the DI setup.

lib/di.dart
serhii-bor: leaving as-is so far, accountEditOrCreateViewModel doesn't have a shared type that auto-picks the right implementation like WalletAccountListViewModel does, so the switch is still needed to pick between Bitcoin and Monero versions directly.

lib/router.dart
serhii-bor: Restored handleRouteWithPlatformAwareness usage

cw_bitcoin/lib/bitcoin_wallet.dart
serhii-bor: fixed

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: leaving as-is for now, for follow-up investigation

lib/new-ui/widgets/coins_page/cards/cards_view.dart
konstantinullrich: If you use Money for `walletBalance` and `walletFiatBalance` and a MoneyText to display these values it will automatically honor the users BalanceDisplayMode setting.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: fixed

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: It will be part of further refactoring

cw_core/lib/account.dart
serhii-bor: it will be part of further refactoring

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: I'd rather leave this as-is for now

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: possible further refactor, I'll take it into account.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: Dart correctly infers the type.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: yes, It needs to stay public

lib/view_model/wallet_account_list/bitcoin_account_list/bitcoin_account_list_view_model.dart
```diff
@@ -0,0 +1,131 @@
...
+
+  String _balanceForAccount(int accountIndex) {
+    if (settingsStore.balanceDisplayMode == BalanceDisplayMode.hiddenBalance) {
+      return '●●●●●●';
+    }
+    final balance = bitcoin!.balanceForAccount(_wallet, accountIndex);
+
+    return AmountParsingProxy(settingsStore.displayAmountsInSatoshi)
+        .asDisplayStringWithSymbol(balance.available);
+  }
```
konstantinullrich, line 129: Why not use `Money` here?

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: leaving as-is. There's no case today where indexes have a hole in the middle, but the suggested simplification would create a possible inconsistency if we ever add an account-deletion feature for example, since it assumes indexes stay contiguous.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: It's not leftover,  it's intentional logging for the quick-sync flow

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: There's no actual race condition, there's no await between the assignment and _updateCurrentAccountBalance(), both run synchronously once fetchBalances() resolves. Also, coin selection for spending doesn't read balance[currency] at all, so it doesn't affect spending.

cw_bitcoin/lib/electrum_wallet.dart
serhii-bor: We'll consider it for a further refactor

cw_bitcoin/lib/electrum_wallet_addresses.dart
serhii-bor: throwing here isn't a good  option,  this value is read via MobX across the app, and an exception in a @computed getter could crash whatever widget/page depends on it. I'll add a printV here instead, just in case.

cw_bitcoin/lib/electrum_wallet_addresses.dart
serhii-bor: changed to default parameter

cw_core/lib/db/sqlite.dart
serhii-bor: I'd prefer to keep it nullable, that lets us set a different default per wallet type. We're also planning to add accounts to EVM, and we'll need to decide separately whether that's enabled or disabled by default

cw_core/lib/db/sqlite.dart
serhii-bor: make sense, switched to PRIMARY KEY (walletInfoId, accountIndex) and dropped the walletInfoAccountId entirely.

lib/new-ui/widgets/coins_page/cards/cards_view.dart
serhii-bor: fixed

lib/view_model/wallet_account_list/bitcoin_account_list/bitcoin_account_list_view_model.dart
serhii-bor: We need to call it in sequence because there can be a race condition. If two triggers happen almost at the same time, fetch() can finish in a different order. Without this check, the older fetch can finish last and overwrite the newer data. Leave it as is until we find a better way to fix it.

lib/new-ui/pages/home_page.dart
serhii-bor: good catch, it looks like a leftover from refactoring. Wallet type selection is already handled by ViewModel factories. I’ve removed it.

lib/router.dart
```diff
@@ -853,6 +846,14 @@ Route<dynamic> createRoute(RouteSettings settings) {
...
+      final dashboardViewModel = settings.arguments as DashboardViewModel?;
+      assert(dashboardViewModel != null,
```
malik1004x, line 851: Since `assert` only triggers in debug mode, this is a crash in debug mode and incorrect behavior in release mode. Since it's not going to work properly anyway, please just let it throw so when it happens in a release we at least get an error we can debug.

lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -243,7 +244,10 @@ abstract class DashboardViewModelBase with Store {
       });
     }
 
-    loadCardDesigns();
+    unawaited(() async {
+      await _migrateLegacyLightningCardDesign();
```
malik1004x, line 248: Why not add it as a regular migration step?

lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -362,8 +366,14 @@ abstract class DashboardViewModelBase with Store {
 
     transactions.clear();
 
+    final allTransactions = wallet.transactionHistory.transactions.values;
+    final filteredTransactions = allTransactions
+        .where((tx) => bitcoin!.isTransactionForCurrentAccount(wallet, tx))
```
malik1004x, line 371: This filtering is done wallet-side for Monero, can't you do the same for Bitcoin?

cw_core/lib/wallet_info.dart
```diff
@@ -244,6 +244,74 @@ class WalletInfoAddress {
...
+      final updated = await txn.update(
+        tableName,
+        {'label': label},
+        where: 'walletInfoId = ? AND accountIndex = ?',
+        whereArgs: [walletInfoId, accountIndex],
+      );
+
+      if (updated > 0) return updated;
+
+      return await txn.insert(tableName, {
```
malik1004x, line 290: Why not just do a simple `insert` with `ConflictAlgorithm.replace`?

lib/new-ui/widgets/coins_page/cards/cards_view.dart
```diff
@@ -1,46 +1,44 @@
 import 'dart:math';
```
malik1004x, line 1: I believe we agreed to revert any UI changes, but this file has a lot of changes still – are they all necessary for making this flow work?

lib/new-ui/pages/wallet_accounts_page.dart
```diff
@@ -0,0 +1,556 @@
...
+          const SizedBox(height: 24),
+          if (showAccountsToggle)
+            Padding(
+              padding: const EdgeInsets.symmetric(horizontal: 24),
+              child: NewListSections(
+                sections: {
+                  "": [
+                    ListItemToggle(
+                        keyValue: "Multiple Accounts",
+                        label: "Multiple Accounts",
```
malik1004x, line 218: Shouldn't this be translated?

malik1004x (changes requested):
Code style choices left as-is in the interest of shipping the feature with recommendation of future refactor.

In addition to comments below, please make sure new lines are formatted with double quotes.

cw_core/lib/wallet_type.dart
serhii-bor: no, currently accounts are supported only by these three types of wallets. I’ve just moved it here from the Balance ViewModel.

lib/view_model/wallet_account_list/bitcoin_account_list/bitcoin_account_list_view_model.dart
serhii-bor: walletBalance = '●●●●●●';
walletFiatBalance = '●●●●●●';

Agreed, it should be refactored. But it needs changes across the whole balance flow, so I'd rather do it in a separate PR as part of the balance refactor.


String _balanceForAccount(int accountIndex) {
Same here, AccountListItem.balance is a shared String (Monero uses it too)

cw_core/lib/wallet_info.dart
serhii-bor: I think this approach is a bit more reliable. With replace, if we add another column to this table later, renaming an account would reset that column to its default value.

PR #3450, opened 2026-07-28
Migrate EVM/Sol/TRX tokens to SQL

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


cw_core/lib/erc20_token.dart
```diff
@@ -47,32 +26,161 @@ class Erc20Token extends CryptoCurrency with HiveObjectMixin {
...
+        iconPath = icon ?? other.iconPath,
+        isPotentialScam = other.isPotentialScam,
+        id = 0,
+        walletName = walletName ?? other.walletName,
+        chainId = chainId ?? other.chainId,
         super(
           name: other.name,
           title: other.symbol.toUpperCase(),
           fullName: other.name,
-          tag: tag,
```
OmarHatem28: why is this removed?

cw_core/lib/erc20_token_legacy.dart
```diff
@@ -0,0 +1,193 @@
...
+          final merged = _mergeByLowercaseContract(box.values.toList());
+
+          for (final rawName in entry.value) {
+            for (final token in merged) {
+              await token.migrateToSqlite(walletName: rawName, chainId: chainEntry.key);
+            }
+          }
+
+          await box.clear();
+          await box.deleteFromDisk();
```
MrCyjaneK: If this process is interrupted in a middle of migration it will cause duplicate data to get inserted (which is probably fine currently, only because we do conflict resolution with replace by default) - so in future we can break old installs by changing some unrelated code

reference: https://github.com/cake-tech/cake_wallet/blob/3b098e35ff45fa8b35ad31fdeea754c8ca498923/cw_core/lib/wallet_info_legacy.dart#L303

MrCyjaneK (commented):
Overall looks good, though did we consider getting rid of wallet name and using wallet id instead?

Blazebrain:
> Overall looks good, though did we consider getting rid of wallet name and using wallet id instead?

Yeah, we did consider it, but the two options for a wallet id aren't really solid compared to the name: `WalletInfo.id` is basically the name and type together (so it changes on rename anyway), and `walletInfoId` is independent of the name but not really durable because of the way we currently handle `WalletInfo.save()`

Plus using the wallet name is similar to what we currently use for other tables like Trade and Bridge

OmarHatem28 (approved):
LGTCzarek :D

PR #3451, opened 2026-07-28
Refactor Receive Flow - SDS

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/new-ui/widgets/receive/payjoin_copy_modal.dart
```diff
@@ -0,0 +1,77 @@
...
+                        iconPath: "assets/new-ui/pjmodal_btc.svg",
+                        onTap: () {
+                          ClipboardUtil.setSensitiveDataToClipboard(
+                            ClipboardData(
+                              text: uri.amount.isNotEmpty
+                                  ? BitcoinURI(amount: uri.amount, address: uri.address).toString()
+                                  : uri.address,
+                            ),
+                            isSensitive: true,
+                          );
```
konstantinullrich, line 53: Why is this sensitive now?

konstantinullrich:
@Blazebrain Why did you copy the files from `lib/new-ui/widgets/receive_page/` to `lib/new-ui/widgets/receive/`? This makes it pretty hard to review the changes?

Blazebrain:
> @Blazebrain Why did you copy the files from `lib/new-ui/widgets/receive_page/` to `lib/new-ui/widgets/receive/`? This makes it pretty hard to review the changes?

Receive covers other pages asides the receive page, and widgets from those pages are also in there, thats the reason for the files move

Also, this is to be a refactor PR, so all the changes that need to be done are actually supposed to be done here, including file naming and restructing.

lib/core/fiat_rate_service.dart
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  ExchangeRate _pairFor(CryptoCurrency crypto, FiatCurrency fiat, double rate) {
+    if (rate * rate < math.pow(10.0, crypto.decimals - fiat.decimals)) {
+      final digits = math.min(crypto.decimals, 20);
+      return ExchangeRate(
+        base: fiat,
+        quote: Money.parse((1 / rate).toStringAsFixed(digits), crypto),
+      );
+    }
```
konstantinullrich:
```suggestion
```
This is not needed since Exchange rate handles different decimals and the price is always in fiat

lib/core/fiat_rate_service.dart
```diff
@@ -0,0 +1,158 @@
...
+    return ExchangeRate(
+      base: crypto,
+      quote: Money.parse(rate.toStringAsFixed(fiat.decimals), fiat),
```
konstantinullrich:
```suggestion
      quote: rate.toMoney(fiat),
```
Please use double.toMoney since it has additional guard rails

lib/core/fiat_rate_service.dart
```diff
@@ -0,0 +1,158 @@
...
+
+  double? rateFor(CryptoCurrency crypto, FiatCurrency fiat) {
+    if (fiat == _settingsStore.fiatCurrency) {
+      final live = _fiatConversionStore.prices[crypto];
+      if (live != null) {
+        return live;
+      }
+    }
+    return _customRates[crypto]?[fiat];
+  }
```
konstantinullrich: Why not use Money here?

lib/core/fiat_rate_service.dart
```diff
@@ -0,0 +1,158 @@
...
+  void _emitRateChanged() {
+    if (_rateController.isClosed) {
+      return;
+    }
+    _rateController.add(null);
+  }
+
+  final FiatConversionStore _fiatConversionStore;
+  final SettingsStore _settingsStore;
+  final StreamController<void> _rateController = StreamController<void>.broadcast();
```
konstantinullrich: Why have an empty notifier stream here? Why not put the change in the stream?

lib/new-ui/viewmodels/receive/receive_bloc.dart
```diff
@@ -0,0 +1,675 @@
...
+import "package:equatable/equatable.dart";
+part "receive_event.dart";
+part "receive_state.dart";
+
+class ReceiveBloc extends Bloc<ReceiveEvent, ReceiveState> {
+  ReceiveBloc({
+    required this.addressService,
+    required this.fiatRateService,
+    required this.activeWalletService,
+    bool lightningMode = false,
```
konstantinullrich: Why not just have initialToken set to CryptoCurrency.btcln

cw_core/lib/amount/exchange_rate.dart
```diff
@@ -11,6 +14,32 @@ class ExchangeRate {
...
+    final crossover = sqrt(pow(10.0, base.decimals - quoteCurrency.decimals));
+    if (rate < crossover) {
+      final quote = (1 / rate).tryToMoney(base);
+      return quote == null ? null : ExchangeRate(base: quoteCurrency, quote: quote);
+    }
+
+    final quote = rate.tryToMoney(quoteCurrency);
+    return quote == null ? null : ExchangeRate(base: base, quote: quote);
+  }
+
```
konstantinullrich, line 42: Please write a test for this function

lib/core/address_service.dart
```diff
@@ -0,0 +1,667 @@
...
+  WalletBase? _preLightningWallet;
+  ReceivePageOption? _preLightningType;
+
+  WalletBase get wallet => _wallet();
+  AmountParsingProxy get _amountParsingProxy => _amountParsingProxyGetter();
+
+  String get walletName => wallet.name;
+  WalletType get walletType => wallet.type;
+  int? get walletChainId => wallet.chainId;
+  String get walletId => wallet.walletInfo.name;
```
konstantinullrich:
```suggestion
  String get walletId => wallet.id;
```

cw_core/lib/amount/exchange_rate.dart
```diff
@@ -11,6 +14,32 @@ class ExchangeRate {
...
+      return null;
+    }
+
+    final crossover = sqrt(pow(10.0, base.decimals - quoteCurrency.decimals));
+    if (rate < crossover) {
+      final quote = (1 / rate).tryToMoney(base);
+      return quote == null ? null : ExchangeRate(base: quoteCurrency, quote: quote);
+    }
+
+    final quote = rate.tryToMoney(quoteCurrency);
```
konstantinullrich, line 39: double.toMoney always returns so why are you using tryToMoney here?

lib/new-ui/pages/addresses_page.dart
```diff
@@ -1,361 +1,737 @@
...
-    setState(() {
-      items = getItems(widget.addressListViewModel.forceRecomputeItems);
-    });
-  }
+  @override
+  Widget build(BuildContext context) => Column(
+        spacing: 12,
+        children: [
+          _TopBar(title: S.of(context).addresses),
+          const Expanded(child: Center(child: CircularProgressIndicator())),
```
konstantinullrich: I think we use the cupertino spinner normally e.g. in the Trezor connection screen

lib/new-ui/widgets/receive/receive_amount_display.dart
```diff
@@ -0,0 +1,96 @@
...
+class ReceiveAmountDisplay extends StatelessWidget {
+  const ReceiveAmountDisplay({
+    required this.displayAmount,
+    required this.cryptoSymbol,
+    required this.fiatEquivalent,
+    required this.largeQrMode,
+    super.key,
+  });
+
+  final String displayAmount;
```
konstantinullrich: Why not use Money here?

lib/new-ui/widgets/receive/receive_amount_display.dart
```diff
@@ -0,0 +1,96 @@
...
+                              ),
+                            ),
+                            Text(
+                              cryptoSymbol,
+                              style: TextStyle(
+                                color: Theme.of(context).colorScheme.primary,
+                                fontSize: 16,
+                                fontWeight: FontWeight.w500,
+                              ),
+                            ),
```
konstantinullrich: Please Use `MoneyText` here

lib/new-ui/widgets/receive/receive_amount_display.dart
```diff
@@ -0,0 +1,96 @@
...
+                    if (fiatEquivalent != null)
+                      Padding(
+                        padding: const EdgeInsets.all(8),
+                        child: Text(
+                          "${fiatEquivalent!.toStringWithPrecision()} ${fiatEquivalent!.currency.symbol}",
+                          style: TextStyle(
+                            color: Theme.of(context).colorScheme.onSurfaceVariant,
+                            fontSize: 16,
+                            fontWeight: FontWeight.w500,
+                          ),
```
konstantinullrich: Same here, please use `MoneyText`

lib/new-ui/widgets/receive/receive_amount_display.dart
```diff
@@ -0,0 +1,96 @@
...
+  const ReceiveAmountDisplay({
+    required this.displayAmount,
+    required this.cryptoSymbol,
+    required this.fiatEquivalent,
+    required this.largeQrMode,
+    super.key,
+  });
+
+  final String displayAmount;
+  final String cryptoSymbol;
```
konstantinullrich: Not needed with MoneyText

lib/new-ui/widgets/receive/payjoin_copy_modal.dart
```diff
@@ -0,0 +1,96 @@
...
+                      ),
+                      ListItemRegularRow(
+                        keyValue: "pj",
+                        label: "Payjoin",
+                        iconPath: "assets/new-ui/pjmodal_pj.svg",
+                        onTap: () {
+                          ClipboardUtil.setSensitiveDataToClipboard(
+                            ClipboardData(text: uri.toString()),
+                            isSensitive: true,
+                          );
```
konstantinullrich, line 66: Same here, a public facing URI is not that sensitive

lib/new-ui/widgets/coins_page/zcash_migration_modal.dart
konstantinullrich, line 1: Except the import this seems to be an unrelated change

lib/new-ui/widgets/coins_page/unconfirmed_balance_widget.dart
konstantinullrich, line 1: Except the import this seems to be an unrelated change

lib/new-ui/widgets/coins_page/action_row/coin_action_row.dart
```diff
@@ -91,8 +89,9 @@ class CoinActionRow extends StatelessWidget {
             ),
             label: S.of(context).receive,
             action: () async {
+              final page =
+                  getIt.get<ReceivePage>(param1: lightningMode ? CryptoCurrency.btcln : null);
```
konstantinullrich, line 93: ❤️

lib/new-ui/widgets/receive/receive_amount_modal.dart
```diff
@@ -0,0 +1,268 @@
...
+                                    child: Row(
+                                      mainAxisAlignment: MainAxisAlignment.center,
+                                      spacing: 4,
+                                      children: [
+                                        Text(
+                                          widget.selectedCurrencySymbol,
+                                          style: TextStyle(
+                                            color: Theme.of(context).colorScheme.onSurface,
+                                          ),
+                                        ),
```
konstantinullrich: Please use `CurrencySymbolText`

lib/new-ui/widgets/receive/receive_amount_modal.dart
```diff
@@ -0,0 +1,268 @@
...
+  });
+
+  final String initialAmount;
+  final String selectedCurrencySymbol;
+  final int selectedCurrencyDecimals;
+  final bool useSatoshi;
+  final bool showTokenPicker;
+  final String tokenIconPath;
+  final String tokenTitle;
+  final void Function(String amount) onAmountSubmitted;
```
konstantinullrich: Why not `void Function(Money amount)`?

lib/new-ui/widgets/receive/receive_amount_modal.dart
```diff
@@ -0,0 +1,268 @@
...
+    required this.tokenTitle,
+    required this.onAmountSubmitted,
+    required this.onCurrencyPickerTap,
+    required this.onTokenPickerTap,
+    super.key,
+  });
+
+  final String initialAmount;
+  final String selectedCurrencySymbol;
+  final int selectedCurrencyDecimals;
```
konstantinullrich: This can be replace by `Currency selectedCurrency`

lib/new-ui/widgets/receive/receive_amount_modal.dart
```diff
@@ -0,0 +1,268 @@
...
+  final bool useSatoshi;
+  final bool showTokenPicker;
+  final String tokenIconPath;
+  final String tokenTitle;
```
konstantinullrich: This can be replace by Currency selectedCurrency

lib/new-ui/widgets/receive/receive_amount_modal.dart
```diff
@@ -0,0 +1,268 @@
...
+                                      Row(
+                                        spacing: 8,
+                                        children: [
+                                          ExcludeSemantics(
+                                            child: TokenImageWidget(
+                                              imageUrl: widget.tokenIconPath,
+                                              size: 32,
+                                            ),
+                                          ),
+                                          Text(widget.tokenTitle.toUpperCase()),
```
konstantinullrich: Please use `CurrencySymbolText`

lib/new-ui/widgets/receive/receive_address_type_selector.dart
```diff
@@ -0,0 +1,361 @@
...
+import "package:cw_core/wallet_type.dart";
+import "package:flutter/material.dart";
+import "package:modal_bottom_sheet/modal_bottom_sheet.dart";
+
+class ReceiveAddressTypeSelector extends StatefulWidget {
+  const ReceiveAddressTypeSelector({
+    required this.options,
+    required this.selected,
+    required this.walletType,
+    required this.isLightning,
```
konstantinullrich: Why not set `selected` to the Lightning Option instead and omit the `isLightning`?

lib/new-ui/widgets/receive/receive_large_amount_preview.dart
```diff
@@ -0,0 +1,79 @@
...
+class ReceiveLargeAmountPreview extends StatelessWidget {
+  const ReceiveLargeAmountPreview({
+    required this.amount,
+    required this.currency,
+    required this.largeQrMode,
+    super.key,
+  });
+
+  final String amount;
+  final String currency;
```
konstantinullrich:
```suggestion
  final Money amount;
```

lib/new-ui/widgets/receive/receive_large_amount_preview.dart
```diff
@@ -0,0 +1,79 @@
...
+                            color: Theme.of(context).colorScheme.onSurface,
+                          ),
+                        ),
+                        Text(
+                          currency,
+                          style: TextStyle(
+                            fontSize: 20,
+                            fontWeight: FontWeight.w500,
+                            color: Theme.of(context).colorScheme.onSurfaceVariant,
+                          ),
```
konstantinullrich:
```suggestion
                        CurrencySymbolText(
                          amount.currency,
                          style: TextStyle(
                            fontSize: 20,
                            fontWeight: FontWeight.w500,
                            color: Theme.of(context).colorScheme.onSurfaceVariant,
                          ),
```

lib/new-ui/widgets/receive/receive_large_amount_preview.dart
```diff
@@ -0,0 +1,79 @@
...
+                      spacing: 4,
+                      children: [
+                        Text(
+                          amount,
+                          style: TextStyle(
+                            fontSize: 20,
+                            fontWeight: FontWeight.w500,
+                            color: Theme.of(context).colorScheme.onSurface,
+                          ),
+                        ),
```
konstantinullrich:
```suggestion
                        MoneyText(
                          amount,
                          showSymbol: false,
                          style: TextStyle(
                            fontSize: 20,
                            fontWeight: FontWeight.w500,
                            color: Theme.of(context).colorScheme.onSurface,
                          ),
                        ),
```

lib/new-ui/widgets/receive/receive_token_display.dart
```diff
@@ -0,0 +1,67 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final chainAsset =
+        walletType == WalletType.bsc ? "bnb" : walletTypeToString(walletType).toLowerCase();
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.center,
+      spacing: 8,
+      children: [
+        Text(
+          token.title,
```
konstantinullrich:
```suggestion
        CurrencySymbolText(
          token,
```

lib/reactions/wallet_utils.dart
konstantinullrich, line 1: Isn't this unrelated?

lib/router.dart
```diff
@@ -462,22 +457,13 @@ Route<dynamic> createRoute(RouteSettings settings) {
       return CupertinoPageRoute<void>(
           fullscreenDialog: true, builder: (_) => getIt.get<SendTemplatePage>());
 
-    case Routes.receive:
-      return CupertinoPageRoute<void>(
-          builder: (context) => getIt.get<ReceivePage>(), settings: settings);
-
     case Routes.addressPage:
-      return handleRouteWithPlatformAwareness((context) => getIt.get<AddressPage>(),
+      return handleRouteWithPlatformAwareness((context) => getIt.get<ReceivePage>(),
```
konstantinullrich, line 461: Why does **addressPage** route to the `ReceivePage` instead of the `AddressesPage`?

konstantinullrich (commented):
I am still reviewing the last 11 Files. But here are more of my findings so far

lib/new-ui/widgets/receive/receive_amount_display.dart
```diff
@@ -0,0 +1,104 @@
...
+                                ),
+                              ),
+                              CurrencySymbolText(
+                                amount.currency,
+                                style: TextStyle(
+                                  color: Theme.of(context).colorScheme.primary,
+                                  fontSize: 16,
+                                  fontWeight: FontWeight.w500,
+                                ),
+                              ),
```
konstantinullrich: NIT: this can be ommited by setting `MoneyText.showSymbol` to true since the style is the same.

lib/new-ui/widgets/receive/receive_amount_modal.dart
```diff
@@ -0,0 +1,276 @@
...
+                                      ),
+                                      color: Theme.of(context).colorScheme.surfaceContainerHigh,
+                                    ),
+                                    child: Row(
+                                      mainAxisAlignment: MainAxisAlignment.center,
+                                      spacing: 4,
+                                      children: [
+                                        CurrencySymbolText(
+                                          widget.selectedCurrency,
+                                          useBaseUnit: widget.useSatoshi,
```
konstantinullrich: **NIT:** not needed since the `CurrencySymbolText` automatically applies this setting depending on the user setting. 
```suggestion
```

lib/router.dart
konstantinullrich: Odd historic naming choice

lib/core/address_service.dart
```diff
@@ -0,0 +1,631 @@
...
+      return;
+    }
+
+    final wallet = this.wallet;
+    _emitPayjoin(_emptyToNull(bitcoin!.getPayjoinEndpoint(wallet)));
+
+    _payjoinDisposer = mobx.reaction<String>(
+      (_) => bitcoin!.getPayjoinEndpoint(wallet),
+      (value) => _emitPayjoin(_emptyToNull(value)),
+    );
```
konstantinullrich, line 514: I might be mistaken, but don't mobx reaction only fire on observables?

Blazebrain: Yeah, mobx reactions only re-run on observable changes. 

This one works because `getPayjoinEndpoint` reads `payjoinEndpoint`, which is an observable on `BitcoinWalletAddresses`, so it fires when the payjoin receiver sets or renews the endpoint.

lib/new-ui/widgets/receive/receive_amount_display.dart
```diff
@@ -51,13 +50,10 @@ class ReceiveAmountDisplay extends StatelessWidget {
...
                       child: Padding(
                         padding: const EdgeInsets.all(8),
-                        child: Row(
-                          spacing: 8,
-                          children: [
-                            if (amount != null) ...[
-                              MoneyText(
+                        child: amount == null
+                            ? const SizedBox.shrink()
+                            : MoneyText(
```
konstantinullrich: **NIT:** `MoneyText.optional` does exactly that

konstantinullrich (approved):

lib/core/active_wallet_service.dart
```diff
@@ -0,0 +1,39 @@
+import "dart:async";
+
+import "package:cake_wallet/store/app_store.dart";
+import "package:cw_core/wallet_base.dart";
+import "package:mobx/mobx.dart" as mobx;
+
+class ActiveWalletService {
```
malik1004x, line 7: You're essentially putting a wrapper around `appStore.wallet` – we do want to drop dependency from MobX eventually but I'd argue this is not the PR to do it in.

The Receive flow does not allow for changing wallets while the page is open – you might as well just pass a `WalletBase` object straight to the Bloc.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+
+  bool get hasTokens => wallet_utils.hasTokens(wallet.type);
+
+  bool get isInfoboxDismissed => wallet.walletInfo.receiveInfoboxDismissed;
+
+  bool get hasAccounts => const {WalletType.monero, WalletType.wownero}.contains(wallet.type);
+
+  bool get hasHiddenAddresses => wallet.walletAddresses.hiddenAddresses.isNotEmpty;
+
+  List<AddressGroup> computeAddressList() {
```
malik1004x, line 64: This class is filled with chain-specific conditions.

What you could do instead to make this cleaner is a `List<AddressGroup> get addressList` getter in `WalletBase`, that is then overridden in specific chains. You could even do a generic implementation that maps the subaddress list for wallets that only have one address type. This way, we avoid wallet-specific conditions in the service layer, and future chain implementations don't involve going through the ViewModels and searching for conditions based on `WalletType`.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+  WalletType get walletType => wallet.type;
+  String get walletId => wallet.id;
+  CryptoCurrency get walletCurrency => wallet.currency;
+
+  List<CryptoCurrency> get receivableTokens =>
+      wallet.balance.keys.whereType<CryptoCurrency>().toList();
+
+  bool get hasTokens => wallet_utils.hasTokens(wallet.type);
+
+  bool get isInfoboxDismissed => wallet.walletInfo.receiveInfoboxDismissed;
```
malik1004x: I don't see how the infobox is related to addresses – in the interest of not cluttering up this class, could you move it to its own?

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+    required SettingsStore settingsStore,
+    required AmountParsingProxy Function() amountParsingProxyGetter,
+  })  : _wallet = wallet,
+        _settingsStore = settingsStore,
+        _amountParsingProxyGetter = amountParsingProxyGetter {
+    _walletSub = walletChanges.listen((_) => _bindPayjoin());
+    _bindPayjoin();
+  }
+
+  final WalletBase Function() _wallet;
```
malik1004x: I don't see why this has to be a function pointer – you could just pass `AppStore` and use the `wallet` getter.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+  final SettingsStore _settingsStore;
+  final AmountParsingProxy Function() _amountParsingProxyGetter;
```
malik1004x: You pass this function pointer here just for one getter, then that one getter is only used in `useSatoshi`, then `useSatoshi` is only used when initializing the state, then the variable in that sate is only read in one place – `ReceiveAmountModal`. You could instead use `BaseUnit.useBaseUnitOf(context)` in that one widget and then that check only goes through one new file instead of five.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+  late final StreamSubscription<WalletBase> _walletSub;
+  mobx.ReactionDisposer? _payjoinDisposer;
+
+  WalletBase? _preLightningWallet;
+  ReceivePageOption? _preLightningType;
+
+  WalletBase get wallet => _wallet();
+  AmountParsingProxy get _amountParsingProxy => _amountParsingProxyGetter();
+
+  String get walletName => wallet.name;
```
malik1004x: You can just use `wallet.name` directly, same for the lines below.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+  CryptoCurrency get walletCurrency => wallet.currency;
+
+  List<CryptoCurrency> get receivableTokens =>
+      wallet.balance.keys.whereType<CryptoCurrency>().toList();
+
+  bool get hasTokens => wallet_utils.hasTokens(wallet.type);
+
+  bool get isInfoboxDismissed => wallet.walletInfo.receiveInfoboxDismissed;
+
+  bool get hasAccounts => const {WalletType.monero, WalletType.wownero}.contains(wallet.type);
```
malik1004x: No changes needed for now as this is genuinely the only way to check for this, but we'll soon have a refactor after which you can probably just do `wallet.supportsAccounts` or `wallet is AccountWallet`.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+    if (isEVMCompatibleChain(type)) {
+      return _singleAddressGroup(evm!.getAddress(wallet));
+    }
+    if (type == WalletType.solana) {
+      return _singleAddressGroup(solana!.getAddress(wallet));
+    }
+    if (type == WalletType.tron) {
+      return _singleAddressGroup(tron!.getAddress(wallet));
+    }
+    if (type == WalletType.nano || type == WalletType.banano) {
```
malik1004x: Why do you have logic for dead wallet types (Wownero and Banano)?

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+          bitcoin!.getSilentPaymentReceivedAddresses(wallet).map(_electrumEntry).toList();
+      return [
+        AddressGroup(entries: main),
+        AddressGroup(header: const SilentPaymentsReceivedHeader(), entries: received),
+      ];
+    }
+
+    var entries = bitcoin!.getSubAddresses(wallet).map(_electrumEntry).toList();
+
+    if (wallet.type == WalletType.litecoin && entries.length >= _mwebTruncationThreshold) {
```
malik1004x, line 130: You seem to be truncating the address list for MWEB, why? I know it's super long, but if the addresses were generated I think we should keep showing them. If the list is so long it causes performance issues, use `ListView` or `CustomScrollView`.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+    wallet.walletAddresses.address = newAddress;
+    return true;
+  }
+
+  Future<bool> addManualAddress(String label) async {
+    final newAddress = await _generateNewAddress(label);
+    return newAddress != null && newAddress.isNotEmpty;
+  }
+
+  Future<String?> _generateNewAddress(String label) async {
```
malik1004x: This could also be a method in `WalletBase` – for the chains that don't support it, we could just keep it as `=> null`. Or better yet, put it in `mixin AddressGenerationWallet` and add `with AddressGenerationWallet` to supported chains, then type-check that whereever you call it. Again, this way we avoid chain-specific logic in the services, that can now stay blind to the wallet-side implementation.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+  String get currentAddress => wallet.walletAddresses.address;
+
+  Future<void> setActiveAddress(String address) async {
+    wallet.walletAddresses.address = address;
+  }
+
+  Future<bool> rotateAddress() async {
+    final wallet = this.wallet;
+    final newAddress = await _generateNewAddress("");
+    if (newAddress == null || newAddress.isEmpty) {
```
malik1004x: If this condition is added because one of the functions has a possibility of returning `null` or `""`, please make sure the functions always either return a good value or throw an exception – then this safeguard will be unnecessary.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+      printV("rotateAddress: no new address for ${wallet.type}");
+      return false;
+    }
+    if (!identical(wallet, this.wallet)) {
+      return false;
+    }
+    wallet.walletAddresses.address = newAddress;
```
malik1004x, line 204: If implemented wallet-side, this assignment should be part of `generateNewAddress` – you could also add a `bool setAsActive = false`. This way you don't have to do the `identical` check because there is no possibility of assigning to a wrong wallet if it's unloaded at that time.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+        AddressGroup(entries: [AddressEntry(address: address)]),
+      ];
+
+  String get currentAddress => wallet.walletAddresses.address;
+
+  Future<void> setActiveAddress(String address) async {
+    wallet.walletAddresses.address = address;
+  }
+
+  Future<bool> rotateAddress() async {
```
malik1004x: Please don't use `bool` to indicate errors. You can use an exception and it's immediately understandable – with a `bool` return value you have to go into the function and check what that value actually is.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+    if (!identical(wallet, this.wallet)) {
+      return false;
+    }
+    wallet.walletAddresses.address = newAddress;
+    return true;
+  }
+
+  Future<bool> addManualAddress(String label) async {
+    final newAddress = await _generateNewAddress(label);
+    return newAddress != null && newAddress.isNotEmpty;
```
malik1004x: Same as above, please don't use `bool` to indicate errors – also, `_generateNewAddress` should probably be written in a way where either the value is correct or it throws an exception of its own.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+    final type = wallet.type;
+
+    if (_isElectrumType(type)) {
+      final isSilentPayments = bitcoin!.hasSelectedSilentPayments(wallet);
+      final before = isSilentPayments
+          ? bitcoin!.getSilentPaymentAddresses(wallet).map((a) => a.address).toSet()
+          : bitcoin!.getSubAddresses(wallet).map((a) => a.address).toSet();
+      await bitcoin!.generateNewAddress(wallet, label);
+      await wallet.save();
+      final after = isSilentPayments
```
malik1004x: So, you're getting _every_ value _twice_ to check the difference? Are you sure we can't have a wallet-side solution that doesn't involve recomputing the entire list for one value?

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+      final fresh = subs.where((s) => !beforeIds.contains(s.id)).firstOrNull;
+      if (fresh == null) {
+        return null;
+      }
+      wallet.walletAddresses.manualAddresses.add(fresh.address);
+      await wallet.save();
+      return fresh.address;
+    }
+
+    if (type == WalletType.wownero) {
```
malik1004x: Why is there logic for Wownero?

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+      }
+      wallet.walletAddresses.manualAddresses.add(fresh.address);
+      await wallet.save();
+      return fresh.address;
+    }
+
+    return null;
+  }
+
+  Future<void> setLabel(String address, String label) async {
```
malik1004x: Again, adding a wallet-side method would mean less lines and no wallet-specific code in the service layer – consider that for a cleaner approach since it's a refactor PR after all.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+    final type = wallet.type;
+    return _isElectrumType(type) ||
+        type == WalletType.decred ||
+        type == WalletType.monero ||
+        type == WalletType.wownero;
+  }
+
+  bool get canHide => wallet.type != WalletType.zcash;
+
+  int? _entryIdFor(String address) {
```
malik1004x: This could have a substantial impact on performance if the wallet has a lot of addresses. The only place it seems used is `setLabel` – have you considered haivng `setLabel` accept an `AddressEntry` object, removing the need for this loop-over?

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+    }
+    if (type == WalletType.wownero) {
+      wownero!.getSubaddressList(wallet).update(
+            wallet,
+            accountIndex: wownero!.getCurrentAccount(wallet).id,
+          );
+    }
+  }
+
+  ReceivePageOption? get selectedAddressType {
```
malik1004x: Other wallets also have a `selectedAddressType` – it's just one option, but it's not `null`. You can again make this a generic getter in `WalletBase` and override it in `LitecoinWallet`, `BitcoinWallet` and `ZcashWallet`.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+      return bitcoin!.getSelectedAddressType(wallet);
+    }
+    if (type == WalletType.zcash) {
+      return zcash!.getSelectedAddressType(wallet);
+    }
+    return null;
+  }
+
+  List<ReceivePageOption> get addressTypeOptions =>
+      wallet.walletAddresses.receivePageOptions.where(wallet.receiveOptionAvailable).toList();
```
malik1004x, line 324: [minor] This filtering can also be done wallet-side.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+
+  Future<void> setAddressType(ReceivePageOption option) => _setAddressTypeOn(wallet, option);
```
malik1004x, line 326: You could also consider making the address type selection _only_ a part of the Receive page – this way, setting an option there has no side effects across the wallet and if something goes wrong you can just re-open the receive page to reset everything. For example, consider `WalletBase` having a `List<ReceivePageOption> get addressTypeOptions` getter, and in the `ReceiveState` you hold the selected option and get addresses based on that option in functions like `AddressEntry getCurrentAddress({ReceivePageOption type = some_sensible_default})` and `List<AddressEntry> allAddressesFor(ReceivePageOption type)`.

We've had several critical bugs stemming from the receive option being persistent across the app instead of isolated to the receive page – namely, Cupcake view-only wallets crashed and became irrecoverable when Silent Payments was selected on the Receive page (that change became persistent and the wallet kept failing to load addresses forever) and Swap becoming unusable because Lightning was selected on the Receive page and `ExchangeViewModel` kept trying to load a Lightning address to swap on-chain Bitcoin. Both of these are fixed with workarounds, but I think it would be even better if we attacked the problem at its root and removed any possibility of that happening.

The same issue stemmed with Coin Control, which persisted across instances of the Send page and had reset logic bound to when the page was opened. We had tons of workarounds for that, but a complete refactor where the Send page holds the coin selection is the only thing that was able to fix it for good. To avoid future headaches please consider a similar approach here.

lib/core/address_service.dart
```diff
@@ -0,0 +1,611 @@
...
+    }
+
+    final current = bitcoin!.getSelectedAddressType(wallet);
+    final lightning = bitcoin!.getBitcoinLightningReceivePageOption();
+
+    if (lightningMode) {
+      if (current == lightning) {
+        return;
+      }
+      _preLightningWallet = wallet;
```
malik1004x: I believe `_preLightningWallet` was created because in the time it takes to select the receive option the wallet can technically change, but are you sure it's the best way to handle that? You can move some logic to the wallet code instead to prevent having this variable for this one specific case. Also, I believe the logic can be made generic, for example by calling `selectReceivePageOption(_lightningMode ? .lightning : .segwit)` from the UI.

lib/core/address_types.dart
```diff
@@ -0,0 +1,43 @@
+import "package:cw_core/amount/money.dart";
+
+class AddressEntry {
```
malik1004x, line 3: You could potentially hold `ReceivePageOption` here to avoid querying the wallets as to what address type is actually selected.

lib/core/fiat_rate_service.dart
```diff
@@ -0,0 +1,130 @@
...
+  final SettingsStore _settingsStore;
+  final FiatConversionStore _fiatConversionStore;
+  late final mobx.ReactionDisposer _disposeReaction;
+  final StreamController<FiatCurrency> _rateChangesController =
+      StreamController<FiatCurrency>.broadcast();
+
+  FiatCurrency get currentFiat => _settingsStore.fiatCurrency;
+  Stream<FiatCurrency> get rateChanges => _rateChangesController.stream;
+
+  final Map<CryptoCurrency, Map<FiatCurrency, ExchangeRate>> _fetchedRates = {};
```
malik1004x: Having a nested `Map` isn't great for readability – you could instead do an `ExchangePair` class with a `CryptoCurrency` and `FiatCurrency` fields and use that as a key. This way, you only have a single map.

lib/core/fiat_rate_service.dart
```diff
@@ -0,0 +1,130 @@
...
+  void _notify(FiatCurrency fiat) {
+    if (_rateChangesController.isClosed) {
+      return;
+    }
+    _rateChangesController.add(fiat);
+  }
+
+  ExchangeRate? _rateFor(CryptoCurrency crypto, FiatCurrency fiat) {
+    if (fiat == _settingsStore.fiatCurrency) {
+      final live = _fiatConversionStore.prices[crypto];
```
malik1004x: This works, but you could also unify this class with `FiatConversionStore`. I don't think having two separate services is necessary if `FiatRateService` accomplishes all goals. You would then also not need special conditions for `_settingsStore.fiatCurrency`.

lib/new-ui/pages/buy_sell/buy_sell_amount_page.dart
```diff
@@ -67,116 +67,113 @@ class _NewBuySellAmountPageState extends State<NewBuySellAmountPage> {
```
malik1004x, line 74: You seem to have reformatted the buy/sell page files by accident, could you revert them please?

lib/new-ui/pages/addresses_page.dart
```diff
@@ -1,361 +1,761 @@
...
+          listenWhen: (previous, current) {
+            final prev = previous is AddressesLoaded ? previous.failureCode : null;
+            final curr = current is AddressesLoaded ? current.failureCode : null;
+            return curr != null && curr != prev;
+          },
+          listener: (context, state) {
+            if (state is! AddressesLoaded || state.failureCode == null) {
+              return;
+            }
+            showBar(context, S.of(context).error_dialog_content);
```
malik1004x: I believe the purpose of this is to show single-fire error popups – `emitPresentation()` would be a much better choice for this, allowing for cleaner logic and proper error classes instead of an `enum`. You could handle each error differently and have different messages, or handle them all in the same generic way replacing only the message. Since you only pass a localization string here, i believe this will only show a generic error popup that doesn't tell you what's actually happening – again, could be much better handled with `emitPresentation`. This would also remove the need for the `listenWhen` above.

Also, `showBar` is an Old UI component that doesn't fit well with the 6.x design and since it's auto-closing it may close before users read the whole message. Please consider using `showPopUp` instead which was adjusted to New UI and has a smaller chance of being missed by the user.

lib/new-ui/pages/addresses_page.dart
```diff
@@ -1,361 +1,761 @@
...
+                  ],
+                ),
+              ),
+            ),
+          ),
+        ),
+      );
+}
+
+class _TopBar extends StatelessWidget {
```
malik1004x, line 339: This seems to be initialized the same way for all subclasses of this page. You could also initialize it once and put `BlocBuilder` around the second element of the column, it would save us some lines.

lib/new-ui/pages/addresses_page.dart
```diff
@@ -1,361 +1,761 @@
...
+import "package:flutter_bloc/flutter_bloc.dart";
+import "package:flutter_mobx/flutter_mobx.dart";
+import "package:mobx/mobx.dart";
+import "package:modal_bottom_sheet/modal_bottom_sheet.dart";
+
+class AddressesPage extends StatelessWidget {
+  const AddressesPage({super.key, this.showHidden = false, this.onSelect});
+
   final bool showHidden;
+  final void Function(String address)? onSelect;
```
malik1004x: This is already done as part of the Old UI removal PR, merged to `dev` a few days ago. Please use the version from that PR as yours has issues with UI styling (checked on a build).

lib/new-ui/pages/addresses_page.dart
```diff
@@ -1,361 +1,761 @@
...
 
   @override
-  State<NewAddressesPage> createState() => _NewAddressesPageState();
+  Widget build(BuildContext context) => BlocProvider<AddressesBloc>(
+        create: (_) => getIt<AddressesBloc>(param1: showHidden),
```
malik1004x: If you go to the hidden addresses page, that's a separate instance of this Bloc, causing address selection to stop working. You could instead pass the Bloc as a parameter and then you could do proper dependency injection instead of `BlocProvider`. `BlocBuilder` and `BlocListener` constructors have a `bloc` input, `BlocProvider` is unnecessary. Then, you also don't need to to `context.read` to get the Bloc instance.

lib/new-ui/pages/addresses_page.dart
```diff
@@ -1,361 +1,761 @@
...
+          child: Semantics(
+            button: true,
+            hint: S.of(context).choose_account,
+            child: InkWell(
+              borderRadius: BorderRadius.circular(16),
+              onTap: () async {
+                final bloc = context.read<AddressesBloc>();
+                await showPopUp<void>(
+                  context: context,
+                  builder: (_) => getIt.get<MoneroAccountListPage>(),
```
malik1004x: You seem to have re-added account switching from this page. This was removed deliberately as requested by UI during the 6.0 beta. You can remove it as `MoneroAccountListPage` doesn't exist in `dev` anymore anyway.

lib/new-ui/pages/receive_page.dart
```diff
@@ -20,381 +21,452 @@ import "package:cake_wallet/new-ui/widgets/receive_page/receive_qr_code.dart";
...
+  Widget build(BuildContext context) => BlocProvider<ReceiveBloc>(
+        create: (_) => getIt<ReceiveBloc>(param1: initialToken),
+        child: _ReceivePageBody(initialToken: initialToken),
+      );
+}
+
+class _ReceivePageBody extends StatefulWidget {
+  const _ReceivePageBody({required this.initialToken});
+
+  final CryptoCurrency? initialToken;
```
malik1004x: This appears to never be passed into the bloc – since it appears to have been left over from a different approach to dependency injection, it's probably okay to remove it now.

Edit: I noticed now that it's passed into `ReceiveFailureWidget` just for one `ReceiveOpened` call. Could you please just retain `initialToken` inside the bloc for any further page resets? This way we avoid passing this value around the widgets.

lib/new-ui/pages/receive_page.dart
```diff
@@ -20,381 +21,452 @@ import "package:cake_wallet/new-ui/widgets/receive_page/receive_qr_code.dart";
...
+
+class ReceivePage extends StatelessWidget {
+  const ReceivePage({super.key, this.initialToken});
+
+  final CryptoCurrency? initialToken;
+
+  @override
+  Widget build(BuildContext context) => BlocProvider<ReceiveBloc>(
+        create: (_) => getIt<ReceiveBloc>(param1: initialToken),
+        child: _ReceivePageBody(initialToken: initialToken),
```
malik1004x: I don't get why `ReceivePage` and `_ReceivePageBody` are split – if it's just for the `BlocProvider`, then you can – same as in `AddressesPage` – replace it with proper dependency injection and just pass the `bloc` variable to `BlocBuilder` instances. Also, why was this renamed?

lib/new-ui/pages/receive_page.dart
```diff
@@ -20,381 +21,452 @@ import "package:cake_wallet/new-ui/widgets/receive_page/receive_qr_code.dart";
...
+            ],
+            begin: Alignment.topCenter,
+            end: Alignment.bottomCenter,
+          ),
+          borderRadius: const BorderRadius.vertical(top: Radius.circular(24)),
+        ),
+        child: SafeArea(
+          child: BlocConsumer<ReceiveBloc, ReceiveState>(
+            listenWhen: (previous, current) {
+              final prev = previous is ReceiveLoaded ? previous.failureCode : null;
```
malik1004x: Again, you can just use `emitPresentation`.

lib/new-ui/pages/receive_page.dart
```diff
@@ -20,381 +21,452 @@ import "package:cake_wallet/new-ui/widgets/receive_page/receive_qr_code.dart";
...
-          ],
+                ),
+            ],
+          ),
         ),
+      ],
+    );
+  }
+
+  bool _isMwebOption(ReceivePageOption? option) {
```
malik1004x: This could be contained in the `ReceivePageOption` class instead of being a UI-side helper function.

lib/new-ui/pages/receive_page.dart
```diff
@@ -20,381 +21,452 @@ import "package:cake_wallet/new-ui/widgets/receive_page/receive_qr_code.dart";
...
+    if (option == null) {
+      return false;
+    }
+    return (option.description ?? "").toLowerCase().contains("mweb");
+  }
+
+  String _qrEmbeddedIcon(ReceiveLoaded state) {
```
malik1004x: This could be a getter inside the state to prevent another UI helper function.

lib/new-ui/pages/receive_page.dart
```diff
@@ -20,381 +21,452 @@ import "package:cake_wallet/new-ui/widgets/receive_page/receive_qr_code.dart";
...
+
+  Future<void> _showLabelModal(BuildContext context, ReceiveLoaded state) async {
+    final bloc = context.read<ReceiveBloc>();
+    await showMaterialModalBottomSheet<String>(
+      context: context,
+      backgroundColor: Colors.transparent,
+      barrierColor: Colors.black.withAlpha(80),
+      builder: (_) => ReceiveLabelModal(
+        initialLabel: state.addressEntry.label ?? "",
+        onSubmit: (label) async => bloc.add(LabelSubmitted(label)),
```
malik1004x: [minor] Instead of passing a function pointer, you could return the label inside the `.pop` call.

lib/new-ui/pages/receive_page.dart
```diff
@@ -20,381 +21,452 @@ import "package:cake_wallet/new-ui/widgets/receive_page/receive_qr_code.dart";
...
-    showMaterialModalBottomSheet(
+    await showMaterialModalBottomSheet<void>(
       context: context,
       backgroundColor: Colors.transparent,
       barrierColor: Colors.black.withAlpha(80),
-      builder: (_) => getIt.get<ReceiveLabelModal>(param1: _addressItemWithLabel),
-    ).then((value) {
-      _reloadAddressWithLabel(widget.addressListViewModel.uri);
-    });
+      builder: (_) => BlocProvider<ReceiveBloc>.value(
```
malik1004x: You can just pass `bloc` to the `BlocBuilder` directly.

lib/new-ui/pages/receive_page.dart
```diff
@@ -20,381 +21,452 @@ import "package:cake_wallet/new-ui/widgets/receive_page/receive_qr_code.dart";
...
+    await FiatCurrencyPickerSheet.show(
+      context: context,
+      selected: state.inputCurrency,
+      cryptoOption: cryptoOption,
+      onSelected: (fiat) => bloc.add(InputCurrencySelected(fiat)),
+      onCryptoSelected: (crypto) => bloc.add(InputCurrencySelected(crypto)),
+    );
+  }
+
+  Future<void> _pickToken(
```
malik1004x: These are only used inside the amount modal, so why not put them there?

lib/new-ui/viewmodels/addresses/addresses_bloc.dart
```diff
@@ -0,0 +1,232 @@
...
+    on<ActiveAddressSet>(_onActiveAddressSet, transformer: sequential());
+    on<AddressHideToggled>(_onHideToggled, transformer: sequential());
+    on<AddressLabelSet>(_onLabelSet, transformer: sequential());
+    on<AddressAdded>(_onAddressAdded, transformer: droppable());
+    on<AddressListRefreshed>(_onListRefreshed, transformer: sequential());
+    on<_WalletChanged>(_onWalletChanged, transformer: restartable());
+
+    _walletSub = activeWalletService.walletChanges.listen((_) {
+      if (!isClosed) {
+        add(const _WalletChanged());
```
malik1004x, line 36: I don't see how this can be triggered. We don't allow changing the wallet from anywhere in the receive page, and it seems like quite a lot of logic for something that can't actually ever be accessed from the flow.

lib/new-ui/viewmodels/addresses/addresses_bloc.dart
```diff
@@ -0,0 +1,232 @@
...
+    try {
+      emit(_buildLoaded(showHidden: event.showHidden));
+    } catch (e) {
+      printV("AddressesBloc _onOpened failed: $e");
+      emit(const AddressesFailure(AddressesFailureCode.addressListUnavailable));
+    }
+  }
+
+  Future<void> _onSearch(SearchTermEntered event, Emitter<AddressesState> emit) async {
+    final loaded = state;
```
malik1004x: [minor] You could use the new nicer pattern recently added to the language:
```
if(state case final AddressesLoaded s) {
    emit(s.copyWith(searchTerm: event.term));
}
```

lib/new-ui/viewmodels/addresses/addresses_bloc.dart
```diff
@@ -0,0 +1,232 @@
...
+      return;
+    }
+    if (!added) {
+      _emitFailure(emit, initial.walletId);
+      return;
+    }
+    _refreshGroups(emit, initial.walletId);
+  }
+
+  void _emitFailure(Emitter<AddressesState> emit, String expectedWalletId) {
```
malik1004x: The `failureCode` seems to only be used for single-fire error messages. As stated above, please use `emitPresentation` – then you can get rid of the failure codes and make proper classes for them instead, you can then emit them independently of the rest of the state and get rid of the `clearFailureCode` constructor parameter.

lib/new-ui/viewmodels/addresses/addresses_state.dart
```diff
@@ -0,0 +1,138 @@
...
+    this.isSaving = false,
+    this.failureCode,
+  });
+
+  final List<AddressGroup> groups;
+  final String activeAddress;
+  final String searchTerm;
+  final bool showHidden;
+  final bool hasAccounts;
+  final String walletId;
```
malik1004x, line 32: The state object has a lot of variables that never change during the lifecycle of the flow: `walletType`, `walletName`, `accountLabel`, `walletId`, `hasAccounts`, `hasHiddenAddresses`, `canSetLabel`, `canHide`, `showAddManualAddresses`. Since their value never changes, these should be removed from the state and instead read by the UI from either getters in the Bloc or in `AddressService`. Keeping them in the state makes this object cluttered and harder to read, as well as making the origin of the values more obscure – you have to check where each is assigned and if there are any further assignments instead of just doing "Go to declaration" on the getter. Without these, this state could be substantially shorter and the code much more readable.

lib/new-ui/viewmodels/addresses/addresses_state.dart
```diff
@@ -0,0 +1,138 @@
+part of "addresses_bloc.dart";
+
+sealed class AddressesState extends Equatable {
```
malik1004x: I don't see the advantage of `Equatable` here – it only functions if the state is emitted as 1:1 the same as the previous state, which seems to never be done in this particular flow.

lib/new-ui/viewmodels/addresses/addresses_state.dart
```diff
@@ -0,0 +1,138 @@
...
+
+  final AddressesFailureCode code;
+
+  @override
+  List<Object?> get props => [code];
+}
+
+enum AddressesFailureCode {
+  addressListUnavailable,
+  saveFailed,
```
malik1004x: In relation to my previous comment on `emitPresentation` – If `saveFailed` becomes a presentation event (as it's single-fire anyway) then the generic `addressListUnavailable` could just become the `AddressesFailure` class – there's no advantage of having an enum that only ever has one value, and you could instead put the exception returned by whatever threw into the state to display an actually meaningful error.

lib/new-ui/viewmodels/receive/receive_bloc.dart
```diff
@@ -0,0 +1,660 @@
...
+    on<AddressTypeSelected>(_onAddressTypeSelected, transformer: sequential());
+    on<AddressRotated>(_onAddressRotated, transformer: droppable());
+    on<LabelSubmitted>(_onLabelSubmitted, transformer: sequential());
+    on<InfoboxDismissed>(_onInfoboxDismissed, transformer: droppable());
+    on<AddressesPageClosed>(_onAddressesPageClosed, transformer: sequential());
+    on<_WalletChanged>(_onWalletChanged, transformer: restartable());
+    on<_FiatRateChanged>(_onFiatRateChanged, transformer: sequential());
+    on<_PayjoinEndpointChanged>(_onPayjoinEndpointChanged, transformer: sequential());
+
+    _walletSub = activeWalletService.walletChanges.listen((_) {
```
malik1004x, line 51: Same as above – the wallet never changes during the Receive flow.

lib/new-ui/widgets/hardware_wallet/sync_key_images_sheet.dart
```diff
@@ -56,6 +56,7 @@ class _HardwareWalletProceedOnDeviceSheetState extends State<SyncKeyImagesSheet>
                         Icons.close,
```
malik1004x, line 56: I believe you may have reformatted this file by mistake, could you revert?

lib/new-ui/widgets/receive_page/receive_label_modal.dart
```diff
@@ -1,154 +1,162 @@
...
+    super.dispose();
+  }
+
+  Future<void> _submit() async {
+    if (_saving) {
+      return;
+    }
+    setState(() => _saving = true);
+    try {
+      await widget.onSubmit(_controller.text);
```
malik1004x: The `onSubmit` you pass is just `bloc.add` which is a synchronous action. `Navigator.pop` is also synchronous. Therefore, the time between saving being set to true and it being saved to false is zero, so the UI changes never render. You can safely remove `_saving` and all associated logic.

lib/new-ui/viewmodels/receive/receive_bloc.dart
```diff
@@ -0,0 +1,660 @@
...
+    add(ReceiveOpened(initialToken: initialToken));
+  }
+
+  final AddressService addressService;
+  final FiatRateService fiatRateService;
+  final ActiveWalletService activeWalletService;
+
+  late final StreamSubscription<WalletBase> _walletSub;
+  late final StreamSubscription<FiatCurrency> _rateSub;
+  late final StreamSubscription<String?> _payjoinSub;
```
malik1004x, line 77: The Payjoin endpoint cannot change once it's been initialized – therefore, you could just a `bitcoin!.getPayjoinEndpoint` call into `onOpened`. This way, you can get rid of this subscription and the stream in `AddressService` itself. Saves on lines and on complexity.

lib/new-ui/viewmodels/receive/receive_bloc.dart
```diff
@@ -0,0 +1,660 @@
...
+
+  @override
+  Future<void> close() async {
+    await _walletSub.cancel();
+    await _rateSub.cancel();
+    await _payjoinSub.cancel();
+    return super.close();
+  }
+
+  Future<void> _onOpened(ReceiveOpened event, Emitter<ReceiveState> emit) async {
```
malik1004x: [minor] Please call the event `Init` and the handler `_init` for parity with other flows.

lib/new-ui/viewmodels/receive/receive_bloc.dart
```diff
@@ -0,0 +1,660 @@
...
+        return;
+      }
+      requestedAmount = fiatRateService.convert(amount, receiveCrypto);
+      if (requestedAmount != null) {
+        fiatEquivalent = amount;
+      } else {
+        rateUnavailable = true;
+      }
+    } else {
+      requestedAmount = amount.currency == CryptoCurrency.btcln
```
malik1004x: [minor] Can you handle this in `FiatRateService` instead, preferrably in a genericized way?

lib/new-ui/viewmodels/receive/receive_bloc.dart
```diff
@@ -0,0 +1,660 @@
...
+    if (amount == null) {
+      requestedAmount = null;
+      fiatEquivalent = null;
+    } else if (amount.currency is FiatCurrency) {
+      final fiatCurrency = amount.currency as FiatCurrency;
+      await fiatRateService.ensureRateFor(receiveCrypto, fiatCurrency);
+      if (isClosed) {
+        return;
+      }
+      if (state case final ReceiveLoaded current when current.walletId != initial.walletId) {
```
malik1004x, line 132: How can this case happen?

lib/new-ui/viewmodels/receive/receive_bloc.dart
```diff
@@ -0,0 +1,660 @@
...
+          ? Money(amount.amount, CryptoCurrency.btc)
+          : amount;
+      fiatEquivalent = fiatRateService.convert(requestedAmount, fiatRateService.currentFiat);
+    }
+
+    if (rateUnavailable) {
+      if (state case final ReceiveLoaded loaded when loaded.walletId == initial.walletId) {
+        emit(
+          loaded.copyWith(
+            clearRequestedAmount: true,
```
malik1004x: [minor] To avoid the `clearSomething` parameters, you could have `copyWith` take in a function pointer to the value instead of take in the value directly.

lib/new-ui/viewmodels/receive/receive_event.dart
```diff
@@ -0,0 +1,67 @@
...
+final class AmountChanged extends ReceiveEvent {
+  const AmountChanged(this.amount);
+
+  final Money? amount;
+}
+
+final class InputCurrencySelected extends ReceiveEvent {
+  const InputCurrencySelected(this.currency);
+
+  final Currency currency;
```
malik1004x, line 20: The pattern you follow here is the same pattern the old ViewModel followed – there's a "token currency" variable that is nullable and null if the selected currency is also the wallet currency – and also an "input currency". I believe in hindsight this isn't the best possible pattern for this page – what we could do instead is just have a non-nullable `cryptoCurrency` and a nullable `fiatCurrency` – if the fiat is null, amounts entered are in crypto. I believe this is worth considering to make the state management clearer.

lib/new-ui/viewmodels/receive/receive_state.dart
```diff
@@ -0,0 +1,201 @@
...
+
+  final PaymentURI paymentUri;
+
+  final bool isSilentPayments;
+  final bool isLightning;
+  final AutoGenerateSubaddressStatus autoGenerateSubaddressStatus;
+  final bool isZCashTransparent;
+  final bool inputUsesSats;
+
+  final String walletId;
```
malik1004x, line 60: Same as above, I believe there is a lot of variables in the state that could be replaced with static getters in the bloc. Worth considering to remove clutter from the state classes.

lib/new-ui/viewmodels/receive/receive_state.dart
```diff
@@ -0,0 +1,201 @@
...
+final class ReceiveFailure extends ReceiveState {
+  const ReceiveFailure(this.code);
+
+  final ReceiveFailureCode code;
+
+  @override
+  List<Object?> get props => [code];
+}
+
+enum ReceiveFailureCode {
```
malik1004x: Same as above, all of theseexcept `addressListUnavailable` are de-facto single-fire events that could be replaced with `emitPresentation`. I also noticed you have a `switch` for these to convert them to a displayable error string – if you did them as separate presentation event classes, that could just be an override in each of the subclasses.

malik1004x (changes requested):

lib/core/active_wallet_service.dart
Blazebrain: The wallet can actually change with the receive page open. If you're on Receive and tap a payment link from another app (browser, chat), Cake comes back to the front and pushes the Send page on top of it, and AnyPay can switch to another wallet from there. If the user backs out of Send, they land back on the receive page, now on the new wallet.

Tested it on a build, opened Receive on a BTC wallet, fired an `ethereum:` link, switched to the ETH wallet, closed Send, and the receive page was showing the ETH address.

That's why the blocs follow the active wallet, same as the old `WalletAddressListViewModel `did through `WalletChangeListenerViewModel`, and same as send and swap. If we pass the WalletBase in directly, that page would keep showing the BTC QR after the switch

lib/core/address_service.dart
Blazebrain: Yepp, it would be cleaner that way. It touches every `cw_` package though, about 10 of them, so I'd rather do it in a follow up PR than grow this one further. Would create a ticket for it, and move the other wallet side suggestions from this file in there too so they land as one

lib/core/address_service.dart
Blazebrain: That's fair actually, but it's not only about addresses, the multichain notice uses it too. I'll move it into the receive bloc since that's the only page that reads or dismisses it, a separate class for one flag feels like an overkill

lib/core/address_service.dart
Blazebrain: Okay, I'll pass ActiveWalletService in directly. It still has to follow the active wallet though, same reason as the first comment

lib/core/address_service.dart
Blazebrain: Ohh, that definitely works better. `BaseUnit.useBaseUnitOf` seems to have been merged to dev recently, after this PR was created. Either way, I'll update and use it in the amount modal and drop the entire chaining

lib/core/address_service.dart
Blazebrain: They are left overs from moving the old logic. But yeah, I'll remove both

lib/core/address_service.dart
Blazebrain: This came over from the old viewmodel, dev does the same thing. The 1000+ aren't addresses the user generated, they're the pregenerated MWEB ones, so we show up to the last one with a transaction plus 20 after it. The list is already a lazy `SliverList`, so it's not for performance, it's so we don't dump 1000 unused addresses on the user

lib/core/address_service.dart
Blazebrain: Yeah, that works better. I'll do it together with the other changes in the follow up

lib/core/address_service.dart
Blazebrain: The identical check isn't protecting anything since the assignment writes to the held wallet anyway, so would drop it now. The `setAsActive` part needs the generation to be wallet side first, so that one goes with the follow up

lib/core/address_service.dart
Blazebrain: No, we don't need to. The electrum `generateNewAddress` already returns the new record, the proxy just throws it away, and decred has the address right there too. I'll return it through the proxy for both.

Monero's `addSubaddress` returns void, so that one stays unless you're okay with touching `cw_monero` in this PR

lib/core/address_service.dart
Blazebrain: I agree, same as the first one. It follows the same branching the old viewmodel had, so would move it wallet side in the follow up with the rest

lib/core/address_service.dart
Blazebrain: Both callers already have the AddressEntry, so yeah, would pass it in and drop the loop

lib/core/address_service.dart
Blazebrain: Okayy, would make it non-nullable now and return the wallet's default option. Moving it to `WalletBase` would go with the wallet side follow up, so these move together instead of half now and half later

lib/core/address_service.dart
Blazebrain: I like that, the wallet wide side effects are actually messy part. Right now the selected type is stored on the wallet, and swap and buy both read from it. They already skip silent payments and lightning, but Taproot or Legacy goes straight through as the swap address.

I'll prefer to do that in a separate PR though, since it changes the swap and buy flows too. For this one I'd reset it on open like dev does

lib/core/fiat_rate_service.dart
Blazebrain: Okay, I'll update that. I think we could use a `(CryptoCurrency, FiatCurrency)` record as the key though, so we can still get a single map without adding a new class

lib/core/fiat_rate_service.dart
Blazebrain: Yeah, they overlap. `FiatConversionStore` is read in a lot of files though, with reactions on it too, so merging them is bigger than this PR. Would be better if done separately

lib/new-ui/pages/addresses_page.dart
Blazebrain: Okayy, would update with the latest dev merge then

lib/new-ui/pages/addresses_page.dart
Blazebrain: I tested this on a build and selection works from the hidden page. Both blocs go through the same AddressService and write to the wallet, and when the hidden page closes the main page refreshes. With auto generate off, picking a hidden address there shows it on the receive page straight away.

What you probably saw is with auto generate on, the wallet's address getter skips hidden addresses, so the pick doesn't stick. That's wallet side, dev behaves the same since it goes through the same setter.

On the `BlocProvider`, I'll inject the bloc through the constructor like the charts page, but I'd keep the `BlocProvider` since it closes the bloc on dispose

lib/new-ui/pages/receive_page.dart
Blazebrain: Yeah, the split is just for the BlocProvider. I'll inject the bloc through the constructor, and keep BlocProvider there so it gets closed on dispose.

For the rename, the old ReceivePage got removed in this PR, and since the old UI removal PR merged in, there's no need for a New prefix anymore, so I dropped it. It's still on the rest though, so we can drop it on those in a separate PR

lib/new-ui/viewmodels/addresses/addresses_bloc.dart
Blazebrain: Same as the first comment, the wallet can change from the AnyPay switch while this is open. Tested it on a build, the receive page rebuilt for the new wallet, and the addresses bloc handles it the same way

lib/new-ui/viewmodels/addresses/addresses_state.dart
Blazebrain: Most of these change when the wallet changes (walletType, walletName, walletId, hasAccounts, canSetLabel, canHide), and that can happen with the page open like I mentioned on the first comment. Keeping them in the state means they update in the same emit as the address list, a getter could read the new wallet while the list still belongs to the old one.

`hasHiddenAddresses` changes on hide and unhide, but I can derive that one from groups instead of storing it. And `accountLabel` doesn't change anymore now that the account switching is gone, so that one can be a getter

lib/new-ui/viewmodels/addresses/addresses_state.dart
Blazebrain: Thta fair enough, `AddressGroup` doesn't have value equality so it never dedups on the addresses state, would remove it there. I'll keep the receive one though as the rate autorun fires on every price update for any currency, and `Equatable` skips the rebuild when nothing on the page changed

lib/new-ui/viewmodels/receive/receive_bloc.dart
Blazebrain: Same as the first one, the AnyPay switch can happen with receive open. That's the exact case I tested

lib/new-ui/viewmodels/receive/receive_bloc.dart
Blazebrain: It does change actually. It's null until the first `initPayjoin` finishes, which needs the network, and `newPayjoinReceiver` sets a new one once a sender starts a payjoin, which is while the QR is on screen. The re-init on sync keeps the same receiver though, so that one doesn't change it.

Tested it on a build, created a wallet offline, opened receive, then turned the network back on. The endpoint came in about 30s later, and the page didn't pick it up. The reaction goes through `getPayjoinEndpoint`, which returns early when the wallet has no spendable coins, so it never tracks the endpoint. Would fix that to read the endpoint directly. Reading it once in `onOpened` would have the same problem though

lib/new-ui/viewmodels/receive/receive_bloc.dart
Blazebrain: It's for a wallet switch landing while the rate fetch is still in flight, the AnyPay case from the first comment. Without it the old wallet's rate result could land on the new wallet's state

lib/new-ui/viewmodels/receive/receive_state.dart
Blazebrain: Same as the addresses state, most of these change with the wallet so they stay in the state. `autoGenerateSubaddressStatus` can be a getter, and `inputUsesSats` goes away with the `useBaseUnitOf` change from earlier

lib/new-ui/viewmodels/receive/receive_bloc.dart
Blazebrain: Okayy, I'll do this after the currency change below and the presentation events, those two remove most of the clear flags. Would look at the remaining ones after

lib/new-ui/viewmodels/addresses/addresses_state.dart
Blazebrain: Yeahh, with `saveFailed` as a presentation event the enum only has `addressListUnavailable` left, so would make that the failure state itself. I'd keep the message coming from `S` though, not the raw exception, so users don't get technical errors on screen

lib/core/active_wallet_service.dart
malik1004x: That's an obscure edge case. It's not worth basing the entire design around an issue nobody has ever reported or cared about. You can just make opening a payment link close this page and not overcomplicate the flow. If the user is still interested in the QR code, it's one tap away.

lib/core/address_service.dart
malik1004x: This _is_ the refactor PR – those extra changes to the wallet-side code would still be easier to review than what you currently have. Why create workarounds for further refactors that may or may not land sometime in the future? All the other refactors currently open also change wallet-side code in the capacity it was necessary to do the refactor properly, this was also what we agreed on when settling on a pattern for the refactors – while refactoring the VM layer, also refactor the part of the wallet-side code necessary to do the refactor properly.

lib/core/address_service.dart
malik1004x: Yes, please modify wallet-side code in this PR, as agreed to by the entire team when planning the refactors.

lib/core/address_service.dart
malik1004x: Well, it does touch swap and buy, but in a bad way – we always want a single "default" address type for those. So you'll be also fixing a couple potential bugs.

lib/core/fiat_rate_service.dart
malik1004x: You could still keep the old methods of `FiatConversionStore` and wire them into the new logic for backwards compatibility – this way, you only get maybe 10 lines of extra diff.

lib/core/active_wallet_service.dart
malik1004x: I spoke with @OmarHatem28 and we agree that doing workarounds for every function for this one edge case is not the correct approach – please just make it pop the Receive page if a QR is received from a deep link. Then, you can remove every workaround done for this case.

OmarHatem28: yup, just make the navigator.push in the linkviewmodel to be navigator.popuntilandpush and it should pop to the dashboard regardless of what screen is on

PR #3461, opened 2026-07-31
Add tbb to solana default tokens

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/src/screens/dashboard/edit_token_page.dart
```diff
@@ -210,28 +222,28 @@ class _EditTokenPageBodyState extends State<EditTokenPageBody> {
...
+                                  ),
                               );
                               return;
                             }
 
                             final isWhitelisted = await widget.homeSettingsViewModel
                                 .checkIfTokenIsWhitelisted(_contractAddressController.text);
 
-                            final hasPotentialError = !isWhitelisted &&
+                            final isTrusted = isWhitelisted || _isJupiterVerified;
```
OmarHatem28: I believe we should also update _isJupiterVerified here.
and see if we need to update the UI to mark it as unverified if it was marked as verified on a previous entry of a mint address

Blazebrain: I don't think we should mark any token as unverified, because jupiter gives us three states, either verified, not verified, or nothing at all (when the mint is not indexed by Jupiter, or maybe the call failed). Marking the token as unverified would wrongly flag legit tokens that Jupiter doesn't index, and every token whenever the lookup fails. We should show the badge when its positive only, as we currently do

Also, the `_isJupiterVerified` flag is already reset whenever a new address check runs, and any responses or data for an old address is dropped. But I'll add a variable to track which address the verification result belongs to, so it always matches correctly and never stale at any point, either the check has run or not

OmarHatem28 (approved):

Issue #3406, opened 2026-07-15
Blind users cannot use this app because it is not accessible with native screen readers such as VoiceOver.


AnonymousAccessibilityAdvocate:
All buttons in Cake Wallet need to be labeled. Open Cake wallet, navigate with the native iOS screenreader (VoiceOver) and you'd expect to hear, "send", "receive", "swap," and so on. Instead the screenreader only reads, "button", "button", "button". This affects all iOSversions.

MrCyjaneK:
#3402

AnonymousAccessibilityAdvocate:
I think that the buttons need to be wrappedin a Semantics widget and setting the label property, or by using the semanticLabel property directly on widgets like.

Issue #3402, opened 2026-07-14
Make Cake wallet and Cupcake VoiceOver accessible


rottenwheel:
**Is your feature request related to a problem? Please describe.**
A user asked in Matrix room whether developers could make both applications voiceover accessible. I know there's been some discussion and progress with these, but I figured an issue could come in handy, to track and discuss, precisely.

**Describe the solution you'd like**
At least, core features, such as balance, send, receive, transaction details should be voiceover accessible, for people who are blind, or have difficulty seeing things.

rottenwheel:
cc. @shortwavesurfer2009.

shortwavesurfer2009:
Cake wallet on Android is accessible, so cake wallet on iOS ought to be, or at least decently close, I would think.

I have not played with Cupcake on either platform.

AnonymousAccessibilityAdvocate:
I have emailed Cakewallet about this issue and was told that it would be passed on to the developers. I hope this is addressed. Blind people should be able to enjoy the benefits of Cakewallet. With only minor tweaks needed to label elements for the screenreader, then I hope this issue is resolved quickly.

PR #3465, opened 2026-07-31
a11y: decorative/meaningful image semantics, labeled icon buttons, audible addresses

sethforprivacy:

# Description

Part 3 of the screen-reader accessibility remediation (VoiceOver/TalkBack): image and compact-icon primitives.

**Changes**
- `CakeImageWidget` (186 call sites): new optional `semanticsLabel`, honored in all four render branches (precompiled vector asset, raster asset, network SVG, network raster). `null` — the default — now means genuinely decorative: the image contributes **no** accessibility node instead of an unnamed `image` node. Built-in loading spinners and the error placeholder no longer create unlabeled focus stops; caller-supplied `errorWidget`s keep their own semantics (several call sites render meaningful initials there).
- `FloatingIconButton`: new **required** `semanticLabel`, exposed as one labeled button node. All 6 call sites migrated (paste ×4, scan QR, address book) using existing ARB keys.
- `RoundedCheckbox`: previously the unchecked state rendered `Offstage()` — selection state was completely invisible to screen readers. It now always exposes `checked:` state (no visual change), with an `excludeFromSemantics` escape hatch for rows that announce their own selection (used by the receive address-type list PR in this series).
- Receive QR code: the QR block is now a single labeled button node ("QR code for your receive address", hint = existing `qr_fullscreen` key), with the QR matrix, embedded logo, badge and wordmark excluded; the payjoin badge merges icon+text into one node.
- `AddressFormatter`: both segmented-address `RichText` builders now carry the **full** address as their semantics label, so screen readers read the actual address instead of 4-character pseudo-words ("beesh onekay…") or "dot dot dot" for the truncation ellipsis. This is the security-critical fix that makes destination-address verification by ear possible.
- 1 new English-only ARB key: `qr_code_receive_address`.

**Notes for reviewers**
- Screen readers will read the address label as a continuous string; a character-by-character "spell out" affordance is a possible future UX improvement, but a correct continuous read is a strict improvement over chunk pseudo-words.
- FloatingIconButton's 30×30 hit target remains below the 44/48pt platform minimums — deliberately not changed here to keep this PR visual-change-free; flagged for a follow-up.

**Verification**
- `flutter analyze` (Flutter 3.41.9, same as CI) — zero new issues vs `dev` baseline. flutter_svg 2.2.x API names verified against the resolved package.
- No visual or pointer-behavior changes; all `ValueKey`s preserved.

# Pull Request - Checklist

- [x] Format code

### DIFF
--- a/lib/src/screens/receive/widgets/qr_image.dart
+++ b/lib/src/screens/receive/widgets/qr_image.dart
@@ -1,3 +1,4 @@
+import 'package:cake_wallet/generated/i18n.dart';
 import 'package:cake_wallet/new-ui/widgets/coins_page/token_image_widget.dart';
 import 'package:cake_wallet/src/widgets/cake_image_widget.dart';
 import 'package:flutter/material.dart';
@@ -13,6 +14,7 @@ class QrImage extends StatelessWidget {
     this.errorCorrectionLevel = qr.QrErrorCorrectLevel.H,
     this.embeddedImagePath,
     this.badgeImageOnEmbeddedImagePath,
+    this.semanticsLabel,
   });
 
   final double? size;
@@ -23,6 +25,11 @@ class QrImage extends StatelessWidget {
   final int errorCorrectionLevel;
   final String? embeddedImagePath;
   final String? badgeImageOnEmbeddedImagePath;
+
+  /// Accessible name announced for the code itself. Defaults to a generic
+  /// localized "QR code" so the code is always discoverable; pass a more
+  /// specific label when the host screen knows what the code encodes.
+  final String? semanticsLabel;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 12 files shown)


sethforprivacy:
Heads-up from the test PR (#3469): the `RoundedCheckbox` fix here is only half-effective as-is. `Semantics(checked: false, child: Offstage())` produces a zero-rect node, and Flutter drops zero-rect nodes from the semantics tree — so *unselected* options still announce no state. The clean fix (give the unchecked branch the checked indicator's 20×20 footprint, which also removes the selection layout-jump) changes visible layout in two pickers, so it was deliberately left out pending a design call. #3469 carries a skipped test asserting the desired behavior. In the receive address-type list this is masked by the row-level `selected` semantics from #3467; the legacy `present_receive_option_picker` call site is where it shows.

lib/src/screens/receive/widgets/qr_image.dart
```diff
@@ -57,20 +57,25 @@ class QrImage extends StatelessWidget {
...
-          foregroundColor: foregroundColor,
-          backgroundColor: backgroundColor,
-          padding: const EdgeInsets.all(12.0),
-        ),
-        centerImageToUse,
-      ],
+    // The QR matrix, the embedded logo and its badge are all purely visual: the
+    // data they encode is announced by the labelled control hosting this widget,
+    // so nothing in here should become a focus stop of its own.
+    return ExcludeSemantics(
```
OmarHatem28: we shouldn't exclude semantics here, we should still say that there is a qr here (which is already handled in the QR package, and also because tapping it opens the full screen, so the user should know that there is a QR on this screen

OmarHatem28: @sethforprivacy

sethforprivacy: Good call — fixed in c360bd6.

`QrImage` no longer excludes its subtree. `QrImageView` keeps its own semantics node and now gets a *localized* label (`scan_qr_code` → "QR code", already translated in every locale) instead of the package's hardcoded English `'qr code'`, plus an optional `semanticsLabel` override for hosts that know what the code encodes. Only the embedded logo and its badge stay excluded — those are genuinely decorative.

You were right that this was the real bug for the nine other call sites (fullscreen QR, node share, anonpay, 2FA setup, exchange/swap external send, UR): they lost the code entirely with nothing announcing a QR was on screen.

On the receive page the QR now carries `qr_code_receive_address` and the subtree is wrapped in `MergeSemantics`, so the QR, the payjoin badge and the fullscreen tap target read as one button node — "QR code for your receive address, Payjoin enabled, button. Tap to open full screen QR code." — rather than three focus stops. The whitespace churn in `receive_qr_code.dart` is just re-indentation from adding that wrapper.

OmarHatem28 (approved):

PR #3466, opened 2026-07-31
a11y: send flow and QR scanner semantics (fields, errors, status announcements)

sethforprivacy:

# Description

Part 4 of the screen-reader accessibility remediation (VoiceOver/TalkBack): the send flow and QR scanner.

**Changes**
- **Text fields** (send address, amount, memo, scanner manual entry): each is now a single semantics node announcing a persistent localized purpose plus its current value — previously the only "name" was a placeholder like `0.00` that vanished on input. No visual changes (the visible label Texts rendered by `send_page.dart` are excluded to avoid double announcement).
- **Validation errors**: address/amount errors are polite live regions labeled with the field name, so newly appearing errors are actually announced.
- **Controls**: asset/currency picker, max-balance chip, recipient dots ("Recipient N", selected state), scanner help button (icon-only, had literally no name), manual-input button, dropdown rows (expanded/collapsed state), confirm-sheet save-contact/add-note buttons — all real labeled buttons now.
- **Status announcements**: send progress (generating/signing/sending), the transaction-error banner, and multi-part animated-QR scan progress ("Scanned 3 of 5 QR code parts") are announced once as polite live regions; transaction success fires a single announcement. The sync chip exposes percentage as a non-live value so progress ticks don't interrupt.
- `PrimaryButton`/`LoadingPrimaryButton` (shared): keep their accessible name while the spinner replaces the visible label, and report disabled state honestly.
- **De-noised**: address overlay duplicate, memo counter, "Max." caption, scanner full-screen tap-catcher, reticle, and the confirm sheet's off-screen page (mounted at `heightFactor: 0`) no longer produce focus stops.
- **Localization fixes found in passing**: hard-coded "Synchronizing...", "Coin Control", and `'ok'` (scan page + `qr_scanner.dart`) now use ARB keys.
- 4 new English-only ARB keys: `opens_in_browser`, `qr_parts_scanned`, `recipient_number`, `show_full_address`.

**Notes for reviewers**
- Fees row: `onTap` is now `null` when there are no fee priorities (it was a silent no-op); the only behavior delta is that the ink ripple no longer fires on a dead tap.
- "Coin Control" now uses the existing `coin_control` key whose value is "Coin Control (optional)" — matching the legacy send card. If the exact old string is wanted, a new key is needed instead.
- Empty memo/manual-entry fields read their name twice (hint text equals the semantics label); chosen deliberately so a *filled* field always has a name.

**Verification**
- `flutter analyze` (Flutter 3.41.9, same as CI) — zero new issues vs `dev` baseline; ARB validated with `jq`.
- No visual or pointer-behavior changes intended; all `ValueKey`s preserved.

**Merge note:** independent of the sibling a11y PRs at compile level; touches hunks adjacent to `a11y/modern-button` in `scan_page.dart`/`send_page.dart`/`fiat_amount_bar.dart`, so whichever merges second may need a trivial rebase.

# Pull Request - Checklist

- [x] Format code


sethforprivacy:
**On-device audit finding (P1) — hold merge, fix incoming.** The automated accessibility-tree audit (cake-tech/cake-qa-trails#6) found that this PR's amount-field change makes the field vanish from the Android accessibility tree entirely. A/B on real builds: without this branch the send screen exposes `"Amount:"` + `EditText "0.00"`; with it there is **no node for the amount field at all** (verified not offscreen/hidden), while the visible `Amount:` caption is also `ExcludeSemantics`'d on the assumption the field announces itself. Net effect: a screen-reader user cannot find the amount field. The Dart looks correct (`MergeSemantics > Semantics(label:) > TextField`) — the failure only shows in the platform tree, and curiously the address field treated similarly survives. Root-cause investigation + fix with a device-shape widget test is running now; a fix commit will land on this branch shortly.

sethforprivacy:
Fix landed (`26bc513f2`). Root cause was subtle and worth recording: `FormField` wraps its builder in its own **non-container** `Semantics(validationResult:)` annotation, and the currency-picker `Semantics` added here — which also lacked `container: true` — was absorbed into that wrapper. The whole FormField collapsed into a single Button node (label "Select Asset") with the amount `EditText` buried inside it, which Android then surfaced as one tappable button and no text field. One-line fix: `container: true` on the picker Semantics (with a comment explaining it's load-bearing). The address field only survived because its inner semantics all happened to be boundaries.

The caption `ExcludeSemantics` stays — with the node restored, the premise ('the field announces this label itself') is now provably true, and 14 new widget tests in #3469 pin it (the key one walks the ancestor chain and failed pre-fix). Note for future work: seven other `excludeSemantics: true` in this PR lack `container: true` (dropdown_row, network_list, fiat_amount_bar ×2, recipient_dot_row, scan_page, send_confirm_sheet); none misbehaves today per the device audit, but each is one absorbing ancestor away from this bug class — flagged as follow-up hardening rather than churned here. Device re-audit against the rebuilt APK will confirm closure.

sethforprivacy:
Follow-on fix landed (`bfa935e29`) for the duplicate announcement the closure audit found. Diagnosis: the duplication is **platform-side** — the framework tree post-fix has exactly one `Amount`-bearing node (verified by walking even merged-away semantics nodes), but `FormField` wraps its builder in an unavoidable wrapper node, and Android re-describes that wrapper from its descendants (hence the phantom `ImageView "Amount: 0.00"`). Since Flutter offers no `labelledBy`, a programmatic field label and a single announcement are mutually exclusive under a `FormField`. The fix takes the shape the device A/B already verified as good: the visible "Amount:" caption is back in the semantics tree and the field wrapper label is gone — one caption stop, one field stop, adjacent, no image role. The `container: true` fix stays (it's what restored the field node). Widget tests pin the shape with verified regression power (re-introducing the old wrapper fails 3 of the new tests). The address field is deliberately untouched — it reads as a single correct node on device — with the address/amount inconsistency flagged for the next device pass.

sethforprivacy:
**Final on-device verification: closed.** Send screen on the rebuilt APK (`83a08a1d2`) and the integration build:

```
"Amount:"                  <- caption stop
[x313] EditText "0.00"     <- the field
  [m886] Button "Paste"
  "BTC"
```

Duplicate announcement gone, no phantom ImageView, amount-row Paste named (via #3465 on integration). One deliberate design note for reviewers: the field's accessible name is the **adjacent caption**, not a field-carried label — the field-carried variant provably triggers platform-side duplication under `FormField`'s wrapper (Flutter's Android bridge has no labeledBy association). This matches the app's pre-series behavior; the only degraded case is navigate-by-control-type, which hears "0.00, edit box". The address field on the same screen carries its label inline and doesn't duplicate (different child-boundary structure) — that inconsistency is flagged as a follow-up design item rather than unified here, to avoid re-risking the duplication. Overrule welcome if you'd prefer a unification attempt.

OmarHatem28 (approved):

PR #3468, opened 2026-07-31
a11y: dashboard and navigation semantics (navbar, cards, sync, switchers)

sethforprivacy:

# Description

Part 6 of the screen-reader accessibility remediation (VoiceOver/TalkBack): the default dashboard and navigation.

**Changes**
- **Bottom navigation** (`NewMainNavBar`): each tab is now a labeled button with `selected` state in a mutually exclusive group (labels reuse the existing `home`/`wallets`/`contacts`/`apps` keys already wired into `NewMainActions`); the animated pill's visible text no longer creates a duplicate stop. Mirrors the legacy `navigation_dock.dart` pattern.
- **Bitcoin ⇄ Lightning switcher**: was an unlabeled `InkWell` whose state existed only as knob position — now a single toggle node with label and `toggled` state.
- **Assets/History tab switcher**: tabs expose button/selected/mutually-exclusive semantics.
- **Sync bar**: one button node whose label is the already-localized sync status plus any active Tor/MWEB/Silent-Payments badge names, with "manage nodes" as the hint; in compact mode the pulsing dot (previously the *only* syncing indicator) gets a "Synchronizing" text equivalent; the chain icon's progress ring exposes a semantic value.
- **Wallet row**: the row, hardware-wallet glyph, and inner accounts button collapsed into one labeled control ("<wallet name>, Hardware wallet" when applicable, hint "Wallet Accounts") instead of two nested unlabeled targets.
- **Balance cards**: labeled selectable nodes; the undiscoverable long-press show/hide-balance gets a semantic long-press action with the existing "Long press card to show balance" string as hint; balances hidden at opacity 0 no longer leak through semantics; card art/corner icons excluded; 3-dots customizer labeled ("Menu"); deposit/withdraw/buy are real buttons.
- **Assets header**: the `Opacity(0)` phantom tappable is now conditionally built (visuals identical) — previously an invisible, focusable no-op target.
- **History/asset rows**: all row types are buttons; leading direction glyphs excluded (row titles already announce sent/received/pending via `formattedTitle`).
- **Misc**: MWEB promo row + its "Don't show anymore" are buttons; account customizer add/select/randomize-name controls labeled; apps cards get a close tooltip and an "opens externally" hint; the invisible pointer-absorbing strip behind the navbar is excluded from the a11y tree.
- 3 new English-only ARB keys: `hardware_wallet`, `lightning_mode`, `opens_externally`.

**Notes for reviewers**
- The audit expected history rows to need direction labels, but `TransactionListItem.formattedTitle` already returns localized received/sent/pending titles — so glyphs were excluded rather than double-labeling. Please confirm that reads correctly on device.
- Balance cards are container nodes (label + selected) rather than fully merged, because a card contains its own buttons.
- Card label falls back to "Balance" for wallets without named accounts; product may prefer something else.
- `ReorderableListView` account reordering still has no accessible move action — follow-up ticket needed.
- The apps close button uses `tooltip:` (the only way `IconButton` gets a name), which also adds a visual long-press tooltip.

**Verification**
- `flutter analyze` (Flutter 3.41.9, same as CI) — zero new issues vs `dev` baseline; ARB validated with `jq`.
- All `ValueKey`s preserved; no visual or pointer-behavior changes beyond the tooltip noted above.

# Pull Request - Checklist

- [x] Format code


lib/new-ui/widgets/coins_page/assets_history/assets_top_bar.dart
```diff
@@ -49,40 +51,45 @@ class AssetsTopBar extends StatelessWidget {
...
-                                  style: TextStyle(color: Theme.of(context).colorScheme.primary),
-                                ),
-                              CakeImageWidget(
-                                  imageUrl: tabs[selectedTab].actionButton?.iconPath,
-                                  colorFilter: ColorFilter.mode(
-                                      Theme.of(context).colorScheme.primary, BlendMode.srcIn)),
-                            ],
+                  // Built conditionally rather than rendered at opacity 0: an
+                  // invisible chip stayed focusable for screen readers.
+                  if (actionButton != null)
```
OmarHatem28: Having Opacity 0 still has some UI effect (padding), so it's not like not adding a widget at all.
So double-check that the UI looks correct first

lib/new-ui/pages/account_customizer.dart
```diff
@@ -172,17 +172,27 @@ class _AccountCustomizerState extends State<AccountCustomizer> {
...
-                        child: Align(
-                          alignment: Alignment.topCenter,
-                          heightFactor: _kStackVisibleFactor,
-                          child: card,
+                      // One labeled, selectable node per account; the card's own
+                      // texts stay reachable underneath it.
+                      child: Semantics(
+                        button: true,
+                        selected: isSelected,
+                        label: _items[index].accountListItem.label,
```
OmarHatem28: took me a bit to understand this, so better for readability to be like this for example

```suggestion
                    final _selectedItemIndex = _items.length - 1;

return Container(
key: ValueKey(index),
                      // One labeled, selectable node per account; the card's own
                      // texts stay reachable underneath it.
                      child: Semantics(
                        button: true,
                        selected: _selectedItemIndex == index,
                        label: _items[index].accountListItem.label,
```

lib/new-ui/widgets/coins_page/assets_history/assets_top_bar.dart
sethforprivacy: Good catch — the invisible chip did reserve 40px of height in that header row, so the bare conditional could change the header height on tabs without an action button. Fixed: `if (actionButton == null) const SizedBox(height: 40) else …` — same geometry as before by construction, still no phantom tap target (a `SizedBox` contributes nothing to the semantics tree).

lib/new-ui/pages/account_customizer.dart
sethforprivacy: Applied — now `final selectedItemIndex = _items.length - 1;` with a comment explaining the bottom-to-top stack ordering, and `selected: selectedItemIndex == index`. (Used no leading underscore since it's a local.)

OmarHatem28 (approved):
I don't think the sized box is correct, but @sethforprivacy if the UI still looks good then feel free to merge it

sethforprivacy:
> I don't think the sized box is correct, but @sethforprivacy if the UI still looks good then feel free to merge it

Testing it out now.

sethforprivacy:
In poking through Bitcoin, Monero, and Ethereum dashboards I can't see any visible changes to the UI. LGTM.

PR #3470, opened 2026-07-31
fix: unify encryption across platforms

MrCyjaneK:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_core/lib/utils/file.dart
+++ b/cw_core/lib/utils/file.dart
@@ -103,8 +103,8 @@ Future<void> _writeXChaCha20({
   );
 
   final tmpFile = File('$path.tmp');
-  await tmpFile.writeAsBytes(encrypted, flush: true);
-  await tmpFile.rename(path);
+  tmpFile.writeAsBytesSync(encrypted, flush: true);
+  tmpFile.renameSync(path);
 }
 
 Future<String> _readXChaCha20({
--- a/cw_core/lib/utils/file.dart
+++ b/cw_core/lib/utils/file.dart
@@ -1,21 +1,84 @@
+import 'dart:convert';
 import 'dart:io';
-import 'package:cw_core/key.dart';
+import 'dart:typed_data';
+
+import 'package:cake_backup/backup.dart' as cwb;
 import 'package:encrypt/encrypt.dart' as encrypt;
 
-Future<void> write({required String path, required String password, required String data}) async =>
-    writeData(path: path, password: password, data: data);
+const _ivEncodedStringLength = 12;
 
-Future<void> writeData(
-    {required String path, required String password, required String data}) async {
-  final keys = extractKeys(password);
-  final key = encrypt.Key.fromBase64(keys.first);
-  final iv = encrypt.IV.fromBase64(keys.last);
-  final encrypted = await encode(key: key, iv: iv, data: data);
-  final f = File(path);
-  f.writeAsStringSync(encrypted);
+Future<void> write({
+  required String path,
+  required String password,
+  required String data,
+  bool? highEntropyPassphrase,
+}) async {
+  await _writeXChaCha20(
+    path: path,
+    password: password,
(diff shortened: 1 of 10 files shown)


pubspec_base.yaml
```diff
@@ -189,6 +188,10 @@ dependency_overrides:
   bech32:
     git:
       url: https://github.com/cake-tech/bech32.git
+  cake_backup:
+    git:
+      url: https://github.com/cake-tech/cake_backup.git
+      ref: b5d86a2a21e1c186cd0baa4df1f3a4f3f9413056
```
OmarHatem28: migrate to pubspec_override

cw_core/lib/utils/file.dart
```diff
@@ -1,21 +1,79 @@
...
+      throw Exception('Failed to decrypt legacy file: invalid password or corrupted data');
+    }
+
+    if (_isJson(data)) {
+      await _writeXChaCha20(
+        path: path,
+        password: password,
+        data: data,
+        highEntropyPassphrase: useHighEntropy,
+      );
```
OmarHatem28, line 65: I think we should wrap this in a try/catch, so a failure here won't stop wallet from opening/read from completing

cw_core/lib/utils/file.dart
```diff
@@ -24,5 +82,57 @@ Future<String> read({required String path, required String password}) async {
...
+}) async {
+  final encrypted = await cwb.encrypt(
+    password,
+    Uint8List.fromList(utf8.encode(data)),
+    highEntropyPassphrase: highEntropyPassphrase,
+  );
+
+  final tmpFile = File('$path.tmp');
+  await tmpFile.writeAsBytes(encrypted, flush: true);
+  await tmpFile.rename(path);
```
OmarHatem28: shouldn't this be `writeAsBytesSync` and `renameSync`? or another way to ensure that 2 file writes don't conflict with each other

OmarHatem28 (changes requested):
also can we add some tests to this migration, add cases for correct migration and for failed migration, and for writing multiple times together, and reading multiple times together, etc...

cw_core/lib/utils/file.dart
MrCyjaneK: we do flush: true and await so there's no difference

cw_core/lib/utils/file.dart
MrCyjaneK: it shouldn't fail at all and if it does it better do so loudly

cw_core/lib/utils/file.dart
OmarHatem28: I am not talking about those 2 functions racing to modify the file, I am talking about 2 calls to this whole `_writeXChaCha20` function, from 2 different places, which I guess is very likely to occur specially in wallets that call save a lot (BTC, LTC)

MrCyjaneK: In that case yeah but I think that file locking should be a separate thing we work on as there are multiple other things we want to protect from potential race conditions

cw_core/lib/utils/file.dart
OmarHatem28: it will not fail loudly, it will just fail to open the wallet silently, and the user will end up with a wallet that can't be opened, and since it passed our tests, so we don't need it to fail loudly, we need it to work, and for any reason the user might have, we shouldn't block him from opening his wallet, we can add another way to check if the migration worked tho, but it will be a bit of a headache

MrCyjaneK: True, but it can only fail if
1) secure random is unavailable (impossible)
2) file wasn't written (out of space / wrong permissions)

and even if it fails a fix is a matter of reopening the app after solving the problem. Unless I'm missing something

OmarHatem28 (approved):

PR #3477, opened 2026-08-03
Integration tests

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


integration_test/funds_suites/swap_funds_test.dart
```diff
@@ -0,0 +1,73 @@
...
+    final appStore = getIt.get<AppStore>();
+    expect(appStore.wallet?.type, type);
+
+    await homePageRobot.openSwapSheet();
+    await swapRobot.isDisplayed();
+
+    await swapRobot.enterMinimumViableDepositAmount();
+    await swapRobot.confirmQuoteReceived();
+
+    await swapRobot.tapSwapButton();
```
OmarHatem28, line 71: you don't set a receive address, so the swap button should be disabled at this stage

OmarHatem28 (approved):
from a quick review I just have this comment, will wait another green light from @konstantinullrich or @malik1004x and you can merge it

integration_test/funds_suites/swap_funds_test.dart
Blazebrain: Yeah, the swap button would be disabled there as I don't set a receive address by default. It sets in the manual funds tests flow where we actually want to test the swap with funds.

Either ways, I've adjusted it to add the address here too, and do a small confirmation that the payout address is the same as what was entered

lib/new-ui/new_dashboard.dart
```diff
@@ -33,7 +33,7 @@ class NewDashboard extends StatefulWidget {
 
   final List<Widget> dashboardPageWidgets = [
     getIt.get<NewHomePage>(),
-    getIt.get<WalletListPage>(),
+    getIt.get<WalletListPage>(param1: (_) => Future<void>.value()),
```
malik1004x, line 36: It's not clear what this does, but could you handle it in the `WalletListPage` constructor instead of here?

.github/workflows/reusable-integration-test.yml
```diff
@@ -0,0 +1,734 @@
...
+          if [[ "$TEST_TIER_INPUT" != "all" ]]; then
+            search_dir="$search_dir/$TEST_TIER_INPUT"
+          fi
+          first_suite=$(find "$search_dir" -name "*_test.dart" -type f | sort | head -1)
+          echo "Priming build with $first_suite"
+          flutter build apk --debug --target-platform android-x64 --target="$first_suite" --dart-define=CI_BUILD=true
+
+      - name: Prepare emulator test script
+        run: |
+          cat > /tmp/emulator_ready.sh << 'EOF'
```
malik1004x, line 277: There is a lot of logic inside this runner, could you extract this into a separate script file?

integration_test/funds_suites/send_funds_test.dart
```diff
@@ -0,0 +1,109 @@
...
+        await homePageRobot.openSendSheet();
+        await sendRobot.isDisplayed();
+
+        await sendRobot.enterAddress(ownAddress);
+        await sendRobot.enterAmount(TestConfig.fundsSendAmountFor(type));
+        await sendRobot.tapSendButtonWhenReady();
+
+        await authFlows.authenticateWithPin(required: false);
+
+        await sendRobot.swipeToConfirm();
```
malik1004x, line 65: FYI, "swiping" isn't necessary as the app also has a "Send" button triggered if it detects the user's using a screen reader – perhaps you could simulate that in your testing harness?

malik1004x (changes requested):

integration_test/funds_suites/send_funds_test.dart
Blazebrain: Okayy, I'll trigger the accessibility check so the button shows

I'll add it as its own case though, so that we always verify both routes are fully functional, as they're both actions that our users would do to move funds

PR #3482, opened 2026-08-03
security: fail closed when the app is paused behind an in-app browser

sethforprivacy:

# Description

Fixes a reported app-lock bypass: with the in-app browser open, backgrounding
and reopening the app skips the PIN prompt entirely, even with
**Require PIN after: Always**. Authentication bypass, CWE-288.

## Root cause

Two defects in `RootState` combine:

**1. The unlock screen was only ever scheduled from `build()`.**
Flutter disables frame production while the lifecycle state is `paused`
(`SchedulerBinding.handleAppLifecycleStateChanged` → `_setFramesEnabledState(false)`).
The `setState()` in the `paused` handler therefore produced no frame, `build()`
never ran, and the unlock route was never pushed. In the ordinary case this only
worked by luck: iOS renders an app-switcher snapshot frame on the way out, which
happens to run `build()`.

**2. `resumed` cleared `_isInactive` unconditionally**, with no check that the
user had actually authenticated. Once cleared, `build()`'s guard no longer held
and the lock was simply gone.

A native in-app browser (`SFSafariViewController` on iOS, Custom Tab on Android)
pauses Flutter for as long as it is on top. I confirmed on a simulator that **no
`resumed` event is delivered at all** while the browser covers the app — so
backgrounding and returning is completely invisible to the lock logic. It only
sees `resumed` once the browser is dismissed, at which point defect 2 has
already thrown the flag away.

## Changes

- `root.dart`: track the owed authentication in a dedicated `_authRequired`
  flag, set when the app is paused while unlocked and cleared **only** by
  `_reset()` after a successful authentication. `build()` gates on that instead
  of `_isInactive` (which keeps its existing "app is backgrounded" meaning for
  the `isInactive` stream and brightness handling).
- `root.dart`: present the unlock screen from the lifecycle handler as well as
  from `build()`, so it no longer depends on a frame being rendered while
  paused. Extracted into `_presentAuthIfRequired()`; the `_postFrameCallback`
  latch keeps it single-shot.
- `root.dart`: dismiss any native in-app browser when the unlock screen is
  pushed — it is presented above the entire Flutter view and would otherwise
  keep covering the PIN screen.
- `transaction_details_view_model.dart` / `anonpay_details_view_model.dart`:
  open block explorer and anonpay status links with
  `LaunchMode.externalApplication`, so wallet-identifying pages are not
  rendered inside an app that considers itself locked. This makes them
  consistent with the trade, bridge, payjoin and order detail screens, which
  already do this.

## Verification

Built a harness on an iOS 26.5 simulator that reproduces `RootState`'s state
machine exactly, and drove the PoC sequence (open explorer link → background →
return). Instrumented lifecycle states, `framesEnabled` and every `build()`.

| Case | Result |
| --- | --- |
| Current logic, frame rendered while paused | PIN pushed, but *underneath* the browser — explorer page fully visible on return |
| Current logic, no frame while paused | No PIN at all — reopens straight into wallet content |
| Patched logic, no frame while paused | PIN screen shown |
| Patched logic + `externalApplication` (this PR) | PIN screen shown immediately on return |

Confirmed along the way: opening the in-app browser alone drives Flutter to
`inactive → hidden → paused`, and no lifecycle event whatsoever is delivered
when the app is foregrounded again while the browser is on top.

# Pull Request - Checklist

- [x] Format code

### DIFF
--- a/lib/src/screens/root/root.dart
+++ b/lib/src/screens/root/root.dart
@@ -23,6 +23,7 @@ import 'package:flutter/material.dart';
 import 'package:mobx/mobx.dart';
 import 'package:trezor_connect/trezor_connect.dart';
 import 'package:uni_links/uni_links.dart';
+import 'package:url_launcher/url_launcher.dart';
 
 class Root extends StatefulWidget {
   Root({
@@ -61,6 +62,7 @@ class RootState extends State<Root> with WidgetsBindingObserver {
       : _isInactiveController = StreamController<bool>.broadcast(),
         _isInactive = false,
         _requestAuth = true,
+        _authRequired = false,
         _postFrameCallback = false;
 
   Stream<bool> get isInactive => _isInactiveController.stream;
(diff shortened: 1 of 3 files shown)


malik1004x:
I was unable to reproduce the issue this build is supposed to fix.

Please specify on what app version and platform this occured on.

sethforprivacy:
> I was unable to reproduce the issue this build is supposed to fix.
> 
> Please specify on what app version and platform this occured on.

This came from an external report, but I was able to reproduce it in an iOS emulator (but haven't tried on a live app/iPhone yet FWIW).

PR #3520, opened 2026-08-14
Feat/bnb srd coin selection

### LINKED ISSUE #3407: Coin selection follows a predictable order (privacy fingerprint)

konstantinullrich:
Closes #3407, #3408

# Description

coin selection walked unspentCoins in a fixed order (address generation order, then coin age), so the input set was deterministic, a fingerprint an analyst can exploit, and always produced residual change.

selection now works like Core/bdk (BranchAndBound<SingleRandomDraw>):

branch-and-bound: searches over effective values for an input set whose excess over amount + fee stays below dust → the change output is dropped and the residue goes to fee (changeless tx, overpay bounded by dust).
single random draw fallback: no match → the pool is shuffled with Random.secure() before the existing accumulate-until-covered loop.
the RBF fee-bump top-up and payjoin receiver candidates are also shuffled.

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_bitcoin/lib/electrum_wallet.dart
+++ b/cw_bitcoin/lib/electrum_wallet.dart
@@ -4,15 +4,6 @@ import 'dart:isolate';
 import 'dart:math' show Random;
 
 import 'package:bitcoin_base/bitcoin_base.dart';
-import 'package:cw_bitcoin/lightning/lightning_wallet.dart';
-import 'package:cw_core/hardware/hardware_wallet_service.dart';
-import 'package:cw_core/root_dir.dart';
-import 'package:cw_core/utils/proxy_wrapper.dart';
-import 'package:cw_core/utils/print_verbose.dart';
-import 'package:cw_bitcoin/bitcoin_wallet.dart';
-import 'package:cw_bitcoin/coin_selection.dart';
-import 'package:cw_bitcoin/litecoin_wallet.dart';
-import 'package:shared_preferences/shared_preferences.dart';
 import 'package:blockchain_utils/blockchain_utils.dart';
 import 'package:collection/collection.dart';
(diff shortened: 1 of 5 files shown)


cw_bitcoin/lib/electrum_wallet.dart
```diff
@@ -873,36 +897,78 @@ abstract class ElectrumWalletBase
     required bool paysToSilentPayment,
     int credentialsAmount = 0,
     int? inputsCount,
+    int feeRate = 0,
+    int? outputsVBytes,
+    List<BitcoinUnspent>? shuffledInputs,
```
OmarHatem28, line 902: it's always null, it's not passed at any call of this function 🙃

konstantinullrich: This is not the case, it is passed on L1231.

PR #3524, opened 2026-08-15
fix: delete leftover ~_BACKUP_TMP after exporting a backup

### LINKED ISSUE #3264: ~_BACKUP_TMP directory never cleaned up after backup export (Windows)
**Describe the bug**
After completing a backup export, the temporary directory ~_BACKUP_TMP is left behind in the app data folder (%APPDATA%\com.cakewallet.cake_wallet\Cake Wallet\). Users may think their backup failed or wonder what the orphaned folder is.

**To Reproduce**
1. Go to Settings → Backup → Create backup
2. Complete the backup export process
3. Navigate to %APPDATA%\com.cakewallet.cake_wallet\Cake Wallet\
4. See leftover ~_BACKUP_TMP directory

**Expected behavior**
The ~_BACKUP_TMP directory should be automatically deleted after a successful backup export.

**Screenshots**
N/A

**Platform:**
- OS: Windows
- Device: PC
- Cake Wallet Version: 5.9.0

**Additional context**
Root cause found in lib/core/backup_service_v3.dart — the cleanup line at the end of the export function is commented out:
// tmpDir.deleteSync(recursive: true);

**Suggested fix**
Uncomment that line, or move it into a finally block to ensure cleanup on both success and failure.

**Credit**
I ran into the problem, but to be fair, Claude tracked it down and wrote this entire description.

o-mid:
## Summary
- Backup export left `~_BACKUP_TMP` in the app directory because the finished zip was returned from inside that folder, so cleanup had to stay commented out.
- Copy the export zip out of the temp dir, delete `~_BACKUP_TMP` in a `finally`, ignore it in later backups, and remove leftovers on startup.

Fixes cake-tech/cake_wallet#3264

## Test plan
- [ ] Windows: export a backup, save it, and confirm `~_BACKUP_TMP` is gone from the Cake app directory
- [ ] If an old `~_BACKUP_TMP` is already present, launch the app and confirm it is removed
- [ ] Confirm the saved backup zip still restores


lib/main.dart
```diff
@@ -215,6 +215,12 @@ Future<void> initializeAppAtRoot({bool reInitializing = false}) async {
 Future<void> initializeAppConfigs({bool loadWallet = true}) async {
   setRootDirFromEnv();
   final appDir = await getAppDir();
+  final leftoverBackupTmp = Directory('${appDir.path}/~_BACKUP_TMP');
```
MrCyjaneK: I'd prefer join from path/path.dart here so it works on windows too (in other places too)

o-mid: Switched those to `p.join` from `package:path/path.dart` — the leftover `~_BACKUP_TMP` on startup and the other temp/export paths in `exportBackupFileV3`.

o-mid:
@MrCyjaneK done — switched those to `p.join` from `package:path/path.dart` so the paths work on Windows. That covers the leftover `~_BACKUP_TMP` cleanup on startup and the other temp/export paths in `exportBackupFileV3`.

Issue #1393, opened 2024-04-16
Request: ARM build for linux, which will possibly enable Cake Wallet for Linux phones


kristianbrox:
It would be great with an ARM build of the Linux version of Cake wallet, as the current version will not work with ARM computers. 
The current version of the Linux version seem to work great, and is even adaptable for narrow screens. This could potentially work great for mobile Linux such as the pinephone, librem 5, and postmarketOS android phones, if there was an ARM version.

The mobile Linux space is missing a good crypto wallet application, and cake wallet would be great to have working.

oopen:
Need ARM64 / AARCH64 Linux build,

to support SBC, like Raspberry PI and all other single board computer.

garethgtt:
Pi Version would be super awesome

PR #3526, opened 2026-08-15
fix: shrink the desktop balance card so history stays on screen

### LINKED ISSUE #3255: Desktop: shrink the size of the card
**Is your feature request related to a problem? Please describe.**
I don't know what the point of this **GIANT** credit-card-like graphic is, maybe it makes sense on mobile, but it takes up a massive amount of room on desktop for no purpose, and pushes transaction history off the bottom of the screen.

**Describe the solution you'd like**
Either make it sanely sized by default on desktop, or add some sort of option to make it small.

**Describe alternatives you've considered**
I don't know what the point of this feature is in the first place, and it's not explained anywhere.

**Additional context**

Image

o-mid:
## Summary
- Portrait desktop windows were treated as mobile UI, so the balance card could grow to 768px and push transaction history off the page.
- Cap the desktop card at 320px (40% of width, whichever is smaller).

Fixes cake-tech/cake_wallet#3255

## Test plan
- [ ] Linux / macOS / Windows: dashboard card is credit-card sized and history is visible without scrolling past a giant card
- [ ] Phone / narrow window: card still uses the existing mobile width

Made with [Cursor](https://cursor.com)

### DIFF
--- a/lib/new-ui/widgets/coins_page/cards/cards_view.dart
+++ b/lib/new-ui/widgets/coins_page/cards/cards_view.dart
@@ -8,6 +8,7 @@ import 'package:cake_wallet/generated/i18n.dart';
 import 'package:cake_wallet/new-ui/modal_navigator.dart';
 import 'package:cake_wallet/new-ui/pages/send_page.dart';
 import 'package:cake_wallet/routes.dart';
+import 'package:cake_wallet/utils/device_info.dart';
 import 'package:cake_wallet/utils/feature_flag.dart';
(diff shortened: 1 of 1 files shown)


RasheedAZ:
Is it possible to cap at a higher percentage of the screen width, say 70%, because 40% will be too small when it runs on Linux phones (open request #1393) which run 'desktop' apps.

Or maybe it can check if it's a small screen device...

`hostnamectl status | grep Chassis`  gives: `handset` on Linux phones like Librem 5, FLX1s, etc.

PR #3538, opened 2026-08-19
Delete CakePay Auth-token if the API returns 401

lapatric:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/cake_pay/src/services/cake_pay_service.dart
```diff
@@ -95,29 +96,51 @@ class CakePayService {
...
+          orderId: orderId,
+        ),
+      );
+
+  /// Wrapper for requests with user token
+  Future<T> _withUserToken<T>(Future<T> Function(String token) request) async {
+    final token = await secureStorage.read(key: cakePayUserTokenKey) ?? '';
+
+    if (token.isEmpty) {
+      throw const CakePayUnauthorizedException();
```
OmarHatem28: shouldn't we call logout here as well? or put this condition in the try below so that the catch does that

and one other note, the exception for the `findOrderById` flow will be swallowed by the generic catch and do nothing, and the user won't know why it's not working, we should at least show a popup with the exception error message, so he knows he should go and login again in cake pay

lib/cake_pay/src/cards/cake_pay_buy_card_page.dart
```diff
@@ -504,9 +505,11 @@ class CakePayBuyCardPage extends BasePage {
     } else {
       try {
         await cakePayBuyCardViewModel.createOrder();
-      } catch (_) {
+      } on CakePayUnauthorizedException {
         cakePayBuyCardViewModel.isSimulatingFlow = false;
-        await cakePayBuyCardViewModel.logout();
```
OmarHatem28, line 509: you should still catch other exceptions

lib/cake_pay/src/services/cake_pay_service.dart
```diff
@@ -95,29 +96,51 @@ class CakePayService {
...
+    }
```
konstantinullrich: Why not use `if (token == null)` instead of `?? ""` and `token.isEmpty`

lib/cake_pay/src/cards/cake_pay_cards_page.dart
```diff
@@ -477,11 +477,18 @@ class _ShopTabState extends State<_ShopTab> {
...
+                      logoUrl: vendor.card?.cardImageUrl,
+                      title: vendor.name,
+                      subTitle: vendor.card?.description ?? '',
+                      onTap: () async {
+                        await Navigator.pushNamed(
+                          context,
+                          Routes.cakePayBuyCardPage,
+                          arguments: [vendor],
+                        );
+                        await viewModel.checkAuth();
```
konstantinullrich, line 489: Why do you check the Auth after pushing to the new page?

lapatric: Awaiting `Navigator.pushNamed` completes when the user returns to this page, and I had observed that the profile state was stale at this point after a 401-related logout. So this keeps it synced.

lib/cake_pay/src/services/cake_pay_service.dart
lapatric: The main thinking here was to also catch the possibility of the user token itself coming back as empty. `isLogged()` above does the same.

konstantinullrich (approved):

OmarHatem28 (changes requested):

OmarHatem28 (approved):

PR #3547, opened 2026-08-21
CW-1615: Harden walletconnect auth

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/src/screens/wallet_connect/services/key_service/wallet_connect_key_service.dart
```diff
@@ -43,30 +43,36 @@ class KeyServiceImpl implements WalletConnectKeyService {
...
-    final keys = [
-      ChainKeyModel(
-        chains: [
-          'eip155:1',
-          'eip155:5',
-          'eip155:56',
-          'eip155:137',
-          'eip155:8453',
-          'eip155:42161',
-          'eip155:80001',
```
OmarHatem28, line 55: why did we remove 5 and 80001? what were those?

lib/src/screens/wallet_connect/utils/method_utils.dart
```diff
@@ -11,9 +11,27 @@ import 'package:flutter/material.dart';
 import 'package:reown_walletkit/reown_walletkit.dart';
 
 class MethodsUtils {
-  static final walletKit = getIt.get<WalletKitService>().walletKit;
+  static ReownWalletKit get walletKit => getIt.get<WalletKitService>().walletKit;
```
OmarHatem28, line 14: is initializing a new object everytime this getter is called, the intended behavior?
I don't think so, because it's used to get the pending requests and sessions
and even if it fetches those anyway, still why initialize a new object everytime

lib/src/screens/wallet_connect/services/key_service/wallet_connect_key_service.dart
Blazebrain: They're for testnets, and we don't use them

That's for Goerli and Mumbai

lib/src/screens/wallet_connect/utils/method_utils.dart
Blazebrain: Yeah, the behavior is intended, but it doesn't create a new instance since we registered the `WalletKitService` as a `lazySingleton,` so it pretty much returns the same instance

The major reason for changing it is so we can get the current `walletKit` object attached to the service especially when we switch wallets. Because we were using static before, it bsaically just held the first instance and never updated it when the user switched wallets, but accesing it via the getter would allow us to get the updated `_walletKit` even if the user switches the wallet

PR #3548, opened 2026-08-21
Enhance Money class with precision handling and zero trimming

konstantinullrich:
# Description

This pull request enhances the `Money` class by introducing the following improvements:
- Added precision handling for better accuracy in monetary calculations.
- Implemented scale alignment to ensure consistency in arithmetic operations.
- Introduced functionality to trim trailing zeros for a cleaner representation of monetary values.

# Pull Request - Checklist  

- [x] Format code


cw_core/lib/amount/utils.dart
```diff
@@ -0,0 +1,32 @@
...
+    return source;
+  }
+
+  if (sourceDecimals > targetDecimals) {
+    return parseFixed(
+      formatFixed(source, sourceDecimals).withMaxDecimals(targetDecimals),
+      targetDecimals,
+    );
+  } else {
+    return parseFixed(formatFixed(source, sourceDecimals), targetDecimals);
```
malik1004x: This does appear slightly overcomplicated compared to the "obvious" solution of multiplying/dividing it by a power of 10. Is there an edge case here that I can't think of?

cw_core/lib/amount/money.dart
```diff
@@ -156,54 +194,64 @@ class Money implements Comparable<Money> {
...
+    return (
+      a: _scaleUp(amount, decimals - this.decimals),
+      b: _scaleUp(other.amount, decimals - other.decimals),
+      decimals: decimals,
+    );
+  }
+
+  static int _getActualDecimals(String value, Currency currency) {
+    final comps = value.split(".");
+    if (comps.length > 2) {
```
malik1004x, line 249: Would this mean there's more than one dot in the string we're parsing? If yes, have you considered throwing a parse failure exception instead of returning this fallback? Is there a case where a properly parsable string (one that can be properly turned into Money) and uses this fallback?

konstantinullrich: If you don't throw localized strings at it typically no

cw_core/lib/amount/utils.dart
konstantinullrich: That is true, but I just moved it here and it's only used in a specific edge case

cw_core/lib/amount/money.dart
malik1004x: In that case I would throw an exception here. Providing this fallback could be dangerous.

cw_core/lib/amount/utils.dart
malik1004x: If it works and is tested, I'm fine with it.

cw_core/lib/amount/money.dart
konstantinullrich: That exception is thrown during formatFixed anyway

malik1004x: Doesn't matter. If this function does not have correct behavior in this case, it should throw an exception as well – what if we change formatFixed later?

malik1004x (approved):
lgtm

cw_core/lib/amount/money.dart
```diff
@@ -7,39 +9,59 @@ import "package:cw_core/parse_fixed.dart";
...
+  static Money? tryParse(
+    String source,
+    Currency currency, {
+    bool isBaseUnit = false,
+    bool strictParsing = true,
+  }) {
+    if (!isBaseUnit) {
+      source = trimTrailingFractionZeros(source);
+    }
+    final decimals = strictParsing ? currency.decimals : _getActualDecimals(source, currency);
```
OmarHatem28: _getActualDecimals can throw, so shouldn't it be in a try/catch? so that this tryParse indeed return null when the source is invalid instead of throwing an uncaught exception

PR #3576, opened 2026-08-28
CW-1468 Home Screen Top Bar Layout

lapatric:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/di.dart
```diff
@@ -1025,11 +1032,39 @@ Future<void> setup({
...
+
+      return MoneroAccountEditOrCreateViewModel(
+        accountList,
+        null,
+        wallet: wallet,
+        accountListItem: account,
+      );
+    }
+
+    if (wallet.type == WalletType.wownero) {
```
malik1004x: Wownero support has been removed in a previous update, all conditions checking for the current wallet type being Wownero are effectively dead code.

lib/di.dart
```diff
@@ -1025,11 +1032,39 @@ Future<void> setup({
...
-          wownero?.getAccountList(getIt.get<AppStore>().wallet!),
-          wallet: getIt.get<AppStore>().wallet!,
-          accountListItem: account));
+      (AccountListItem? account, _) {
+    final wallet = getIt.get<AppStore>().wallet!;
+
+    if (wallet.type == WalletType.monero) {
+      final accountList = monero?.getAccountList(wallet);
+      if (accountList == null) {
+        throw StateError("Monero account support is unavailable");
```
malik1004x: [minor] This can never trigger – if `appStore.wallet!.type` is Monero, the `monero` object is definitely non-null (since otherwise the app would crash). It's safe to remove this condition and just add an explicit null check on `monero`.

lib/view_model/monero_account_list/monero_account_edit_or_create_view_model.dart
```diff
@@ -33,7 +33,7 @@ abstract class MoneroAccountEditOrCreateViewModelBase with Store {
   @observable
   String label;
 
-  final MoneroAccountList _moneroAccountList;
+  final MoneroAccountList? _moneroAccountList;
```
malik1004x: Not sure when this can be null – this ViewModel should only be accessible from the account creation modal, yes? If so, this will never be non-null as that modal's not openable outside of Monero wallets. You seem to be null-checking it in a few places anyway.

lib/view_model/monero_account_list/monero_account_edit_or_create_view_model.dart
```diff
@@ -48,25 +48,26 @@ abstract class MoneroAccountEditOrCreateViewModelBase with Store {
     return ret.isNotEmpty ? ret : CardDesign.allGradients;
   }
 
-  Future<void> _saveRandomCardDesign() async {
+  Future<void> _saveRandomCardDesign(int accountIndex) async {
```
malik1004x: You seem to have refactored this for Wownero's sake – remember Wownero has been removed, so it's safe to revert.

lib/new-ui/widgets/coins_page/wallet_info.dart
```diff
@@ -1,102 +1,68 @@
...
-      onCustomizeButtonTap();
-      HapticFeedback.mediumImpact();
-    }
-  }
 
   @override
   Widget build(BuildContext context) {
-    // The row, the hardware-wallet glyph and the inner accounts button are one
-    // control for a screen reader: a single labeled node opening the customizer.
+    final iconPath = hardwareWalletIcon;
```
malik1004x: [minor] Why this? Why not just use `hardwareWalletIcon` directly?

lib/new-ui/pages/home_page.dart
```diff
@@ -130,70 +128,77 @@ class _NewHomePageState extends State<NewHomePage> {
...
-                      Observer(
-                        builder: (_) => WalletInfoBar(
-                            lightningMode: _lightningMode,
-                            hardwareWalletType: widget.dashboardViewModel.wallet.hardwareWalletType,
-                            name: widget.dashboardViewModel.wallet.name,
-                            hasCustomize: accountListViewModel != null,
-                            onCustomizeButtonTap: openAccountCustomizer),
-                      ),
                       Column(
+                        spacing: isEVMWallet ? 24 : 20,
```
malik1004x: Why do EVM wallets have different home page spacing?

lib/new-ui/widgets/coins_page/assets_history/assets_top_bar.dart
```diff
@@ -26,7 +26,12 @@ class AssetsTopBar extends StatelessWidget {
 
     return SliverToBoxAdapter(
       child: Padding(
-        padding: const EdgeInsets.only(top: 32.0, bottom: 0.0, left: 12.0, right: 18.0),
+        padding: EdgeInsets.only(
+          top: dashboardViewModel.isEVMWallet ? 24 : 20,
```
malik1004x: Why would EVM have different spacing?

lib/new-ui/widgets/coins_page/top_bar_widget/lightning_switcher.dart
```diff
@@ -30,64 +30,59 @@ class LightningSwitcher extends StatelessWidget {
...
-                            BlendMode.srcIn,
-                          ),
+                Row(
+                  children: [
+                    AnimatedSwitcher(
+                      duration: const Duration(milliseconds: 150),
+                      transitionBuilder: (child, animation) =>
+                          FadeTransition(opacity: animation, child: child),
+                      child: CakeImageWidget(
+                        imageUrl: 'assets/new-ui/switcher-bitcoin.svg',
```
malik1004x: [minor] Please use double quotes here.

lib/new-ui/widgets/coins_page/top_bar_widget/top_bar.dart
```diff
@@ -65,19 +89,24 @@ class _TopBarState extends State<TopBar> {
...
-            (widget.dashboardViewModel.hasLightning)
+        padding: EdgeInsets.only(left: 18, right: 18, top: 10 + _additionalTopPadding(context)),
+        child: Observer(
+          builder: (_) {
+            final syncBar = SyncBar(
+              dashboardViewModel: widget.dashboardViewModel,
+              isSyncHeavy: widget.dashboardViewModel.isSyncHeavy,
+              showSyncedMessage: showSyncedMessage,
+            );
+            final replacesWalletName = syncBar.replacesWalletName;
```
malik1004x: `replacesWalletName` should not be logic passed into the `SyncBar` – this requires putting the widget in this variable instead of in the regular widget tree, but everything that logic needs is explicitly passed to the syncbar here anyway. You can move `replacesWalletName` to a getter in this widget, then you'll no longer need these extra variables here, and if needed, you can pass it as a `bool` to the sync bar.

lib/new-ui/widgets/coins_page/top_bar_widget/top_bar.dart
```diff
@@ -101,15 +126,74 @@ class _TopBarState extends State<TopBar> {
...
+            }
+
+            return Row(
+              spacing: 12,
+              children: [
+                leading,
+                Expanded(
+                  child: SizedBox(
+                    height: 36,
+                    child: replacesWalletName
```
malik1004x, line 139: This means the wallet name awkwardly appears with no fade-in (I tested this on a build to confirm). Consider using an `AnimatedSwitcher` here to get a nicer feel.

lib/new-ui/widgets/coins_page/top_bar_widget/top_bar.dart
```diff
@@ -101,15 +126,74 @@ class _TopBarState extends State<TopBar> {
...
+                              if (syncBar.hasCompactContent) ...[
+                                const SizedBox(width: 6),
+                                syncBar,
+                              ],
+                            ],
+                          ),
+                  ),
+                ),
+                if (hasLightning && !replacesWalletName)
+                  SizedBox(
```
malik1004x: I assume this is to get the settings button the same size as the Lightning switcher. If so, maybe do it the other way around – adjust the Lightning switcher's size to fir the settings button. Then, you don't need a special condition here.

lib/new-ui/widgets/coins_page/top_bar_widget/top_bar.dart
```diff
@@ -101,15 +126,74 @@ class _TopBarState extends State<TopBar> {
               },
               svgPath: "assets/new-ui/top-settings.svg",
               semanticLabel: S.of(context).settings_title,
-            ),
-          ],
+            );
+
+            if (hasLightning && widget.dashboardViewModel.isSyncHeavy && replacesWalletName) {
```
malik1004x: Why is there a special condition for what I assume is Silent Payments with Lightning? The pattern is still the same – `leading`, then the sync bar, then the settings button, but here `syncBar` has a hardcoded size – why? If possible, please put any SP-specific UI into conditions rather than an entire separate tree. Please disregard if there was some special UI for this case requested that I'm unaware of.

malik1004x (changes requested):

lib/new-ui/pages/home_page.dart
lapatric: Not sure, this was in the Figma file, I'll check with the designer

lapatric: Asked designers, removing special padding for EVM for now

malik1004x (approved):

OmarHatem28 (approved):

PR #3582, opened 2026-08-31
Use baseunit provider

konstantinullrich:
# Description

This PR adds a new Utility class to replace AmountParsingProxy

# Pull Request - Checklist  

- [x] Format code


malik1004x (approved):

OmarHatem28 (approved):
should this layer consider the hide balance option as well or that would be handled by another layer before?
if another layer, then merge it

konstantinullrich:
Hiding the balance is more a UI feature so `MoneyText` handles that.

PR #3584, opened 2026-09-01
CW-1529: Improved Seed UI

Blazebrain:

# Description

Improves the flow for seed verification. Adds reminders and steps to ensure the user actually verifies their recovery phrase either at creation/restore or at the point where the first transaction enters the wallet

# Pull Request - Checklist  

- [x] Format code


cw_core/lib/db/sqlite.dart
```diff
@@ -152,10 +152,17 @@ CREATE TABLE IF NOT EXISTS BalanceCardStyleSettings (
...
-
     if (oldVersion <= 10) {
       await _createImportedNFTTable(db);
     }
+    if (oldVersion <= 11) {
+      await _addColumnIfNotExists(
+        db,
+        table: 'WalletInfo',
+        column: 'showSeedBackupReminder',
+        definition: 'BOOLEAN DEFAULT FALSE',
```
OmarHatem28: why default false? shouldn't this be true at first

lib/core/auth_service.dart
```diff
@@ -36,6 +36,7 @@ class AuthService with Store {
 
   static const List<String> _alwaysAuthenticateRoutes = [
     Routes.showKeys,
+    Routes.seed,
```
OmarHatem28: isn't this the page on the wallet creation flow? why require the pin there?

lib/view_model/wallet_new_vm.dart
```diff
@@ -164,6 +164,7 @@ abstract class WalletNewVMBase extends WalletCreationVM with Store {
   @override
   Future<WalletBase> process(WalletCredentials credentials) async {
     walletCreationService.changeWalletType(type: type);
+    credentials.walletInfo!.showSeedBackupReminder = newWalletArguments!.mnemonic == null;
```
OmarHatem28, line 167: so it's only for newly created wallets, right?
existing wallets won't ever show that reminder, correct?
nothing to change, just making sure I get it correctly

Blazebrain: Yeah, it won't show for existing wallets

lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -1122,6 +1140,17 @@ abstract class DashboardViewModelBase with Store {
     sharedPreferences.setBool(PreferencesKey.showDecredInfoCard, false);
   }
 
+  @action
+  void loadSeedBackupReminder() {
```
malik1004x, line 1144: Have you checked if this reloads properly if the active wallet changes?

res/pictures/key.svg
malik1004x, line 1: [minor] Could you please change the `fill` to `white`? So it doesn't get this color if called without `colorFilter`. We shouldn't display this without a filter anyway, but just in case (this is what I do for the other images).

lib/new-ui/widgets/seed/seed_checkbox_row.dart
```diff
@@ -0,0 +1,77 @@
...
+                    child: Text(
+                      text,
+                      style: Theme.of(context).textTheme.bodySmall?.copyWith(
+                            color: Theme.of(context).colorScheme.primary,
+                            letterSpacing: -0.06,
+                          ),
+                    ),
+                  ),
+                  ExcludeSemantics(
+                    child: Container(
```
malik1004x, line 53: This is generally okay, but we also have `NewSimpleCheckbox` if you'd like to avoid duplicated UI code.

malik1004x (commented):
Not requesting changes since there is no major issue present, but please take a look at the comments.

OmarHatem28 (approved):
@Blazebrain please fix the conflicts and merge it, check Robert's comments as well if anything should be changed while resolving the conflicts

lib/view_model/dashboard/dashboard_view_model.dart
Blazebrain: Yeah it does, the wallet change reaction calls `loadSeedBackupReminder()` right after `_onWalletChange`, so it re-reads the flag from the new wallet's `walletInfo` on every switch

res/pictures/key.svg
Blazebrain: Alrighty

lib/new-ui/widgets/seed/seed_checkbox_row.dart
Blazebrain: Alright, would switch it

PR #3597, opened 2026-09-03
fix: brazillian localle bug that makes app unresponsive

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code

### DIFF
--- a/cw_core/lib/amount/money_local.dart
+++ b/cw_core/lib/amount/money_local.dart
@@ -1,5 +1,5 @@
 import "package:cw_core/amount/money.dart";
-import "package:intl/intl.dart";
+import "package:cw_core/crypto_amount_format.dart";
 
 extension WithLocalSeparator on Money {
   String toLocalStringWithSymbol({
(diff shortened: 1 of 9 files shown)


cw_core/lib/amount/money_local.dart
```diff
@@ -37,10 +37,7 @@ extension WithLocalSeparator on Money {
 
   String _withLocalSeparator(String amount, {String? locale}) {
     final isNegative = amount.startsWith("-");
-    final formater = NumberFormat("#,###", locale);
```
konstantinullrich, line 40: I would rather have this fixed than to rely on String.withLocalSeperator. Since the string extension was just a quick patch for when money was not a thing and numberFormat might have an issue with large numbers like the ones you can get with certain eth tokens

lib/src/widgets/new_list_row/list_item_selector_widget.dart
konstantinullrich, line 1: Seems like an unrelated change

konstantinullrich (changes requested):

lib/src/widgets/new_list_row/list_item_selector_widget.dart
OmarHatem28: can be kept, it's ok

cw_core/lib/amount/money_local.dart
Blazebrain: Ohh, alright. I didn't know the extension has been retired, I'll push an update for it

Maybe we can plan to remove retired non-money extensions and other components

konstantinullrich: I was mistaken, I forgot I fixed the bigint issue

konstantinullrich (approved):

PR #3600, opened 2026-09-04
CW-1513/CW-1610: Add account archival and UI improvements

lapatric:

# Description

# Pull Request - Checklist  

- [ ] Format code


lib/new-ui/pages/account_customizer.dart
```diff
@@ -417,89 +584,173 @@ class _AccountCreationModalState extends State<AccountCreationModal> {
...
+                        const SizedBox(height: 24),
+                        Text(
+                          S.of(context).account_creation_description,
+                          textAlign: TextAlign.center,
+                          style: TextStyle(
+                            color: Theme.of(context).colorScheme.onSurfaceVariant,
+                            fontSize: 14,
+                          ),
+                        ),
+                        const SizedBox(height: 24),
```
malik1004x, line 685: [minor] Why not `spacing`?

lib/new-ui/widgets/coins_page/accounts_promo.dart
```diff
@@ -0,0 +1,128 @@
...
+
+  @override
+  void initState() {
+    super.initState();
+    _dismissed = widget.preferences.getBool(PreferencesKey.accountsHomePromoDismissed) ?? false;
+  }
+
+  Future<void> _dismiss() async {
+    setState(() => _dismissed = true);
+    await widget.preferences.setBool(PreferencesKey.accountsHomePromoDismissed, true);
```
malik1004x, line 37: Please use `SettingsStore` for this – it wraps the keys in observable variables, meaning you can just wrap the widget in `Observer` and read the dismissed state directly in the build method. This way, this widget becomes stateless and has no chance of stale state.

lib/new-ui/widgets/coins_page/accounts_promo.dart
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    if (_dismissed) {
+      return const SizedBox.shrink();
+    }
+
+    final theme = Theme.of(context);
+    final strings = S.of(context);
```
malik1004x, line 47: [minor] Why put them in variables? You can use `S.of(context)` and `Theme.of(context)` directly.

lib/new-ui/pages/settings_page.dart
```diff
@@ -3,24 +3,24 @@ import 'package:cake_wallet/entities/new_ui_entities/list_item/list_item.dart';
...
+import 'package:cake_wallet/new-ui/pages/coin_control_page.dart';
 import 'package:cake_wallet/new-ui/widgets/receive_page/receive_top_bar.dart';
 import 'package:cake_wallet/routes.dart';
 import 'package:cake_wallet/src/widgets/cake_image_widget.dart';
 import 'package:cake_wallet/src/widgets/new_list_row/new_list_section.dart';
 import 'package:cake_wallet/view_model/dashboard/dashboard_view_model.dart';
+import 'package:cw_core/currency_for_wallet_type.dart';
 import 'package:cw_core/wallet_info.dart';
-import "package:cw_core/wallet_type.dart";
+import 'package:cw_core/wallet_type.dart';
```
malik1004x, line 14: [minor] You changed double quotes to single quotes, why? Would be more beneficial to change the rest of the imports to doubles.

lib/new-ui/pages/settings_page.dart
```diff
@@ -59,160 +61,581 @@ class SettingsSectionData {
...
+}
+
+/// Resolves the nested wallet settings for every [WalletType].
+///
+/// This switch deliberately has no wildcard/default branch. Adding a wallet
+/// type therefore requires an explicit decision about its settings here.
+class WalletSettingsResolver {
+  const WalletSettingsResolver();
+
+  List<WalletSettingsItemType> settingsFor(WalletType walletType) => switch (walletType) {
```
malik1004x, line 104: While this does generally work, we generally try to avoid hardcoding wallet-specific conditions in newly shipped code, plus it doesn't allow flexibility this page may need.

What you could do instead is resolve these settings based on what the wallet actually provides, for example: you can check for the Lightning username settings by getting `hasLightningSupport` from `WalletBase`. This way we can, for example, only show some settings if certain features are available or enabled (ex. Lightning features should not be visible on hardware wallets). For some things there will be no simple getter, you can either add a few lines of logic or create a getter in `WalletBase` or `DashboardViewModel`. This would be in-line with what we currently do for regular settings rows and more maintainable in the future – if for example we add Coin Control to Solana it just starts working, no need to go in here and enable it.

Also, I see you've added lists for Haven, Wownero and Banano – these wallet types are no longer supported in the app and it's safe to remove these.

lib/new-ui/pages/settings_page.dart
```diff
@@ -59,160 +61,581 @@ class SettingsSectionData {
...
+            WalletSettingsItemType.accounts,
+            WalletSettingsItemType.nodes,
+          ],
+      };
+
+  String titleFor(WalletType walletType, S strings) =>
+      '${walletTypeToString(walletType)} ${strings.settings_title}'.trim();
+
+  String iconPathFor(WalletType walletType) => switch (walletType) {
+        WalletType.none => 'assets/new-ui/wallet-setting.svg',
```
malik1004x, line 208: This switch expression essentially always returns `'assets/new-ui/network_icons/${walletType.name.toLowerCase()}.svg'` – Wownero, Haven, Banano and "none" can never be loaded, therefore it's safe to replace with a simple getter.

Also, please use double quotes.

lib/new-ui/pages/settings_page.dart
```diff
@@ -59,160 +61,581 @@ class SettingsSectionData {
...
+  lightningUsername,
+  silentPayments,
+  mweb,
+  walletConnect,
+  resyncDevice,
+}
+
+extension on WalletSettingsItemType {
+  bool get isCoreSetting => switch (this) {
+        WalletSettingsItemType.accounts ||
```
malik1004x, line 84: [minor] You could instead use a `static const List<WalletSettingsItemType> _coreSettings` and then use `.contains` on that list – saves a few lines and requires no modification if new setting items are added.

You could also consider making this a class with a `bool isCore` field, would be cleaner in the long term and also make the aforementioned conditional display logic easier to implement.

lib/new-ui/pages/settings_page.dart
```diff
@@ -59,160 +61,581 @@ class SettingsSectionData {
...
+        WalletType.decred ||
+        WalletType.dogecoin ||
+        WalletType.base ||
+        WalletType.arbitrum ||
+        WalletType.zcash ||
+        WalletType.bsc =>
+          'assets/new-ui/network_icons/${walletType.name.toLowerCase()}.svg',
+      };
+
+  List<List<SettingsListItem>> resolveSections(S strings, DashboardViewModel viewModel) {
```
malik1004x, line 232: You don't need to pass `S strings` – if you need strings without `BuildContext`, use `S.current`.

lib/new-ui/pages/settings_page.dart
```diff
@@ -59,160 +61,581 @@ class SettingsSectionData {
...
-  ]);
-
-  static List<SettingsSectionData> all = [walletSettings, appSettings, otherSettings];
+  SettingsListItem _buildItem(
+    WalletSettingsItemType type,
+    S strings,
+    WalletType walletType,
+  ) =>
+      switch (type) {
+        WalletSettingsItemType.accounts => SettingsListItem(
```
malik1004x, line 265: Again, this showcases the advantages of making `WalletSettingsItemType` a class instead of an enum – instead of doing this switch-case you could generate `SettingsListItem` from fields of the class, or even make `WalletSettingsItemType` a subclass of `SettingsListItem`.

lib/new-ui/pages/settings_page.dart
```diff
@@ -59,160 +61,581 @@ class SettingsSectionData {
...
-                    }
-                  }
-                })
-            : null)
-        .whereType<ListItem>()
-        .toList();
+    final strings = S.of(context);
+    final wallet = dashboardViewModel.wallet;
+    final walletSettingsTitle = _walletSettingsResolver.titleFor(wallet.type, strings);
+    final appSettings = _sectionsResolver.appSettings(strings);
```
malik1004x, line 424: [minor] Why are these variables? You can use them directly in the tree, and then you're able to use an expression body instead of a block body.

lib/new-ui/pages/settings_page.dart
```diff
@@ -59,160 +61,581 @@ class SettingsSectionData {
...
                       ),
-                    NewListSections(sections: {section.title: buildItems(section)}),
-                  ];
-                }).toList(),
-              ),
+                    ],
+                  },
+                ),
+                const SizedBox(height: 24),
+                NewListSections(
```
malik1004x, line 467: You seem to have hardcoded the sections here. Could you instead make a function that returns a list of sections as well as their header? This way you don't need to hardcode values in the widget tree.

lib/new-ui/pages/account_education_page.dart
```diff
@@ -0,0 +1,355 @@
...
+  static bool shouldShow(SharedPreferences preferences) =>
+      !(preferences.getBool(PreferencesKey.accountsEducationSeen) ?? false);
+
+  static Future<void> markSeen(SharedPreferences preferences) async {
+    await preferences.setBool(PreferencesKey.accountsEducationSeen, true);
+  }
+
+  static Future<void> show(BuildContext context, SharedPreferences preferences) async {
+    await showCupertinoModalBottomSheet<void>(
+      context: context,
```
malik1004x, line 25:
```suggestion
useRootNavigator: true,
      context: context,
```

You are opening a modal sheet inside a modal sheet, this results in some clunky animations.

lib/new-ui/pages/account_education_page.dart
```diff
@@ -0,0 +1,355 @@
...
+class AccountEducationPage extends StatefulWidget {
+  const AccountEducationPage({required this.preferences, super.key});
+
+  final SharedPreferences preferences;
+
+  static bool shouldShow(SharedPreferences preferences) =>
+      !(preferences.getBool(PreferencesKey.accountsEducationSeen) ?? false);
+
+  static Future<void> markSeen(SharedPreferences preferences) async {
+    await preferences.setBool(PreferencesKey.accountsEducationSeen, true);
```
malik1004x, line 20: Again, please use `SettingsStore` for this. This widget can't become stateless since it also holds a `PageController`, but I think it's worth it for clean code. Or, you could use an SQLite table if taking the genericized approach described below.

lib/new-ui/pages/account_education_page.dart
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    await _pageController.nextPage(
+      duration: const Duration(milliseconds: 250),
+      curve: Curves.easeOutCubic,
+    );
+  }
+
+  @override
+  Widget build(BuildContext context) {
```
malik1004x, line 76: This approach is fine on its own, however, the UI team is currently working on more education pages like this to be added to flows across the app.

Therefore, I'd like to propose a genericized approach – declare `abstract class EducationPage` containing `List<EducationSlide> slides`, `String title` and `String key`. If we want to make an education page for a specific flow, we extend that and override the getters with whatever information we need – therefore, all the logic for the pages stays the same and we don't have to duplicate it. For dismissing the page, we could store `key` along with a `dismissed` boolean in SQLite instead of relying on Shared Preferences. This way, we won't need to copy-paste logic in the future.

lib/router.dart
malik1004x, line 664: Please save the `getIt.get` result in a variable (`final page = getIt.get<...`) and then pass that variable to `handleRouteWithPlatformAwareness`. This pattern prevents unnecessary rebuilds of the page if inside a modal sheet.

lib/new-ui/widgets/coins_page/wallet_info.dart
```diff
@@ -15,6 +15,7 @@ class WalletInfoBar extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final iconPath = hardwareWalletIcon;
```
malik1004x, line 18: You can use `hardwareWalletIcon` directly.

lib/new-ui/widgets/coins_page/assets_history/assets_top_bar.dart
```diff
@@ -26,8 +26,8 @@ class AssetsTopBar extends StatelessWidget {
 
     return SliverToBoxAdapter(
       child: Padding(
-        padding: const EdgeInsets.only(
-          top: 20,
+        padding: EdgeInsets.only(
+          top: dashboardViewModel.isEVMWallet ? 24 : 20,
```
malik1004x, line 30: Why would EVM wallets have different padding?

lib/new-ui/pages/account_customizer.dart
```diff
@@ -1,42 +1,58 @@
...
+import "package:cw_core/crypto_amount_format.dart";
+import "package:cw_core/generate_name.dart";
+import "package:cw_core/sync_status.dart";
+import "package:cw_core/wallet_type.dart";
+import "package:flutter/cupertino.dart";
+import "package:flutter/material.dart";
+import "package:flutter_bloc/flutter_bloc.dart";
+import "package:modal_bottom_sheet/modal_bottom_sheet.dart";
+
+bool supportsAccountEducationAndArchival(WalletType walletType) => walletType == WalletType.monero;
```
malik1004x, line 36: Why would some wallet types support accounts but *not* support education and archival? Both are just UI-side features that sit in our local database, I don't see why we should block support for any wallet type that reports having account support.

lib/new-ui/pages/account_customizer.dart
```diff
@@ -47,217 +63,306 @@ class AccountCustomizer extends StatefulWidget {
...
+            ),
+            Padding(
+              padding: const EdgeInsets.symmetric(vertical: 24, horizontal: 18),
+              child: Text(
+                S.of(context).account_customizer_desc,
+                textAlign: TextAlign.center,
+                style: TextStyle(color: Theme.of(context).colorScheme.onSurfaceVariant),
+              ),
             ),
+            Expanded(child: _buildBody()),
```
malik1004x, line 233: Don't make functions that return `Widget` – if necessary, split this into its own widget class. Functions mean the tree is unnecessarily rebuilt on unrelated changes since the enigne can't compute what's actually in need of rebuild.

lib/new-ui/pages/account_customizer.dart
```diff
@@ -269,45 +374,122 @@ class _AccountCustomizerState extends State<AccountCustomizer> {
...
+          value: bloc,
           child: Material(
             child: CardCustomizer(
               cryptoTitle: widget.dashboardViewModel.wallet.currency.fullName ??
                   widget.dashboardViewModel.wallet.currency.name,
               cryptoName: widget.dashboardViewModel.wallet.currency.name,
+              accountNumber: account.id + 1,
+              balance: account.balance ?? "0.00",
+              fiatBalance: _fiatBalance(account),
+              onArchive: _supportsAccountArchival
```
malik1004x, line 439: I'm not sure about the approach with the `onArchive` function pointer and then checking return value of the sheet – do you think it would be possible to put the logic for this in one place (preferrably just as the sheet return value and not a function pointer)? Currently it seems like the path is:
- User clicks account archival,
- Flow goes to this file to run this function,
- Flow goes back to card customizer just to run `.pop(true)`,
- Flow goes back here to check for that return value.
This is, in my opinion, an unnecessary split which could be replaced with one function that handles all archival logic.

lib/new-ui/pages/account_customizer.dart
```diff
@@ -322,90 +504,75 @@ class _AccountCustomizerState extends State<AccountCustomizer> {
...
   }
 
-  Future<void> saveCardOrder() async {
-    for (int i = 0; i < _items.length; i++) {
-      final idx = _items.indexWhere((element) => element.accountListItem.id == i);
-      printV("$i: $idx");
+  Future<void> saveCardOrder({int? excludingAccountId}) async {
+    for (int position = 0; position < _items.length; position++) {
+      final item = _items[position];
+      if (item.accountListItem.id == excludingAccountId) {
```
malik1004x, line 533: I assume `excludingAccountId` was added to avoid overwriting `hidden: true` – I think it would be possible to remove it and simplify this function if the archiving logic was consolidated in one place, since then we could ensure `saveCardOrder` is never called before the list is rebuilt.

lib/new-ui/pages/card_customizer.dart
```diff
@@ -271,90 +286,142 @@ class _CardCustomizerState extends State<CardCustomizer> {
...
-                              )
+                              ),
                             ],
                           ),
-                        )),
-                  ],
+                        ),
+                      ),
+                      if (bloc.canHide && widget.onArchive != null)
+                        Padding(
```
malik1004x, line 372: Not 100% sure, but I think this could be replaced with a `NewListSections` and a `ListItemRegularRow`. This way we'd avoid duplicated styling logic.

lib/new-ui/pages/hidden_accounts.dart
```diff
@@ -0,0 +1,734 @@
...
+import "package:cake_wallet/utils/show_pop_up.dart";
+import "package:cake_wallet/view_model/dashboard/dashboard_view_model.dart";
+import "package:cake_wallet/view_model/monero_account_list/account_list_item.dart";
+import "package:cake_wallet/view_model/monero_account_list/monero_account_list_view_model.dart";
+import "package:cw_core/balance_card_layout.dart";
+import "package:cw_core/balance_card_style_settings.dart";
+import "package:cw_core/crypto_amount_format.dart";
+import "package:flutter/material.dart";
+
+Future<bool> confirmAccountArchival(
```
malik1004x, line 21: [minor] This seems to only be called from the function pointer passed to `CardCustomizer` – please put it inside that class instead of making a loose function.

lib/new-ui/pages/hidden_accounts.dart
```diff
@@ -0,0 +1,734 @@
...
+      rightAlertButtonStyle: AlertButtonStyle.secondary(dialogContext),
+      actionLeftButton: () => Navigator.of(dialogContext).pop(false),
+      actionRightButton: () => Navigator.of(dialogContext).pop(true),
+    ),
+  );
+
+  return result ?? false;
+}
+
+Future<bool> confirmAccountUnarchival(
```
malik1004x, line 55: [minor] Can't this be a method of `HiddenAccountsPage` instead of a loose function?

malik1004x (changes requested):

lib/new-ui/pages/account_customizer.dart
lapatric: Exactly, but if the page is closed prematurely and `dispose` is called, it would also call `saveCardOrder` which could then overwrite an archived account if that procedure isn't finished. So removing the `excludingAccountId` would also need to cover that case somehow. For now I'd keep the `excludingAccountId` for simplicity.

lib/new-ui/pages/account_education_page.dart
lapatric: Went with the keyed SharedPreferences for the dismissal tracking instead of the SQLite approach in order to avoid a migration, let me know what you think ;)

cw_core/lib/db/sqlite.dart
```diff
@@ -164,6 +163,16 @@ CREATE TABLE IF NOT EXISTS BalanceCardStyleSettings (
...
     }
+    if (oldVersion <= 12) {
+      await _addColumnIfNotExists(
+        db,
+        table: "BalanceCardStyleSettings",
+        column: "hidden",
+        definition: "BOOLEAN DEFAULT FALSE",
+      );
+      // Version 11 account builds predate the ImportedNFT migration.
+      await _createImportedNFTTable(db);
```
malik1004x, line 174: Builds, yes – but migrations are made with production in mind, not testing builds – so this is unnecessary. You can't update a testing build into a testing build from another branch anyway.

cw_core/lib/wallet_base.dart
```diff
@@ -142,6 +142,16 @@ abstract class WalletBase<BalanceType extends Balance, HistoryType extends Trans
   bool get hasLightningSupport => false;
   bool get hasSilentPaymentsScanning => false;
 
+  bool get hasCoinControl => [
```
malik1004x, line 145: We try to avoid wallet-specific conditions in generic code. You could instead do `bool hasCoinControl => false` and then voerride that with `true` in `ElectrumWallet`, `MoneroWallet` and `DecredWallet`.

lib/new-ui/pages/hidden_accounts.dart
```diff
@@ -0,0 +1,485 @@
...
+  final DashboardViewModel dashboardViewModel;
+  final bool isBusy;
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    if (isBusy) {
+      return const SizedBox.square(
+        dimension: 20,
+        child: CircularProgressIndicator(strokeWidth: 2),
```
malik1004x, line 421: Please use `CupertinoActivityIndicator` instead – it fits the 6.x UI much better. It also has a `radius` parameter so you don't need a `SizedBox`.

lib/new-ui/widgets/account_confirmation_content.dart
```diff
@@ -0,0 +1,222 @@
...
+
+double? _accountAmount(AccountListItem account) {
+  final balance = account.balance;
+  if (balance == null) {
+    return null;
+  }
+  return double.tryParse(balance.trim().replaceAll(",", ""));
+}
+
+String? accountFiatBalance(AccountListItem account, DashboardViewModel dashboardViewModel) {
```
malik1004x, line 202: Why are these loose functions rather than methods? It's not very ergonomic considering you have to pass `dashboardViewModel` here as well.

PR #3615, opened 2026-09-10
feat: restrict history api calls when toggle is off in privacy settings

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


OmarHatem28 (commented):
in updateScanProviderUsageState in both evms and tron, you call cancel on the `_transactionsUpdateTimer` but that also cancels the periodic updateBalance call.
and since you already return early in `fetchTransaction` call, then no need to cancel the updateTimer

Blazebrain:
> in updateScanProviderUsageState in both evms and tron, you call cancel on the `_transactionsUpdateTimer` but that also cancels the periodic updateBalance call. and since you already return early in `fetchTransaction` call, then no need to cancel the updateTimer

Yeah, no reason for it to stay with the early return there. I've removed it on both chains and pushed an update

OmarHatem28 (approved):

PR #3640, opened 2026-09-21
feat: add unit tests CI workflow

Blazebrain:

# Description

# Pull Request - Checklist  

- [ ] Format code


MrCyjaneK:
it's duplicating the build process, can we make it depend on the previous build to make sure that at least cache is populated?

Blazebrain:
> it's duplicating the build process, can we make it depend on the previous build to make sure that at least cache is populated?

Yeah, that makes sense. But instead of making it depend on another build, I'm thinking we can have a flag in `build.sh` that makes the job only pull from ghcr and exit if the cache is not available yet. 

When that happens, we only have to rerun the test when say linux workflow is completed, and if its available, the test starts right away

MrCyjaneK:
> I'm thinking we can have a flag in build.sh that

I guess fine for now, I am working on the CI refactor anyway

MrCyjaneK:
Though I still don't think it's a good idea to move the tests out of the linux runner - especially after linux will get more love and get all dependencies the build time will be ~30 minutes - same as android - and will cause 2 jobs that share 95% of the tasks to split slightly at the end

test/core/lightning_invoice_service_test.dart
konstantinullrich: Why is this Test removed?

OmarHatem28:
> Though I still don't think it's a good idea to move the tests out of the linux runner - especially after linux will get more love and get all dependencies the build time will be ~30 minutes - same as android - and will cause 2 jobs that share 95% of the tasks to split slightly at the end

I think it should be a dependency and 2 jobs that start after it so there is no duplication or anything.
will make claude do it and see if it does a good job or not

OmarHatem28:
yup, something like this, but I will try to make it in parallel

Screenshot 2026-09-25 at 12 33 35 AM

PR #3643, opened 2026-09-21
Default to onionbalance onion endpoints when built-in Tor is enabled

sethforprivacy:
## Summary

Makes Cake's onionbalance frontends the default onion endpoints whenever built-in Tor is turned on.

**Node lists**
- Replace the legacy XMR onion and both legacy ZEC onions with the `cakexmr…` / `cakezec…` onionbalance frontends (flagged official; XMR also trusted like the clearnet Cake node).
- Add `cakebtc…:50001` and `cakeltc…:50001` Fulcrum onion entries.

**Tor toggle**
- New `SettingsStore.updateNodesForTor`: on enable, any wallet type currently on an official Cake node moves to that type's builtin onion node; on disable, builtin onion nodes move back to the yml default. Custom user nodes are never touched. Called from both toggle paths before reconnecting.

**Migration 72**
- Users pinned to a removed onion move to the new onion when built-in Tor is on, otherwise to the clearnet default (`_changeDefaultNode` convention).

**Zcash**
- zkool supports a `socks5h://` proxy but the wallet never set it, so Zcash was bypassing built-in Tor and could not reach an onion lightwalletd. `connectToNode` now passes the Tor SOCKS port when Tor is running, and `runWithCoin` copies the proxy to its per-call Coin.

**Mempool API**
- All five `ProxyWrapper.get` mempool calls pass the `cakemem…` onion as `onionUri`, so fee estimates and block lookups stay inside Tor.

## Test plan
- [ ] Enable built-in Tor on XMR/BTC/LTC/ZEC wallets that use the Cake default node; confirm current node switches to `Cake Wallet (Onion)` and syncs.
- [ ] Disable built-in Tor; confirm node returns to the clearnet Cake default.
- [ ] Enable built-in Tor on a wallet using a custom node; confirm node is unchanged.
- [ ] Upgrade from a build where the legacy XMR/ZEC onion was selected; confirm migration lands on the expected node.
- [ ] BTC fee estimates load with Tor enabled (mempool onion).


lib/store/settings_store.dart
```diff
@@ -1123,6 +1123,23 @@ abstract class SettingsStoreBase with Store {
     return node;
   }
 
+  Future<void> updateNodesForTor(bool torEnabled) async {
+    final builtin = await Node.getAllBuiltin();
+    runInAction(() {
+      for (final type in nodes.keys.toList()) {
+        final current = nodes[type]!;
+        final isOnion = current.uriRaw.contains('.onion');
```
MrCyjaneK, line 1131: pretty bad idea, if user would have node by any domain that has .onion anywhere in the domain it would switch to it.

lib/store/settings_store.dart
```diff
@@ -1123,6 +1123,23 @@ abstract class SettingsStoreBase with Store {
...
+        Node? target;
+        if (torEnabled && !isOnion && current.isOfficial) {
+          target = builtin.firstWhereOrNull((n) => n.type == type && n.uriRaw.contains('.onion'));
+        } else if (!torEnabled && isOnion && current.isBuiltin) {
+          target = builtin.firstWhereOrNull((n) => n.type == type && n.isDefault);
+        }
+        if (target != null) nodes[type] = target;
```
MrCyjaneK, line 1138: I don't think this is a good idea to randomly default to our node after user explicitly switched away from it. It should be an explicit opt-in switch in settings to enable the behavior and we should allow user to select which nodes to switch between. Of course the defaults would be our nodes but I don't think it is a good idea to forcibly switch users back to our node with an update without explicit notification.

cw_zcash/lib/src/zcash_wallet.dart
MrCyjaneK, line 1: Good catch but we need an explicit note about this in testing so testers know what to test here

lib/entities/default_settings_migration.dart
```diff
@@ -653,6 +657,26 @@ Future<void> defaultSettingsMigration(
...
+            sharedPreferences: sharedPreferences,
+            type: WalletType.zcash,
+            currentNodePreferenceKey: PreferencesKey.currentZcashNodeIdKey,
+            newDefaultUri: builtinTor ? cakeWalletZcashOnionUri : null,
+            oldUri: [
+              '2c4whzg26j6hgjh22rxynj3oig4gm22ga7x74weclujywxye23v3u5id.onion',
+              'bvp2l442g5ogma7rywzahm7eqyhpp3g26n3gvvxeioqn5csio2ir6myd.onion',
+            ],
+          );
+          break;
```
MrCyjaneK, line 679: Hallucination. we auto-update the nodes based on .yaml content so this is not needed

lib/view_model/dashboard/dashboard_view_model.dart
```diff
@@ -1467,6 +1467,7 @@ abstract class DashboardViewModelBase with Store {
       unawaited(ensureTorStarted(context: context).then((_) async {
         if (settingsStore.currentBuiltinTor == false)
           return; // return when tor got disabled in the meantime;
+        await settingsStore.updateNodesForTor(true);
```
MrCyjaneK, line 1470: same as above, I don't think we should automatically with no user consent, no way to opt-in/out do something like switching the node back to our controlled servers

lib/main.dart
```diff
@@ -309,7 +309,7 @@ Future<void> initializeAppConfigs({bool loadWallet = true}) async {
     payjoinSessionSource: payjoinSessionSource,
     anonpayInvoiceInfo: anonpayInvoiceInfo,
     havenSeedStore: havenSeedStore,
-    initialMigrationVersion: 72,
+    initialMigrationVersion: 73,
```
MrCyjaneK, line 312: unnecessary because the migration is unnecessary

MrCyjaneK (changes requested):

lib/store/settings_store.dart
sethforprivacy: Agreed. Onion detection now uses the parsed URI host (`node.uri.host.endsWith('.onion')`) instead of a substring match on the raw URI, so a clearnet domain containing ".onion" anywhere else no longer matches.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

lib/store/settings_store.dart
sethforprivacy: Reworked as opt-in. There is a new "Switch to onion nodes while Tor is enabled" toggle in Connection and sync, default off. When on, each wallet type remembers the node it was on and moves to its onion node while Tor runs, then restores that remembered node when Tor stops (never Cake's default). If the user picks a different onion node while Tor is on, that choice is remembered and used next time, so both sides of the switch are user-selectable through the existing node list. Custom nodes are never rewritten.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

cw_zcash/lib/src/zcash_wallet.dart
sethforprivacy: Added a dedicated "Zcash through Tor" section to the PR test plan, including verifying with a network monitor that lightwalletd traffic goes to the local Tor port rather than the node's public IP, and covering the runWithCoin paths (mempool watcher, address rotation).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

lib/entities/default_settings_migration.dart
sethforprivacy: Right, `validateBuiltinNodes()` already syncs the yml lists into the DB on launch. Dropped the migration case and the two URI constants.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

lib/view_model/dashboard/dashboard_view_model.dart
sethforprivacy: Same opt-in rework as the settings_store thread: this call is now a no-op unless the user has enabled the new setting, and on Tor-off it restores the user's previously selected node rather than switching to Cake's default.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

lib/main.dart
sethforprivacy: Reverted; `initialMigrationVersion` is back to dev's value.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

